PR #3, opened 2024-02-27
Added apiHandler

scruzin:
This is more code copied from VidGrind.
A few of the errors have been wrapped, but otherwise it is unchanged.

### DIFF
--- a/cmd/datablue/handlers.go
+++ b/cmd/datablue/handlers.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math/rand"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/datablue/handlers.go
```diff
@@ -451,6 +453,72 @@ func varsHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprint(w, resp)
 }
 
+// apiHandler handles API requests which take the form:
+//
+//	/api/operation/property/value
```
ao-david, line 458: can we make some comment about what the path looks like after /value? ie chunk or even just "/.."

cmd/datablue/handlers.go
```diff
@@ -451,6 +453,72 @@ func varsHandler(w http.ResponseWriter, r *http.Request) {
...
+			h.Add("Content-Type", "application/octet-stream")
+			h.Add("Content-Disposition", "attachment; filename=\""+val+"\"")
+			rand.Read(body)
+			var i int64
+			for i = 0; i < n; i += chunk {
+				w.Write(body[i : i+chunk])
+			}
+			return
+		}
+
```
ao-david, line 515: Do we know what the future api routes might look like? Is it worth adding in extra cases which return "not yet implemented"

ao-david (approved):

Issue #8, opened 2024-03-18
Add License file


ao-david:
For these packages to be more accessible to the open source community, they should have a correct license file so that gopkg will display them.

see here: https://pkg.go.dev/license-policy

PR #9, opened 2024-03-18
Add GPL licence (#8):

ao-david:
A licence.txt file was added to the root directory to allow packages to be shown on the pkg.go.dev page. This should promote more usage of the packages for open source devs.


trekhopton:
Shouldn't this file contain the actual licence? All good if you've tested that it's smart enough to recognise that it's GPL.

ao-david:
I have now used the GitHub feature that is built in which has automatically generated the file in the home directory, so this should work.

trekhopton (approved):

PR #11, opened 2024-04-17
Initial revision

scruzin:
This is all of the broadcast-related code factored out of VidGrind.

main.go has a single entry point, /broadcast. Currently, only /broadcast/save is supported.

Other files are largely unchanged, except for:
- using cloud/notify for notifications.
- removing used functions form utils.go,
- updating licenses to refer to Ocean TV instead of VidGrind.


cmd/oceantv/main.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	err = notifier.Init(ctx, projectID, senderEmail, &timeStore{})
+	if err != nil {
+		log.Fatalf("could not set up email notifier: %v", err)
+	}
+}
+
+// timeStore implements a notify.TimeStore that uses iotds.Variable for persistence.
+type timeStore struct {
```
saxon-milton, line 142: Has gofmt been run ? I feel like it would have made this 
`type timeStore struct {}`

saxon-milton (approved):

ao-david:
I think that given that Ocean TV has lots of testing, that we should try and integrate CircleCI, or some other testing platform with GitHub as soon as possible.

cmd/oceantv/main.go
scruzin: Yes, gofmt has been run and it was happy.

PR #20, opened 2024-05-14
cmd/datablue/handlers.go: log device error responses regardless of debug

trekhopton:
This was done because it was rather hard to debug errors being returned to ESP clients without redeploying datablue in debug mode (undesirable).


cmd/datablue/handlers.go
```diff
@@ -538,7 +538,5 @@ func writeDeviceError(w http.ResponseWriter, dev *iotds.Device, err error) {
 	}
 	w.Header().Add("Content-Type", "application/json")
 	fmt.Fprint(w, `{"er":"`+err.Error()+`"`+rc+`}`)
-	if debug {
-		log.Println("Wrote error: " + err.Error())
-	}
+	log.Println("Wrote error: " + err.Error())
```
scruzin, line 541: Now that's we're always logging, let's make it slightly more descriptive, e.g., "Wrote device error".

saxon-milton (approved):

PR #25, opened 2024-05-16
ts/nav-menu.ts: if there is no site selected, don't show nav menu links

trekhopton:
This was done because, unintentionally, before even logging in, all the nav menu links were showing, it wasn't until the site menu had a site selected that the menu dynamically showed links.


cmd/oceanbench/ts/nav-menu.ts
```diff
@@ -84,6 +84,14 @@ export class NavMenu extends LitElement {
         const ul = this.shadowRoot?.querySelector('#menu') as HTMLUListElement;
         for (const item of items) {
             ul.appendChild(item);
+            const perm = Number(item.dataset && item.dataset['perm']);
+            if(perm && (perm & this.selectedPerm) != 0){
```
saxon-milton, line 88: A couple of questions. 
1) Could you just do
`if(perm & this.selectedPerm) != 0){`
2) Is lit not smart enough to let you do 
`if(perm & this.selectedPerm){`
?

saxon-milton (approved):

cmd/oceanbench/ts/nav-menu.ts
trekhopton: After playing around it looks like I can do that. In javascript undefined bitwise and 0 is just 0 so that still works and 0 evaluates to false so that works too.

PR #55, opened 2024-05-21
Use GitHub ausocean packages

scruzin:
Mechanical changes resulting from replacing github.com/ausocean packages with bitbucket.org/ausocean counterparts.

Imports of "golang.org/x/net/context" have also been replaced with simply "context".

There should be no more references to bitbucket in this repo!


cmd/oceanbench/set.go
```diff
@@ -63,7 +64,7 @@ type quantity struct {
 
 // defaultQuantities provides a list of common quantities we might measure.
 // This was originally defined in the netreceiver code, see
-// bitbucket.org/ausocean/iotsvc/netreceiver/lists.py:420
+// github.com/ausocean/clientsvc/netreceiver/lists.py:420
```
trekhopton, line 67: There's an error here, we didn't migrate this code so the link should remain the same. That being said, I don't think we need the reference to netreceiver, it doesn't provide any extra information.

trekhopton (changes requested):
Small error in a comment but otherwise LGTM.

cmd/oceanbench/set.go
scruzin: Fixed.

trekhopton (approved):

Issue #57, opened 2024-05-22
bench: remove device request handlers


scruzin:
Device requests are now handled by the datablue service.

PR #58, opened 2024-05-23
Removed unused code

scruzin:


cmd/oceanbench/video.go
```diff
@@ -162,7 +159,7 @@ func writeMtsMedia(ctx context.Context, mid int64, gh string, ts int64, data []b
 			// Output up to the start of this PSI, then start a new clip.
 			ts = int64(t)
 			sz := i + psiSize + j
-			if sz > datastore.MaxBlob && trimMTS {
```
trekhopton, line 165: Why was this change made? It's good to provide comments on changes that are unexpected in PRs, eg. I expected this PR to just be removing unused code but there a MTS video handling change here.

scruzin: It got picked up with an earlier change to obsolete the trimMTS variable. Basically, we _always_ trip invalid MTS data now. 

BTW, it should never happen. This was added a while ago to defend against invalid MTS/audio files produced by SCUHU.

trekhopton: Okay, cheers.

trekhopton (approved):
Nice. A satisfying PR :)

scruzin:
Resolves Issue #57

Issue #16, opened 2024-05-12
broadcast: provide logger to youtube binding


saxon-milton:
This will allow us to wrap youtube binding logs in a message containing broadcast id.

PR #59, opened 2024-05-28
Overhaul logging for broadcast code

saxon-milton:
Addresses issue #16 

There were a lot of calls to "log.Printf" of the form "Broadcast: %s, id: %s ....", so we're factoring this pattern out. We're doing this by requiring that most interfaces that the broadcast code uses take a "log" parameter of type func(string,...interface{}). This means that we can pass around a log function that wraps any messages provided with the broadcast name and id.

We're now also using this in areas where the broadcast name and id were not specified. This means that all logging related to a particular broadcast is labelled as such.

### DIFF
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -239,9 +239,14 @@ func performChecksInternalThroughStateMachine(
 	cfg *BroadcastConfig,
 	timeNow func() time.Time,
 	store datastore.Store,
-	svc BroadcastService,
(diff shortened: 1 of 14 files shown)


trekhopton (approved):

scruzin:
This is a much-needed change. But should we bite the bullet and add a log **level**, so we can de-clutter the logs?

scruzin (commented):
This is a much-needed change. But should we bite the bullet and add a log level, so we can de-clutter the logs?

PR #65, opened 2024-05-29
oceanbench/main.go: if environment is not appengine, use project level template path

trekhopton:
This is done to support running oceanbench locally using the datstore (not standalone).


scruzin:
LGTM, but please test the change in production if you have not already done so.

scruzin (approved):
LGTM

PR #73, opened 2024-05-31
oceanbench broadcast: remove templates from the js and separate the js from the HTML

trekhopton:
This was done as an initial step for getting templates out of the javascript. We shouldn't use templates in javascript because it upsets HTML and JS formatters.

We also no longer need StartTime and EndTime in the broadcast config. There are already two other fields that store this information. The javascript can just use the timestamps.


cmd/oceanbench/main.go
```diff
@@ -168,6 +168,16 @@ var templateFuncs = template.FuncMap{
 	"localdate":     formatLocalDate,
 	"localtime":     formatLocalTime,
 	"localdatetime": formatLocalDateTime,
+	"json":          toJSON,
+}
+
+func toJSON(v interface{}) string {
```
scruzin, line 174: Please move this function into utils.go

cmd/oceantv/broadcast.go
```diff
@@ -97,10 +97,8 @@ type BroadcastConfig struct {
 	Description       string        // The broadcast description shown below viewing window.
 	Privacy           string        // Privacy of the broadcast i.e. public, private or unlisted.
 	Resolution        string        // Resolution of the stream e.g. 1080p.
-	StartTime         string        // Start time of the broadcast in yy/mm/dd, hh:mm format.
 	StartTimeUnix     string        // Start time of the broadcast in unix format.
```
scruzin, line 100: Prefer Timestamp over TimeUnix.

scruzin (changes requested):

saxon-milton:
> This was done as an initial step for getting templates out of the javascript. We shouldn't use templates in javascript because it upsets HTML and JS formatters.
> 
> We also no longer need StartTime and EndTime in the broadcast config. There are already two other fields that store this information. The javascript can just use the timestamps.

Are there other reasons why we shouldn't use templates in HTML and JS formatters ? Because I don't think this is a very good one.

cmd/oceanbench/main.go
saxon-milton: This could also be called "marshal", but up to you.

saxon-milton (approved):

ao-david (approved):

trekhopton:
> > This was done as an initial step for getting templates out of the javascript. We shouldn't use templates in javascript because it upsets HTML and JS formatters.
> > We also no longer need StartTime and EndTime in the broadcast config. There are already two other fields that store this information. The javascript can just use the timestamps.
> 
> Are there other reasons why we shouldn't use templates in HTML and JS formatters ? Because I don't think this is a very good one.

HTML templates belong in HTML, not javascript. We could hypothetically use templates in any code but it's not good practice imo and there isn't much support for that sort of thing. Not being able to format HTML code is just a hindrance to development work flow. With a little more effort we can keep templates in the HTML and things are much neater.

cmd/oceantv/broadcast.go
trekhopton: I'll do this in another PR.

cmd/oceanbench/main.go
trekhopton: I moved it to utils. True, I prefer toJSON.

Issue #71, opened 2024-05-31
datablue: Add support for binary data


scruzin:

scruzin:
Implemented in Pull Request #74

PR #74, opened 2024-05-31
Port binary data

scruzin:
This addresses Issue #71.

### DIFF
--- a/cmd/datablue/handlers.go
+++ b/cmd/datablue/handlers.go
@@ -246,7 +246,7 @@ func pollHandler(w http.ResponseWriter, r *http.Request) {
 			err = writeScalar(r, ma, pin, n)
 
 		case 'B':
-			// Not implemented.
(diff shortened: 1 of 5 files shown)


cmd/datablue/handlers.go
```diff
@@ -360,6 +360,23 @@ func writeText(r *http.Request, ma, pin string, n int) error {
 	return model.WriteText(r.Context(), mediaStore, &model.Text{MID: mid, Timestamp: ts, Data: string(data), Type: tt})
 }
 
+// writeBinary writes binary data.
```
saxon-milton, line 363: Unless this comment is expanded, I don't think it's worth having (unexported functions don't need comments). If we want to keep, it might be worth saying that the data is pulled from the request body and written to the datastore in a Binary entity.

saxon-milton (approved):

Issue #77, opened 2024-06-01
notify: Add options to specify sender, recipient, filter(s), etc.


scruzin:
It is proposed to change notifier.Init from:

`func (n *Notifier) Init(ctx context.Context, projectID string, sender string, store TimeStore) error`

to 

`func (n *Notifier) Init(ctx context.Context, projectID string, option... Option) error
`
where Option would be defined as:

`type Option func(*Notifier,string) error`

The following four options are envisioned:

- WithStore - specifies using a datastore for notification persistence
- WithSender - specifies which sender to use
- WithRecipient - specifies which recipient to use
- WithFilter - specifies a substring that is used to filter messages. More than one WithFilter can be applied to achieve a compound filter, i.e, A AND B.

Clients may instantiate multiple Notifiers, each configured with its own options.

saxon-milton:
Nice; the functional options pattern would have to be one of my favourites.

I'm wondering what the string parameter is for in the Option type however ? The way that I have seen this used (and used personally), is where you provide parameters for the option through the factories for the Options e.g.

```
type Option func(*Notifier) error

func WithSender(sender string) Option {
    return func(n *Notifier) error {
        n.sender = sender
        return nil
    }
}
```

scruzin:
You're right. The Option doesn't need to take a string. I was conflating the fact that all option params are of type string.

scruzin:
With this approach, notify.Send no longer takes a recipient, as the latter is now specified as an option to notify.Init.
If necessary, different Notifiers can/should be instantiated for different recipients. For example, we might define broadcast-ops@, device-ops@, networkblue-ops@, etc. for different operations needs (which could be handled by different teams).

scruzin:
Implemented in Pull Request #80

PR #80, opened 2024-06-01
Notifier options

scruzin:
This implements the Notify package support for Issue #77 .

Note that **SendOps** is now deprecated as the functionality can be readily achieved by initializing a **Notifier** with the _WithRecipient_ option set to the ops email address. The GetOpsEnvVars helper function has been added to make this a bit easier.


saxon-milton (approved):
Nice!

Issue #82, opened 2024-06-01
notify: Support multiple recipients


scruzin:
It would be useful if notifications could be to multiple recipients simultaneously. We currently achieve this by means of using email groups, but group functionality may not always be available or desirable.

scruzin:
Implemented.

PR #84, opened 2024-06-02
Support for multiple recipients

scruzin:
This implements Issue #82

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -28,20 +28,17 @@ import (
 	mailjet "github.com/mailjet/mailjet-apiv3-go"
 )
 
-const (
-	defaultSender    = "vidgrindservice@gmail.com"
-	defaultRecipient = "ops@ausocean.org"
-)
+const defaultSender = "vidgrindservice@gmail.com"
 
 // Notifier represents a notifier that uses the MailJet API to send email.
 type Notifier struct {
-	mutex       sync.Mutex // Lock access.
-	sender      string     // Sender email address.
-	recipient   string     // Recipient email address.
-	store       TimeStore  // Notification store (optional).
-	filters     []string   // Message filters (optional).
-	publicKey   string     // Public key for accessing MailJet API.
-	privateKey  string     // Public key for accessing MailJet API.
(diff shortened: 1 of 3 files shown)


notify/notify.go
```diff
@@ -80,29 +77,33 @@ func (n *Notifier) Init(options ...Option) error {
...
 	if n.publicKey != "" && n.privateKey != "" {
 		clt := mailjet.NewMailjetClient(n.publicKey, n.privateKey)
+		var recipients mailjet.RecipientsV31
+		for _, recipient := range n.recipients {
+			recipients = append(recipients, mailjet.RecipientV31{Email: recipient})
+		}
 		info := []mailjet.InfoMessagesV31{{
 			From:     &mailjet.RecipientV31{Email: n.sender},
-			To:       &mailjet.RecipientsV31{mailjet.RecipientV31{Email: n.recipient}},
+			To:       &recipients,
```
davidsutts, line 106: This doesn't look quite right, why don't we still need to use the RecipientsV31 type here?

notify/options.go
```diff
@@ -36,18 +36,31 @@ func WithSender(sender string) Option {
 	}
 }
 
-// WithRecipient sets the recipient email address.
+// WithRecipient sets a single recipient email address.
 func WithRecipient(recipient string) Option {
```
davidsutts, line 40: Could it make sense to make the WithRecipient option append a recipient rather than set that as the only recipient?

davidsutts: or maybe check that it isn't trying to override a previous recipient?

davidsutts (approved):
I think there might be a few changes to be made, but otherwise LGTM!

notify/notify.go
scruzin: RecipientsV31 is nothing more than []RecipientV31, which recipients is.

notify/options.go
scruzin: Unlike filters, we need a way to override a recipient. The new WithRecipients can be used for multiple recipient.

Issue #78, opened 2024-06-01
broadcast: only send bad health notification after consecutive issues


saxon-milton:
We frequently get a situation where there's a temporary health issue like "noData". This is generally just due to a momentary lapse in bandwidth. We therefore don't care all that much unless it continues. So let's only use email notification if we get consecutive problems.

scruzin:
This might also be a good use case for sending an email alert to a dedicated email address, such as broadcast-ops@.
Less serious alerts could continue to use ops@.

PR #94, opened 2024-06-03
Only email notify on consecutive bad health issues

saxon-milton:
Depends on PR #92 

Resolve issue #78 

We used to email on any issue, even if it might have been just temporary. This is uneccessary and creates false concerns. So now we're only performing email notification on consecutive issues.

This has involved modifying how we check for issues i.e. we return the string of the issue rather than a bool. We also provide this to the badHealthCallback, where we can handle with health notification if we wish.

### DIFF
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -319,7 +319,13 @@ func performChecksInternalThroughStateMachine(
 				context.Background(),
 				cfg,
 				func() { bus.publish(goodHealthEvent{}) },
-				func() { bus.publish(badHealthEvent{}) },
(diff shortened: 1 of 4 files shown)


trekhopton (approved):

scruzin (approved):
It would be nice to be able to configure the number of consecutive bad health checks that constitute a serious enough problem to warrant an email notification. Otherwise, LGTM.

saxon-milton:
> It would be nice to be able to configure the number of consecutive bad health checks that constitute a serious enough problem to warrant an email notification. Otherwise, LGTM.

Yes fair. I'm going to create an issue for this.

Issue #75, opened 2024-06-01
broadcast: doesn't seem like directLiveUnhealthy is being transitioned to


saxon-milton:
We had an episode where a broadcast was reporting no data health statuses for about 50 minutes before the broadcast was completed/revoked. There is no sign in the logs that the unhealthy state was transitioned into and that we tried to remedy the problem with hardware restarts. We need to check this transition is working correctly.

saxon-milton:
It has been discovered that this was due to the Issue tally stored in the config not being saved on update i.e. it was at 0 every time the SM was spun up.

PR #95, opened 2024-06-03
Save broadcast issues tally

saxon-milton:
Depends on PR #94 

Resolves issue #75

It turns out that we weren't getting any badHealthEvents published and subsequently never transitioning to the Unhealthy states because we weren't actually saving the changes to the issues tally to the datastore. This change fixes this.

It has required that we alter the interface of HandleHealth slightly to require a parameter of Store interface type.


trekhopton (approved):

scruzin (approved):
Good fix!

PR #96, opened 2024-06-03
Fix direct state behaviour

saxon-milton:
Depends on PR #95 

The behaviour has been incorrect. To start, we should not stop the broadcast on exit of the directLive state. Maybe once upon a time this worked because there was no state other than idle that we'd exit to, but now we have a directLiveUnhealthy. So, we only try to stop the broadcast (if exists) on entry to directIdle.

Secondly, the directLiveUnhealthy similarly should not try to stop the broadcast on exit. It should also implement the fixable state so that it can try to fix the hardware in this state.

### DIFF
--- a/cmd/oceantv/broadcast_states.go
+++ b/cmd/oceantv/broadcast_states.go
@@ -307,26 +307,49 @@ func newDirectLive(ctx *broadcastContext) *directLive {
 	return &directLive{ctx}
 }
 func (s *directLive) enter() {}
-func (s *directLive) exit() {
-	err := s.man.StopBroadcast(context.Background(), s.cfg, s.store, s.svc)
-	if err != nil {
-		s.log("could not stop broadcast")
-	}
-}
+func (s *directLive) exit()  {}
 
 type directLiveUnhealthy struct {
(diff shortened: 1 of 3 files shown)


cmd/oceantv/broadcast_states.go
```diff
@@ -307,26 +307,49 @@ func newDirectLive(ctx *broadcastContext) *directLive {
...
-	err := s.man.StopBroadcast(context.Background(), s.cfg, s.store, s.svc)
-	if err != nil {
-		s.log("could not stop broadcast", s.cfg.Name, s.cfg.ID)
+func (s *directLiveUnhealthy) exit()  {}
+func (s *directLiveUnhealthy) fix() {
+	notify := func(msg string) {
+		notifier.SendOps(context.Background(), s.cfg.SKey, "health", msg)
 	}
+
+	const resetInterval = 5 * time.Minute
```
scruzin, line 328: Again, I'd prefer to these consts given more descriptive names (e.g., directLiveUnhealthyResetInterval.) and promoted to one place where all of them can be configured

scruzin (approved):
LGTM

PR #97, opened 2024-06-03
Implement stateWithTimeout for hardwareStarting

saxon-milton:
Depends on PR #96 

Up until this point there has been no way to know if a start failure happened specifically because the hardware failed to start.

We're fixing this by having the hardwareStarting state implement the stateWithTimeout interface so that when it times out we can emit a hardwareStartFailureEvent and precisely indicate the cause of failure.

This has required that we now save hardware state data, so there's a new field of the config called HardwareStateData. It has required that we add JSON marshalling tags to the other states.


scruzin (approved):
LGTM. A very useful change.

PR #98, opened 2024-06-03
Report start failure on starting state timeout

saxon-milton:
Depends on PR #97 

This should not happen under normal circumstances given that we now report start failure on both broadcast live transition issues and hardware camera startup failures, but it will cover us with any other unforseen conditions.

We're modifying the timeout to 10 minutes, which will give the hardware enough time to timeout (5 minutes).


scruzin (approved):
I'd like to see these timeouts as consts with nice names, but okay for now.

PR #100, opened 2024-06-05
Move http request creation inside rpc action func

scruzin:
The http req was previously created outside the action func. Moving it inside is probably safer.


trekhopton:
It would be helpful if you added a quick description to the PR stating why this change was made.

trekhopton:
> It would be helpful if you added a quick description to the PR stating why this change was made.

Okay, I've just read your email. It'd still be helpful to put it here.

trekhopton (approved):

Issue #102, opened 2024-06-05
bench: selecting "Cancel" when prompted to keep changes reloads anyway


trekhopton:
When the dialog appears asking a user if they want to change sites or keep the current one, even if they click cancel the page still reloads and the lose any data they have filled out.

trekhopton:
FIxed in PR #103

Issue #101, opened 2024-06-05
bench: selecting a different site while on the broadcast page causes two dialog boxes to appear and the page doesn't load.


trekhopton:
Selecting a different site while on the broadcast page causes two dialog boxes to appear and the page doesn't load.

This happened due to a recent "fix".

trekhopton:
Fixed in PR #103

PR #90, opened 2024-06-03
Handle Site Changes on Broadcast Page

ao-david:
The site-menu now has the ability to fire events for pages to listen for when the site is changed. This can allow for more control over what happens when the site is changed on sensitive pages.

On the broadcast page, the page will listen for the event, and check with the user that they want to change the site, if they do, the page 'refreshes' but does not send any of the old data. This addresses #88

PR #103, opened 2024-06-05
bench: only prompt once and don't reload when user doesn't want to.

trekhopton:
This was done to fix a change in PR #90 that wasn't really tested.

Addresses issues #101 and #102.

Also fixes an error that happens if the sensor list is empty.


cmd/oceanbench/s/broadcast.js
```diff
@@ -8,34 +8,40 @@ document.addEventListener('DOMContentLoaded', function() {
...
   // Check that the user really wants to change the site.
-  let resp = confirm("Unsaved changes may be lost!\nAre you sure you want to change sites?")
+  let resp = confirm("Unsaved changes may be lost!\nAre you sure you want to change sites?");
+
+  // Remove the site menu to prevent other events triggering more dialog boxes.
+  const menu = document.getElementById("sitemenu");
+  menu.remove(menu);
   if (resp) {
-    location.assign("/admin/broadcast"); // This will empty the form.
+    location.assign("admin/broadcast"); // This will empty the form.
```
ao-david, line 34: I seem to be getting navigated back to the index page with this change.

ao-david: This takes you to a relative path ie /admin/admin/broadcast

cmd/oceanbench/s/broadcast.js
```diff
@@ -8,34 +8,40 @@ document.addEventListener('DOMContentLoaded', function() {
...
 });
 
 function handleSiteChange(event) {
   // Check that the user really wants to change the site.
-  let resp = confirm("Unsaved changes may be lost!\nAre you sure you want to change sites?")
+  let resp = confirm("Unsaved changes may be lost!\nAre you sure you want to change sites?");
+
+  // Remove the site menu to prevent other events triggering more dialog boxes.
+  const menu = document.getElementById("sitemenu");
+  menu.remove(menu);
```
ao-david, line 32: This menu needs to be reinstated if the user decides to stay on the page.

cmd/oceanbench/s/broadcast.js
```diff
@@ -8,34 +8,40 @@ document.addEventListener('DOMContentLoaded', function() {
...
-    location.assign("/admin/broadcast"); // This will empty the form.
+    location.assign("admin/broadcast"); // This will empty the form.
   } else {
     // make a request to keep the old site.
     let xhr = new XMLHttpRequest();
     xhr.open("GET", "/api/set/site/"+event.detail["previousSite"]);
     xhr.onreadystatechange = ()=> {
       if (xhr.readyState == XMLHttpRequest.DONE && xhr.responseText == "OK") {
-        location.reload();
+        console.log("request to keep previous selected site successful");
```
ao-david, line 41: probably reinstate the site-menu here

ao-david (changes requested):

cmd/oceanbench/s/broadcast.js
trekhopton: Fixed.

cmd/oceanbench/s/broadcast.js
trekhopton: It never gets removed now.

ao-david (approved):

cmd/oceanbench/s/broadcast.js
```diff
@@ -8,35 +8,31 @@ document.addEventListener('DOMContentLoaded', function() {
...
+  document.getElementById("header").addEventListener("site-change", handleSiteChange);
 });
 
 function handleSiteChange(event) {
-  // Check that the user really wants to change the site.
-  let resp = confirm("Unsaved changes may be lost!\nAre you sure you want to change sites?")
-  if (resp) {
-    location.assign("/admin/broadcast"); // This will empty the form.
-  } else {
-    // make a request to keep the old site.
```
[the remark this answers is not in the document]
trekhopton: Yeah that's right. I think for now it's fine to assume the user knows they can lose data if they don't click save and change the site menu. It would be nice to warn if not saved in the future but not needed.

PR #107, opened 2024-06-05
bench: add /learn and /learn/mooring for mooring-sim

trekhopton:
This was done to help us and schools calculate rig mooring line lengths.


ao-david (approved):
LGTM.

Would be good to be able to access this via cloudblue and return to cloudblue eventually (or does this work well in an iframe?)

PR #118, opened 2024-06-12
OceanTV: Add controller MAC to broadcast config

ao-david:
The broadcast config now contains the controller MAC allowing the state machine to make queries about the status of the controller.

This value is also read from the broadcast page.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -89,7 +89,8 @@ const (
 type broadcastRequest struct {
 	BroadcastVars      []model.Variable // Holds prior saved broadcast configs.
 	CurrentBroadcast   BroadcastConfig  // Holds configuration data for broadcast config in form.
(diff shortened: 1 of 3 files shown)


cmd/oceanbench/broadcast.go
```diff
@@ -114,6 +115,7 @@ type BroadcastConfig struct {
 	End               time.Time     // End time in native go format for easy operations.
 	VidforwardHost    string        // Host address of vidforward service.
 	CameraMac         int64         // Camera hardware's MAC address.
+	ControllerMac     int64         // Controller hardware's MAC adress (controller used to power camera).
```
saxon-milton, line 118: I think should be ControllerMAC

saxon-milton: We'll have to change CameraMac above at some point too

saxon-milton (approved):
Just the naming. I think it's go idiomatic to use capitals in acronyms unless you can't like 
```var macAddress string```

Issue #114, opened 2024-06-12
OceanTV: Check status of controller and notify 


ao-david:
When the broadcast tries to start initially, it will make requests to the hardware to start, however, it doesn't currently check the status of the ESP. It would be helpful to first check the status of the ESP and notify if the ESP is not reporting. This would typically indicate that the rig has gone flat and has not yet booted back up, this would help debugging.

PR #119, opened 2024-06-12
OceanTV: Check status of controller before starting broadcast (#114)

ao-david:
createBroadcastAndRequestHardware now makes a call to getDeviceStatus for the controller to ensure that the controller will even respond to hardware start events before trying to create a new broadcast. This will give a pretty good indicator that the controller is not powered, likely due to a low power event on the rig.

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -34,6 +34,18 @@ func newHardwareStarting(ctx *broadcastContext) *hardwareStarting {
 }
 func (s *hardwareStarting) enter() {
 	s.LastEntered = time.Now()
+	if s.cfg.ControllerMAC != 0 {
+		controllerIsOn, err := model.DeviceIsUp(context.Background(), settingsStore, model.MacDecode(s.cfg.ControllerMAC))
+		if err != nil {
+			s.log("error getting controller status: %v", err)
+			s.bus.publish(controllerFailureEvent{})
+			return
+		} else if !controllerIsOn {
+			s.log("controller not responding")
+			s.bus.publish(controllerFailureEvent{})
+			return
+		}
+	}
 	s.camera.start(s.broadcastContext)
 }
 func (s *hardwareStarting) exit() {}
@@ -125,6 +137,8 @@ func (sm *hardwareStateMachine) handleEvent(event event) error {
 		sm.handleHardwareStartRequestEvent(event.(hardwareStartRequestEvent))
 	case hardwareStopRequestEvent:
 		sm.handleHardwareStopRequestEvent(event.(hardwareStopRequestEvent))
+	case controllerFailureEvent:
+		sm.handleControllerFailureEvent(event.(controllerFailureEvent))
 	default:
 		// Do nothing.
 	}
@@ -244,6 +258,16 @@ func (sm *hardwareStateMachine) handleHardwareResetRequestEvent(event hardwareRe
 		sm.transition(newHardwareStarting(sm.ctx))
 	case *hardwareRestarting, *hardwareStarting, *hardwareStopping:
 		// Ignore.
+	default:
+		sm.unexpectedEvent(event, sm.currentState)
+	}
+}
+
+func (sm *hardwareStateMachine) handleControllerFailureEvent(event controllerFailureEvent) {
+	sm.log("handling controller failure event")
+	switch sm.currentState.(type) {
+	case *hardwareOn, *hardwareRestarting, *hardwareStopping, *hardwareStarting:
+		sm.transition(newHardwareOff())
 	default:
 		sm.unexpectedEvent(event, sm.currentState)
 	}
--- a/cmd/oceantv/broadcast_machine.go
+++ b/cmd/oceantv/broadcast_machine.go
@@ -68,6 +68,8 @@ func (sm *broadcastStateMachine) handleEvent(event event) error {
 		sm.handleGoodHealthEvent(event.(goodHealthEvent))
 	case fixFailureEvent:
 		sm.handleFixFailureEvent(event.(fixFailureEvent))
+	case controllerFailureEvent:
+		sm.handleControllerFailureEvent(event.(controllerFailureEvent))
 	}
 
 	// After handling of the event, we may have some changes in substates of the current state.
@@ -94,7 +96,7 @@ func (sm *broadcastStateMachine) handleHardwareStartFailedEvent(event hardwareSt
 	sm.log("handling hardware start failed event")
 	switch sm.currentState.(type) {
 	case *vidforwardPermanentStarting, *vidforwardSecondaryStarting, *directStarting:
-		onFailureClosure(sm.ctx, sm.ctx.cfg)(errors.New("hardware start failed"))
+		onFailureClosure(sm.ctx, sm.ctx.cfg, false)(errors.New("hardware start failed"))
 	case *vidforwardPermanentLive, *vidforwardPermanentLiveUnhealthy:
 		sm.logAndNotify(broadcastHardware, "hardware failure event in permanent live state, moving to failure slate state")
 		sm.transition(newVidforwardPermanentFailure(sm.ctx))
@@ -153,6 +155,26 @@ func (sm *broadcastStateMachine) handleGoodHealthEvent(event goodHealthEvent) er
 	return nil
 }
 
+var errControllerFailure = errors.New("controller not reporting")
+
+func (sm *broadcastStateMachine) handleControllerFailureEvent(event controllerFailureEvent) error {
+	sm.log("handling controller failure event")
+	switch sm.currentState.(type) {
+	case *directStarting:
+		onFailureClosure(sm.ctx, sm.ctx.cfg, true)(errControllerFailure)
+		sm.transition(newDirectIdle(sm.ctx))
+	case *vidforwardPermanentStarting:
+		onFailureClosure(sm.ctx, sm.ctx.cfg, true)(errControllerFailure)
+		sm.transition(newVidforwardPermanentIdle(sm.ctx))
+	case *vidforwardSecondaryStarting:
+		onFailureClosure(sm.ctx, sm.ctx.cfg, true)(errControllerFailure)
+		sm.transition(newVidforwardSecondaryIdle(sm.ctx))
+	default:
+		// Do nothing.
+	}
+	return nil
+}
+
 func (sm *broadcastStateMachine) handleTimeEvent(event timeEvent) {
 	sm.log("handling time event: %v", event.Time)
 	switch sm.currentState.(type) {
@@ -194,12 +216,12 @@ func (sm *broadcastStateMachine) handleTimeEvent(event timeEvent) {
 	case *directStarting:
 		withTimeout := sm.currentState.(stateWithTimeout)
 		if withTimeout.timedOut(event.Time) {
-			onFailureClosure(sm.ctx, sm.ctx.cfg)(errors.New("direct starting timed out"))
+			onFailureClosure(sm.ctx, sm.ctx.cfg, false)(errors.New("direct starting timed out"))
 		}
 	case *vidforwardPermanentStarting:
 		withTimeout := sm.currentState.(stateWithTimeout)
 		if withTimeout.timedOut(event.Time) {
-			onFailureClosure(sm.ctx, sm.ctx.cfg)(errors.New("permanent starting timed out"))
+			onFailureClosure(sm.ctx, sm.ctx.cfg, false)(errors.New("permanent starting timed out"))
 		}
 	case *vidforwardPermanentTransitionSlateToLive:
 		withTimeout := sm.currentState.(stateWithTimeout)
--- a/cmd/oceantv/broadcast_states.go
+++ b/cmd/oceantv/broadcast_states.go
@@ -611,13 +611,13 @@ func createBroadcastAndRequestHardware(ctx *broadcastContext, cfg *BroadcastConf
 		ctx.svc,
 	)
 	if err != nil {
-		onFailureClosure(ctx, cfg)(fmt.Errorf("could not create broadcast: %v", err))
+		onFailureClosure(ctx, cfg, false)(fmt.Errorf("could not create broadcast: %v", err))
 		return
 	}
 	if onCreation != nil {
 		err = onCreation()
 		if err != nil {
-			onFailureClosure(ctx, cfg)(fmt.Errorf("could not create broadcast: %v", err))
+			onFailureClosure(ctx, cfg, false)(fmt.Errorf("could not create broadcast: %v", err))
 			return
 		}
 	}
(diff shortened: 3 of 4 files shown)


cmd/oceantv/broadcast_states.go
```diff
@@ -571,7 +571,16 @@ func broadcastCfgToState(ctx *broadcastContext) state {
 }
 
 func createBroadcastAndRequestHardware(ctx *broadcastContext, cfg *BroadcastConfig, onCreation func() error) {
-	err := ctx.man.CreateBroadcast(
+	background := context.Background()
```
saxon-milton, line 574: This works OK, but I'm thinking it might be better if we put this in the hardware state machine for better encapsulation. We can check this in response to the handling of a hardwareStartRequestEvent, and if not OK, then we can emit a new event controllerFailedEvent or something along those lines. Then the broadcast state machine can handle this, send an email notification, and disable the broadcast. 

Also, using onFailureClosure as you might know performs the increment of the failure counter and disables after a few failures. I'm thinking on this failure we instead disable straight away. We can factor out the disabling code from the onFailureClosure so both mechanisms can use it.

saxon-milton (changes requested):

cmd/oceantv/broadcast_states.go
ao-david: Thanks for the feedback, I thought I would just try something and get your thoughts on where it should go, as I have only a passing understanding of the full structure.

saxon-milton: No worries, you managed to do this quickly! It's good, but I think it could be slightly better :)

ao-david: Also, do you have any thoughts on broadcasts without controllers? Should we just skip the check if the controller MAC is empty?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -32,6 +32,16 @@ func newHardwareStarting(ctx *broadcastContext) *hardwareStarting {
 }
 func (s *hardwareStarting) enter() {
 	s.LastEntered = time.Now()
+	if s.cfg.ControllerMAC != 0 {
+		controllerIsOn, err := getDeviceStatus(context.Background(), s.cfg.ControllerMAC, settingsStore)
+		if err != nil {
+			disableBroadcast(s.cfg, 1, fmt.Errorf("failed to get controller status: %v", err))
```
saxon-milton, line 38: So I'm thinking in both of the error cases we should instead publish a controllerFailureEvent (doesn't exist yet). We'll add a handler for this in this state machine, which we then use to transition from this state to hardwareOff. The broadcast state machine will also handle this event type and there we transition to the idle state for secondary and direct broadcasts (permanent broadcasts should go to slate). That's also where we can call onFailureClosure (my bad for suggesting we factor out disableBroadcast; that could work, but now I'm thinking onFailureClosure could take another parameter that can be used to ignore maxFailures?) Thoughts apprciated.

saxon-milton (changes requested):

cmd/oceantv/broadcast_hardware_machine.go
ao-david: These seems reasonable, maybe instead of ignoring maxFailures, you can pass the number of failures you want to allow? This would make this value more configurable for different attempts?

ie controller failures only get 1 attempt, youtube failures get 3, and hardware failures get 5. This would also make it easier to allow the user to specify this via cloudblue

cmd/oceantv/broadcast_states.go
saxon-milton: Yeah let's just skip the check in that case.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: That could work, but what feels weird is that we don't have different issue tallies for each type of issue. Instead maybe we need an issue severity level or something, i.e. a factor that is multiplied by the issue tally increment i.e. we still have a consistent max limit but different issues will have different severities and go over it faster. Alternatively, we could just have two severity levels. One that increments the tally, and one that just fails completely immediately.

cmd/oceantv/broadcast_states.go
scruzin: Why are we not using the context that is supplied to the function, and instead using a background context?

cmd/oceantv/broadcast_hardware_machine.go
ao-david: This feels like maybe beyond the scope of this PR. Maybe for now, it should include a instant fail bool which skips the tally, and any additional changes can be implemented later?

cmd/oceantv/broadcast_states.go
ao-david: The context supplied to the function is a broadcast context, which I don't think implements the context interface (not that I think there is one). Could be wrong though.

 @saxon-milton?

cmd/oceantv/broadcast_hardware_machine.go
ao-david: I have currently implemented a disable on first fail boolean argument which will allow failures to skip the count and fail immediately, and have written an issue to make this better implemented #144.

ao-david:
The check now fires a controllerFailureEvent which is handled by both the hardware state machine (transition back to hardware off) and the broadcast state machine (transitions back to idle or slate and calls onfailureclosure)

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -247,6 +259,20 @@ func (sm *hardwareStateMachine) handleHardwareResetRequestEvent(event hardwareRe
 	}
 }
 
+func (sm *hardwareStateMachine) handleControllerFailureEvent(event controllerFailureEvent) {
+	sm.log("handling controller failure event")
+	switch sm.currentState.(type) {
+	case *hardwareOn, *hardwareRestarting, *hardwareStarting:
+		sm.transition(newHardwareStopping(sm.ctx))
```
saxon-milton, line 266: I'm almost wondering if this should be a transition straight to hardwareOff ? If there's an ESP failure then are we really stopping anything; isn't it already stopped ?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -247,6 +259,20 @@ func (sm *hardwareStateMachine) handleHardwareResetRequestEvent(event hardwareRe
...
+	case *hardwareOff:
+		sm.transition(newHardwareStarting(sm.ctx))
```
saxon-milton, line 268: If the hardware is in an off state, and we get a controller failure event, do we want to try transitioning to a hardware starting state ?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -247,6 +259,20 @@ func (sm *hardwareStateMachine) handleHardwareResetRequestEvent(event hardwareRe
...
+	case *hardwareStopping:
```
saxon-milton, line 269: I'm thinking this should go up into the same case as on, restarting and starting.

cmd/oceantv/broadcast_machine.go
```diff
@@ -164,6 +166,24 @@ func (sm *broadcastStateMachine) handleGoodHealthEvent(event goodHealthEvent) er
...
 
+func (sm *broadcastStateMachine) handleControllerFailureEvent(event controllerFailureEvent) error {
+	sm.log("handling controller failure event")
+	switch sm.currentState.(type) {
+	case *directStarting:
+		onFailureClosure(sm.ctx, sm.ctx.cfg, true)(fmt.Errorf("Controller not reporting. Likely caused by low voltage."))
+		sm.transition(newDirectIdle(sm.ctx))
+	case *vidforwardPermanentStarting:
+		onFailureClosure(sm.ctx, sm.ctx.cfg, true)(fmt.Errorf("Controller not reporting. Likely caused by low voltage."))
+		sm.transition(newVidforwardPermanentSlate())
```
saxon-milton, line 177: Maybe we should just go to vidforwardPermanentIdle considering it hasn't even been able to start yet.

cmd/oceantv/broadcast_machine.go
```diff
@@ -164,6 +166,24 @@ func (sm *broadcastStateMachine) handleGoodHealthEvent(event goodHealthEvent) er
 	return nil
 }
 
+func (sm *broadcastStateMachine) handleControllerFailureEvent(event controllerFailureEvent) error {
+	sm.log("handling controller failure event")
+	switch sm.currentState.(type) {
+	case *directStarting:
+		onFailureClosure(sm.ctx, sm.ctx.cfg, true)(fmt.Errorf("Controller not reporting. Likely caused by low voltage."))
```
saxon-milton, line 173: Each onFailureClosure call uses the same error; consider defining it once for use by all of these.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -32,6 +32,16 @@ func newHardwareStarting(ctx *broadcastContext) *hardwareStarting {
 }
 func (s *hardwareStarting) enter() {
 	s.LastEntered = time.Now()
+	if s.cfg.ControllerMAC != 0 {
+		controllerIsOn, err := getDeviceStatus(context.Background(), s.cfg.ControllerMAC, settingsStore)
+		if err != nil {
```
saxon-milton, line 37: Maybe we should just combine these cases. Also, we should probably perform a notification under this circumstance e.g. 
```
if err != nil || !controllerIsOn {
    notifier.Send(
        context.Background(),
        _cfg.SKey,
        "health",
        fmt.Sprintf("Could not get good controller status on hardware startup, status: %v, err: %v",controllerIsOn,err),
    )
    s.bus.publish(controllerFailureEvent{})
    return
}
```

cmd/oceantv/broadcast_states.go
```diff
@@ -633,7 +633,7 @@ func onFailureClosure(ctx *broadcastContext, cfg *BroadcastConfig) func(err erro
 				_cfg.StartFailures++
 				// TODO: make this configurable in config.
 				const maxStartFailures = 3
-				if _cfg.StartFailures >= maxStartFailures {
+				if disableOnFirstFail || _cfg.StartFailures >= maxStartFailures {
 					_cfg.StartFailures = 0
 					_cfg.Enabled = false
 					notifier.Send(
```
saxon-milton, line 639: I'm thinking maybe we could supply the disableOnFirstFail status to the notify message + the last error that occured (which is a cheeky inclusion into this PR).

saxon-milton (changes requested):

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Hmmm... This must have come from where I copied this layout from. Definitely not I wouldn't think

cmd/oceantv/broadcast_machine.go
ao-david: Was there somewhere these types of errors are being defined in a file, couldn't find any others. Otherwise I'll just put it at the tip of this file

cmd/oceantv/broadcast_hardware_machine.go
ao-david: I think that the 2 cases represent different things though? One is a data store error (I think) and the other is a hardware error. Would probably be good to know if we're having data store errors if we are?

saxon-milton: Your handling for both of these cases was the same i.e. to publish a controllerFailureEvent, which I think is fair. 

The suggested code provides the error in the notification message, so if the controllerFailureEvent occurred because of a datastore error, we can be informed of that.

cmd/oceantv/broadcast_machine.go
saxon-milton: Just put at top of this method I reckon. Also, you don't need to use fmt.Errorf considering you're not providing any additional arguments for formatting.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: The controller failure event handling will already notify (through the onfailureclosure call). Is there a good way of handling the error through this event to make a datastore error visible to the notifier to send?

cmd/oceantv/broadcast_machine.go
ao-david: Done.

cmd/oceantv/broadcast_machine.go
ao-david: Done.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Done.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Done.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Done.

cmd/oceantv/broadcast_states.go
ao-david: I have included the disableOnFirstFailure status to the message now. I don't think there is a good way to also send the last error to occur without changing all calls to onfailure closure, so I think that should be a different PR.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: I don't think it matters too much if we notify in both places, but that raises a good question; maybe it'd be better to store a failure message in the event. That way we can get that and use it for the notify of the handling of the event. 

Yep, it's in the code I provided :)

cmd/oceantv/broadcast_states.go
saxon-milton: The error is provided as a parameter to the returned closure already, so you can use that.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Haven't we tried to avoid over notifying recently? I think I'm more inclined at this point to not differentiate between a datastore error and controller not reporting, as we will probably be able to tell when we check the voltage after receiving the notification.

cmd/oceantv/broadcast_states.go
```diff
@@ -688,14 +688,19 @@ func onFailureClosure(ctx *broadcastContext, cfg *BroadcastConfig) func(err erro
...
 						context.Background(),
 						_cfg.SKey,
 						"health",
```
scruzin, line 697: Not for this PR, but let's start using different kinds of messages other than "health" for everything, so we have better notification granularity. All broadcast notification kinds can have the prefix "broadcast-". For example,this one could be "broadcast-hardware" since it is a hardware failure.

scruzin (approved):

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -34,6 +34,18 @@ func newHardwareStarting(ctx *broadcastContext) *hardwareStarting {
 }
 func (s *hardwareStarting) enter() {
 	s.LastEntered = time.Now()
+	if s.cfg.ControllerMAC != 0 {
+		controllerIsOn, err := model.DeviceIsUp(context.Background(), settingsStore, model.MacDecode(s.cfg.ControllerMAC))
+		if err != nil {
+			s.broadcastContext.log("(%s) error getting device status: %v", controllerFailureEvent{}.String(), err)
```
saxon-milton, line 40: You can do ```s.log``` because broadcastContext is embedded :)

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -34,6 +34,18 @@ func newHardwareStarting(ctx *broadcastContext) *hardwareStarting {
...
+			s.bus.publish(controllerFailureEvent{})
+			return
+		} else if !controllerIsOn {
+			s.broadcastContext.log("(%s) controller not responding", controllerFailureEvent{}.String())
```
saxon-milton, line 44: ditto

cmd/oceantv/broadcast_machine.go
```diff
@@ -170,6 +172,26 @@ func (sm *broadcastStateMachine) handleGoodHealthEvent(event goodHealthEvent) er
 	return nil
 }
 
+var errCtrlNotReport = errors.New("Controller not reporting. Likely caused by low voltage.")
```
saxon-milton, line 175: Conventional in go to have lower case errors. We can probably just shorten to "controller not reporting" as well. It's kind of implicit for us what the cause is i.e. low voltage or just broken hehe

saxon-milton (approved):

cmd/oceantv/broadcast_machine.go
scruzin: Besides, we don't really know the cause of why the controller not reporting. It is best to keep speculation out of error messages and report only what we know.

cmd/oceantv/broadcast_states.go
saxon-milton: We should probably define a "NotificationKind" enum so that it's clear what options there are.

scruzin: Good idea. I'll add this to the notify package.

saxon-milton:
closes #114

PR #123, opened 2024-06-13
Check Controller Status before starting hardware

ao-david:
Checking the controller status will prevent the hardware from trying to be started when the controller isn't connected. This will prevent a powered off rig from trying to start and will give a more informative error message.

We should still be adding testing for these functions.


saxon-milton:
Is this a duplicate of #119

saxon-milton:
> Is this a duplicate of #119 ?

Which one should we be looking at ?

davidsutts:
> Is this a duplicate of #119 ?

Yes, not sure how that happened. Ignore this one.

ao-david:
> > Is this a duplicate of #119 ?
> 
> Which one should we be looking at ?

Yes, not sure how this happened. Ignore this PR.

PR #130, opened 2024-06-15
Use background context for token exchange

saxon-milton:
oauth2.NoContext is reported to be deprecated; context.Background() is suggested.


scruzin (approved):
Too easy.

PR #134, opened 2024-06-16
Parameterise auth token URI for broadcast

saxon-milton:
The bucket object containing the required auth tokens for YouTube Live API usage was hardcoded to that belonging to AusOcean's account, however, we now require the ability to change this depending on the YouTube account for which we want to broadcast to.

We're now storing the Account email in the broadcast config, and using this according to a schema for the object name, which a common function, TokenURIFromAccount enforces. We provide this to the YoutTubeBroadcast service which will use it to get the auth token where required.

These changes also affect health checking and chat message posting
so to make things better, we're incorporating this functionality
into the broadcast manager/broadcast services.

### DIFF
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -40,8 +40,8 @@ import (
 	"github.com/ausocean/cloud/cmd/oceantv/broadcast"
 	"github.com/ausocean/cloud/gauth"
 	"github.com/ausocean/cloud/model"
+	"github.com/ausocean/cloud/utils"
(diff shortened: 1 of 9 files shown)


scruzin (approved):
LGTM. 

Not sure why handleChatMessage got moved from broadcast.go, though.

saxon-milton:
> LGTM.
> 
> Not sure why handleChatMessage got moved from broadcast.go, though.

It's just because it would need to use BroadcastService.PostChatMessage and I didn't want to parameterise the BroadcastService over just making it a method of BroadcastService.

Issue #131, opened 2024-06-15
oceantv: allow for broadcasting to accounts other than ausocean


saxon-milton:
This would require that we re-enable token generation, and allow storage of these tokens on a per account basis (as opposed to having only one token the ausocean youtube account uses).

ao-david:
This seems to be working. Although, loooking at the logs of the HVR stream from the weekend, it looks like the social@ausocean.org token might be getting revoked more quickly than the the standard ausocean one?

```
"2024/06/22 01:01:09 (name: HVR Test Broadcast, id: qFIcilpFZV4) could not stop broadcast: could not get broadcast status: get broadcast status error: could not list broadcasts: Get "https://youtube.googleapis.com/youtube/v3/liveBroadcasts?alt=json&id=qFIcilpFZV4&part=status&prettyPrint=false": oauth2: token expired and refresh token is not set"
```

scruzin:
The error message ` "refresh token is not set"`  suggests that it is not refreshing properly, though.

ao-david:
See #158 and #163

scruzin:
Can we close this now @saxon-milton ?

saxon-milton:
> Can we close this now @saxon-milton ?

Yep!

PR #136, opened 2024-06-16
Add token generation button to broadcast page

saxon-milton:
Depends on PR #134

Resolves issue #131 

To associate and authenticate a broadcast with a particular YouTube account, we need to generate a token for the account.

We're providing a button on the broadcast page that performs the generation and stores the account email in the broadcast config, which can be used to later retreive the token for refresh and/or use.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -38,9 +38,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ausocean/cloud/cmd/oceantv/broadcast"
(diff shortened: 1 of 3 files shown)


scruzin (approved):
LGTM.

But, why does oceanbench now need YOUTUBE_SECRETS? Previously, only oceantv required it.

saxon-milton:
> LGTM.
> 
> But, why does oceanbench now need YOUTUBE_SECRETS? Previously, only oceantv required it.

It's because it does the token exchange which requires a google client config, which requires API secrets.

PR #138, opened 2024-06-16
Use model.DeviceIsUp

scruzin:


model/device.go
```diff
@@ -406,8 +406,8 @@ func IsMacAddress(mac string) bool {
 // otherwise. Up status is determined by checking the uptime variable
 // associated with the device. The device is considered to be up if
 // the uptime is less than twice the monitor period.
```
trekhopton, line 408: Should be: 

if the uptime has been updated within the last two monitor periods.

trekhopton (approved):
One small comment but LGTM.

Issue #151, opened 2024-06-20
Bench: Hide advanced features on broadcast page


ao-david:
Whilst we are configuring a more well documented and understandable UI, we should hide away any features that non-ausocean users won't need to touch to get their broadcast working. This should be hidden behind an advanced button or similar behaviour.

scruzin:
This is a P1 because it is customer facing.

PR #152, opened 2024-06-20
Bench/Broadcast: Hide advanced options (#151)

ao-david:
This change hides advanced options on the broadcast page by default. These options can be shown by clicking the advanced options checkbox.

This code also adds a cookie for the advanced option checkbox so that the setting persists across reloads.

The settings that are shown by default:
image


trekhopton (approved):
LGTM! Nice one :D

PR #153, opened 2024-06-20
Bench/Broadcast: Update Broadcast UI

ao-david:
The broadcast page UI has been updated to be more logically ordered, and using bootstrap styling to match the site more seamlessly.

This is just a styling change. (I can separate the PR into commits if that is preferred, but they are all mostly pretty huge due to the indent changing of the formatter.)

image


trekhopton (approved):
Can't say I checked every line but the UI looks good.

Issue #157, opened 2024-06-22
broadcast: improve fix method for permanent live unhealthy state


saxon-milton:
We should take inspiration from the directLiveUnhealthy fix method i.e. have a max number of fix attempts, however instead of emitting a finish event when exceeding the limit we should go to slate. In future we can have an "unhealthy stream" slate, but for now, this'll just be our standard one.

PR #165, opened 2024-06-25
Improve permanent live unhealthy state behaviour

saxon-milton:
Resolves issue #157 

We're now allowing only limited health fixes when in the permanent live unhealthy state. Fixes include hardware reset and requesting the forwarding service for stream (in case problems are on this front).

In the case that we max out our fixes allowance, we transition to a new permanent failure state. In this state we request that the forwarder goes to slate mode. We've also introduced a state to the broadcast object "InFailure" which is true if the broadcast is in a failure state. This is accessible from the broadcast UI. We'll need to uncheck this once we've resolved any issues and would like to resume normal behaviour.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -151,6 +151,7 @@ type BroadcastConfig struct {
 	Transitioning     bool          // If the broadcast is transition from live to slate, or vice versa.
 	StateData         []byte        // States will be marshalled and their data stored here.
 	Account           string        // The YouTube account email that this broadcast is associated with.
+	InFailure         bool          // True if the broadcast is in a failure state.
 }
 
 // SensorEntry contains the information for each sensor.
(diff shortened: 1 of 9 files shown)


cmd/oceanbench/broadcast.go
```diff
@@ -151,6 +151,7 @@ type BroadcastConfig struct {
 	Transitioning     bool          // If the broadcast is transition from live to slate, or vice versa.
 	StateData         []byte        // States will be marshalled and their data stored here.
 	Account           string        // The YouTube account email that this broadcast is associated with.
+	InFailure         bool          // True if the broadcast is in a failure state.
```
scruzin, line 154: `InFailure` means we are in a serious failure state, not just some fleeting failure, right?

If so, I think the term `CriticalFailure` might better the severity of the situation.

scruzin (approved):

cmd/oceanbench/broadcast.go
saxon-milton: We kind of have unhealthy -> infailure -> broadcast disabled. Broadcast disabled is probably more akin to a critical failure because it's at the point where we can't do anything. This normally happens with direct broadcasts because we don't have the option of a slate. I'm personally not super settled on this though, and things will likely get clearer when we add better and more error condition handling; can discuss further.

Issue #158, opened 2024-06-24
TV: Token expiring without refresh token set


ao-david:
When trying to send a broadcast to the network blue channel, we get the following error:
```
2024/06/24 00:49:09 (name: HVR Test Broadcast, id: SH41160z3gg) could not stop broadcast: could not get broadcast status: get broadcast status error: could not list broadcasts: Get "https://youtube.googleapis.com/youtube/v3/liveBroadcasts?alt=json&id=SH41160z3gg&part=status&prettyPrint=false": oauth2: token expired and refresh token is not set
```
This should not happen, as we do not want users to have to reauthenticate regularly

scruzin:
Tokens _should_ expire, just not too soon. 

If users have a one-year subscription to _Cloud Blue Premium_, I would suggest a one-year expiration is appropriate.

ao-david:
It seems based on the token I generated this morning that the tokens have an expiry of a single day or less.

I'm not quite familiar enough with oauth and the golang API to change settings, but have observed the following.
- Running locally (standalone) I have been able to generate tokens with refresh tokens to ensure that these tokens can be used ongoing.
- Running on the current deployment, the generated/saved tokens don't appear to have refresh tokens

scruzin:
There are two types of tokens:

- The access token, which has a short lifespan, permits access to the resource.
- The refresh token, which has a long lifespan, acquires a new access token when the original one expires. 

It sounds like the refreshing may not be happening.

ao-david:
After some back and forth with @saxon-milton about this, we have discovered the following. We didn't have the refresh token in the bucket, meaning that we weren't able to refresh our access token properly. It turns out that this happened because a refresh token will only be returned on the first call to exchange token, any other subsequent calls will only return an access token. Somehow we stored not the first version of the token in the bucket, meaning that we didn't have a refresh token to refresh with.

This was rectified by removing cloudblue (/vidgrind project) access from our youtube connected application settings. This meant that when we re-authenticated we got the refresh token. Meaning this should not be an ongoing issue.

The main issue that we have now (being addressed by #161 and next prs) is that the callback pattern matching is being registered on every call to generateToken. This means that we can only generate a token once on an instance before the instance will crash.

ao-david:
After merging #164 this issue is still occuring, likely due to the token being overwritten instead of refreshed when getting from the oauth.

[...]

PR #167, opened 2024-06-26
Bench: Check for existing token before generating new token (#158)

ao-david:
Broadcast.go now calls AuthToken which first checks if a token for the given user already exists.
- If the token exists, the call returns
- If the token does not exist, the call generates a new token and saves it using GenerateToken.

This will prevent subsequent token generations of tokens which will not have refresh tokens (as only the first token will come with a refresh token).

### DIFF
--- a/cmd/oceantv/broadcast/youtube.go
+++ b/cmd/oceantv/broadcast/youtube.go
@@ -39,9 +39,11 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
+	"cloud.google.com/go/storage"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/googleapi"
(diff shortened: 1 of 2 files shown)


cmd/oceantv/broadcast/youtube.go
```diff
@@ -103,6 +105,24 @@ func GetService(ctx context.Context, scope string, tokenURI string) (*youtube.Se
 	return s, nil
 }
 
+// AuthChannel checks for a current token under the passed tokenURI, and generates one if it does not
+// yet exist.
+func AuthChannel(ctx context.Context, w http.ResponseWriter, r *http.Request, scope, tokenURI string) error {
+	// Don't regenerate a token if one already exists.
+	_, err := getToken(ctx, tokenURI)
+
+	// TODO: Check if token has expired.
```
scruzin, line 114: Don't we need to implement the token expiry check?

scruzin (changes requested):

trekhopton (approved):

cmd/oceantv/broadcast/youtube.go
ao-david: As it turns out, apparently google refresh tokens won't expire unless revoked by the user. Which I don't think we are able to determine with just the token itself.

Instead, we will just need to handle the error properly in the broadcast manager when we encounter a response indicating that the token has been invalidated.

cmd/oceantv/broadcast/youtube.go
```diff
@@ -103,6 +105,23 @@ func GetService(ctx context.Context, scope string, tokenURI string) (*youtube.Se
...
+// AuthChannel checks for a current token under the passed tokenURI, and generates one if it does not
+// yet exist.
+func AuthChannel(ctx context.Context, w http.ResponseWriter, r *http.Request, scope, tokenURI string) error {
+	// Don't regenerate a token if one already exists.
+	_, err := getToken(ctx, tokenURI)
+
+	if err == nil {
+		// Token exists.
+		return nil
+	} else if !errors.Is(err, storage.ErrObjectNotExist) && !errors.Is(err, os.ErrNotExist) {
```
saxon-milton, line 117: Don't need ```else if```, just ```if```.

saxon-milton (approved):

PR #169, opened 2024-06-26
Initial revision

scruzin:
This initial revision of Ocean Center only serves static files, which is sufficient for software upgrades, but not fresh installations.


cmd/oceancenter/main.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/ausocean/cloud/model"
+	"github.com/ausocean/openfish/datastore"
+)
+
+const (
+	projectID = "oceancenter"
+	version   = "v0.1.0"
+)
+
+var (
```
saxon-milton, line 50: We could avoid these globals if we created an ```oceanCenter``` struct with these fields + methods ```indexHandler```, ```installHandler``` etc then created a value of this struct in ```func main```. Then we would just register the methods with ```http```.

cmd/oceancenter/main.go
```diff
@@ -0,0 +1,155 @@
...
+// TODO: Implement this.
+func installHandler(w http.ResponseWriter, r *http.Request) {
+	logRequest(r)
+}
+
+// writeError writes http errors to the response writer, in order to provide more detailed
+// response errors in a concise manner.
+func writeError(w http.ResponseWriter, code int, msg string, args ...interface{}) {
+	errorMsg := "%s: "
+	if msg != "" {
```
saxon-milton, line 131: Probably don't need this check right ? If ```msg == ""``` then ```errorMsg += msg``` just won't do anything ?

cmd/oceancenter/main.go
```diff
@@ -0,0 +1,155 @@
...
+		errorMsg += msg
+	}
+	if len(args) > 0 {
+		errorMsg += ": "
+		errorMsg = fmt.Sprintf(errorMsg, http.StatusText(code), args)
```
saxon-milton, line 136: I think this needs to be ```errorMsg = fmt.Sprintf(errorMsg, http.StatusText(code), args...)```

Also, you could just do this (can still use ```args...``` even if empty)
```
if len(args) > 0 {
    errorMsg += ": "
}

errorMsg = fmt.Sprintf(errorMsg, http.StatusText(code), args...)
```

saxon-milton (changes requested):

cmd/oceancenter/main.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// logRequest logs a request if in debug mode and standalone mode.
+// It does nothing in App Engine mode as App Engine logs requests
+// automatically.
+func logRequest(r *http.Request) {
+	if !(debug || standalone) {
+		return
+	}
+	if r.URL.RawQuery == "" {
```
ao-david, line 150: I think that you can just use this instead:
```golang
log.Println(r.RequestURI)
```

cmd/oceancenter/main.go
```diff
@@ -0,0 +1,155 @@
...
+	// Perform one-time setup or bail.
+	ctx := context.Background()
+	setup(ctx)
+
+	// Serve static files when running locally
+	http.Handle("/s/", http.StripPrefix("/s", http.FileServer(http.Dir("s"))))
+	http.Handle("/dl/", http.StripPrefix("/dl", http.FileServer(http.Dir("dl"))))
+
+	http.HandleFunc("/install", installHandler)
+	http.HandleFunc("/", indexHandler)
```
ao-david, line 84: Are there any other endpoints that should be predefined with TODO messages?

ao-david (changes requested):
I don't think that the writeError function will quite work as expected at the moment.

cmd/oceancenter/main.go
scruzin: Code removed, since not used.

cmd/oceancenter/main.go
```diff
@@ -0,0 +1,155 @@
...
+
+// writeError writes http errors to the response writer, in order to provide more detailed
+// response errors in a concise manner.
+func writeError(w http.ResponseWriter, code int, msg string, args ...interface{}) {
+	errorMsg := "%s: "
+	if msg != "" {
+		errorMsg += msg
+	}
+	if len(args) > 0 {
+		errorMsg += ": "
```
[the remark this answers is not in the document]
scruzin: Code removed.

trekhopton (approved):

cmd/oceancenter/main.go
```diff
@@ -0,0 +1,136 @@
...
+// service defines the properties of our web service.
+type service struct {
+	setupMutex    sync.Mutex
+	settingsStore datastore.Store
+	debug         bool
+	standalone    bool
+}
+
+// app is an instance of our service.
+var app *service = &service{}
```
saxon-milton, line 59: Does this even need to be global? No where else uses it besides main; we could just define it in there.

scruzin: It needs to be global, since in GAE `main` is only called once, and is not called subsequently on warm startups.

saxon-milton: But ```main``` is not returned from right ? ```http.ListenAndServe``` is blocking so the service value will still exist. IIUC for each new instance ```main``` is called.

PR #171, opened 2024-06-26
bench: use site timezone for broadcast, don't expect offset decimal

trekhopton:
This was done because the javascript was being used that interpreted the UTC offset string as a decimal, which caused the time to be off by an 30 minutes. This change should fix that.
The site's timezone can be used to keep things consistent and avoid error from user input.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -97,6 +97,7 @@ type broadcastRequest struct {
 	Settings           Settings         // A struct containing options for some settings that have limited options.
 	Action             string           // Holds value of any button pressed.
 	ListingSecondaries bool             // Are we listing secondary broadcasts?
+	Site               *model.Site
 	commonData
 }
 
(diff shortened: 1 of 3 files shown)


saxon-milton:
Would recommend using a rebase on to master if you need to get up to date as opposed to a merge.

saxon-milton (approved):
Good; I think we've been dealing with timezone issues for a long time :laughing:

ao-david:
Changes look good, but I wonder whether it even makes sense for the timezone to be changeable? Should we just use the site timezone and make it clear to the user that all times are based on the site timezone? I can't imagine many cases where a user would want to schedule a broadcast in a different location based on the timezone of their current location (mainly due to daylight hours etc...

At very least maybe this should also be an advanced feature that gets hidden behind the advanced checkbox?

ao-david (approved):

scruzin:
I think we should always use the site timezone. Otherwise we're simply
asking for trouble.

PR #173, opened 2024-06-26
Improve permanent start/transition error behaviour

### LINKED ISSUE #156: broadcast: re-evaluate permanent starting/transition state timeout behaviour
This is specifically concerned with the behaviour on handling of timeEvents (handleTimeEvent). For example, vidforwardPermanentStarting should follow directStarting's behaviour e.g. on timeout, call onFailureClosure.

Additionally, vidforwardPermanentTransitionSlateToLive on timeout should stay in slate mode. Therefore, on timeout, we don't want to transition from this state but just perform an email notification to inform of the problem.

The behaviour for vidforwardPermanentTransitionLiveToSlate should probably do something similar i.e. perform an email notify, so that we can try and fix the slate mechanism. We should ensure that the camera does power down. The effect from a viewer's perspective is that the broadcast will stay buffering until the problem is remedied

saxon-milton:
Resolves #156

We're making the error handling behaviour a little smarter when it comes to error condition behaviour. For many of these cases, we were transitioning to the idle state, but this isn't great for a permanent broadcast.

Firstly, if we time out transitioning from live to slate, we're now notifying ops to check fordwarder software and staying in the live state.

Secondly, if we time out on permanent starting, we use the onFailure closure which gives us some more chances, but will notify and disable when a limit is reached.

Thirdly, if we time out transitioning from slate to live, we enter the permanent failure state which stays in slate mode and notifies ops that there may be a problem with the hardware.

### DIFF
--- a/cmd/oceantv/broadcast_machine.go
+++ b/cmd/oceantv/broadcast_machine.go
@@ -199,8 +199,17 @@ func (sm *broadcastStateMachine) handleTimeEvent(event timeEvent) {
 			sm.ctx.bus.publish(startEvent{})
 			return
 		}
-	case *vidforwardPermanentStarting, *vidforwardPermanentTransitionLiveToSlate, *vidforwardPermanentTransitionSlateToLive:
-		sm.transitionIfTimedOut(sm.currentState, newVidforwardPermanentIdle(sm.ctx), event)
+	case *vidforwardPermanentTransitionLiveToSlate:
+		withTimeout := sm.currentState.(stateWithTimeout)
+		if withTimeout.timedOut(event.Time) {
+			notifier.Send(
+				context.Background(),
+				sm.ctx.cfg.SKey,
+				"health",
+				"transition from live to slate timed out, staying in live state, check forwarding service",
+			)
+			sm.transition(newVidforwardPermanentLive())
+		}
 		sm.publishHealthEvent(event)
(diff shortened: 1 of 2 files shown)


cmd/oceantv/broadcast_machine.go
```diff
@@ -209,6 +218,22 @@ func (sm *broadcastStateMachine) handleTimeEvent(event timeEvent) {
...
+			onFailureClosure(sm.ctx, sm.ctx.cfg)(errors.New("permanent starting timed out"))
+		}
+	case *vidforwardPermanentTransitionSlateToLive:
+		withTimeout := sm.currentState.(stateWithTimeout)
+		if withTimeout.timedOut(event.Time) {
+			notifier.Send(
+				context.Background(),
+				sm.ctx.cfg.SKey,
+				"health",
+				"transition from slate to live timed out, transitioning to failure slate state",
```
ao-david, line 233: It was mentioned in the PR comment that this would notify that the hardware was probably at fault, but doesn't seem to say that here?

ao-david (approved):

cmd/oceantv/broadcast_machine.go
saxon-milton: Yeah I meant more implicitly.

PR #174, opened 2024-06-28
Bench: Prevent Changing of Youtube Channel (#155)

### LINKED ISSUE #155: Bench: Dissallow insertion of account into save request
Since we started showing the account in the broadcast field, we have also been reading this value to make sure that a save will correctly retain the associated account.

Whilst the field is readonly, a user can edit their request and send a different account value as to what they have access to. This will allow the user to stream to any channel that has already generated a token, even if they don't have access to that channel.

Since we are only changing the account value through the generate token action, we should in all other cases read the account not from the form, but from the datastore to make sure that no malicious requests can be made.

ao-david:
When the broadcast save action is called we will get the previously associated account for the broadcast and update the current config with that value to prevent users from changing this. The value can only be changed when the generate-token action is called. We will also not parse the value from the form to ensure this value can "never" be leaked through.

closes #155

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -219,7 +219,6 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 			CamOff:          r.FormValue("cam-off"),
 			CheckingHealth:  r.FormValue("check-health") == "checking-health",
 			Enabled:         r.FormValue("enabled") == "enabled",
-			Account:         r.FormValue("account"),
 			InFailure:       r.FormValue("in-failure") == "in-failure",
 		},
 		Action:             r.FormValue("action"),
@@ -326,10 +325,15 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 
 		// Store the account email in the broadcast config.
 		cfg.Account = profile.Email
-
-		// Fallthrough to save broadcast configuration.
-		fallthrough
+		err := saveBroadcast(ctx, &req.CurrentBroadcast)
+		if err != nil {
+			reportError(w, r, req, "could not save broadcast: %v", err)
+			return
+		}
 	case broadcastSave:
+		// If we haven't just generated a token we should keep the same account
+		// that the config previously had.
+		cfg.Account = getExistingAccount(req.BroadcastVars, cfg)
 		err := saveBroadcast(ctx, &req.CurrentBroadcast)
 		if err != nil {
 			reportError(w, r, req, "could not save broadcast: %v", err)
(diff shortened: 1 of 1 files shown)


cmd/oceanbench/broadcast.go
```diff
@@ -326,10 +325,15 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
...
-		fallthrough
+		err := saveBroadcast(ctx, &req.CurrentBroadcast)
+		if err != nil {
+			reportError(w, r, req, "could not save broadcast: %v", err)
+			return
+		}
 	case broadcastSave:
+		// If we haven't just generated a token we should keep the same account
+		// that the config previously had.
+		cfg.Account = getExistingAccount(req.BroadcastVars, cfg)
```
scruzin, line 336: Is getExistingAccount guaranteed to never fail?

cmd/oceanbench/broadcast.go
```diff
@@ -458,6 +462,17 @@ func loadExistingSettings(r *http.Request, req *broadcastRequest) (bool, error)
 	return true, nil
 }
 
+// getExistingAccount will return the current associated account of the broadcast with the current config
+// name. This should be used to ensure that the associated account is only updated using the generate token method.
+// If no broadcast/account is found, then an empty string will be returned
+func getExistingAccount(broadcasts []model.Variable, cfg *BroadcastConfig) string {
+	cfg, err := broadcastFromVars(broadcasts, cfg.Name)
+	if err != nil {
+		return ""
```
scruzin, line 471: Log upon failure.

scruzin (changes requested):

ao-david:
I have just realised that this method is flawed. If the user changes the name of the broadcast, the broadcastFromVars function will be unable to find the old variable. We currently have no way of statefully finding the broadcast that has just been updated. It also looks like OceanTV will just lose track of an old config (broadcast variable) if the user changes the site name. This means that every time a user changes the name of a broadcast that a new variable is created as far as I can tell.

For now this means that we will just lose the value of the current YouTube channel when we change the name of a stream. Eventually this will be fixed when we associate the channel with the site rather than the broadcast.

ao-david:
> I have just realised that this method is flawed. If the user changes the name of the broadcast, the broadcastFromVars function will be unable to find the old variable. We currently have no way of statefully finding the broadcast that has just been updated. It also looks like OceanTV will just lose track of an old config (broadcast variable) if the user changes the site name. This means that every time a user changes the name of a broadcast that a new variable is created as far as I can tell.
> 
> For now this means that we will just lose the value of the current YouTube channel when we change the name of a stream. Eventually this will be fixed when we associate the channel with the site rather than the broadcast.

see #175

saxon-milton (approved):

PR #179, opened 2024-06-29
Slate on hardware failure for permanent live states

### LINKED ISSUE #54: broadcast: stay in slate if hardware failure
If during transition from slate to live we cannot power on the hardware we currently try to do the whole broadcast creation and start process again. This behaviour could be more intelligent. We should instead just stay in the slate state. This is more sensible.

saxon-milton:
closes #54 

We're now handling hardware failure events when we're in the permanent live/live unhealthy and slate to live states. Under these circumstances, we should stay in slate. We can do this by entering the vidforwardPermanentInFailure state in which case we wait until admin has taken the broadcast out of this state i.e. maybe fixed the problem.

### DIFF
--- a/cmd/oceantv/broadcast_states.go
+++ b/cmd/oceantv/broadcast_states.go
@@ -22,6 +22,10 @@ type broadcastContext struct {
 func (ctx *broadcastContext) log(msg string, args ...interface{}) {
 	logForBroadcast(ctx.cfg, msg, args...)
 }
+func (ctx *broadcastContext) logAndNotify(msg string, args ...interface{}) {
+	logForBroadcast(ctx.cfg, msg, args...)
+	notifier.Send(context.Background(), ctx.cfg.SKey, "health", fmtForBroadcastLog(ctx.cfg, msg, args...))
+}
 
 type state interface {
 	enter()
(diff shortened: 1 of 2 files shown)


scruzin (commented):
Is a manual (operator) intervention required to exit the vidforwardPermanentInFailure state? If so, how will the operator know?

saxon-milton:
> Is a manual (operator) intervention required to exit the vidforwardPermanentInFailure state? If so, how will the operator know?

Yes, when we enter this state there will be a notification email. The operator would then have to investigate the issue and try to ameliorate it (or wait until hardware is healthy i.e. sufficient voltage). Once hardware is healthy, we can take it out of this state. One step beyond this would be to periodically check hardware health and capacity to stream, under which conditions we can automatically go back to whatever state we should be in.

scruzin (approved):

cmd/oceantv/broadcast_states.go
```diff
@@ -22,6 +22,10 @@ type broadcastContext struct {
 func (ctx *broadcastContext) log(msg string, args ...interface{}) {
 	logForBroadcast(ctx.cfg, msg, args...)
 }
+func (ctx *broadcastContext) logAndNotify(msg string, args ...interface{}) {
```
scruzin, line 25: Pass in _notify.Kind_ to logAndNotify.

scruzin (changes requested):

PR #186, opened 2024-06-29
Added type Kind

scruzin:
This is a simple change to improve typesafe usage of the `notify` package, which is backwards compatible with clients, providing the kind is supplied as literal.

I considered making `Kind` an interface (modeled on `Error`), but that would not be backwards compatible. Further, it is not clear that `Kind` needs to be more than a string.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -41,6 +41,9 @@ type Notifier struct {
 	privateKey string     // Public key for accessing MailJet API.
 }
 
+// Kind represents a kind of notification.
+type Kind string
+
 // Init initializes a notifier with the supplied options. See
 // WithSender, WithRecipient, WithFilter, WithStore and WithSecrets
 // for a description of the various options. Secrets are required to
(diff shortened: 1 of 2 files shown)


notify/notify.go
```diff
@@ -41,6 +41,9 @@ type Notifier struct {
 	privateKey string     // Public key for accessing MailJet API.
 }
 
+// Kind represents a kind of notification.
+type Kind string
```
saxon-milton, line 45: Should we add a bunch of recommended kinds i.e.
```
const (
    BroadcastService  Kind = "broadcast-service"
    BroadcastAPI        Kind = "broadcast-api"
    BroadcastNetwork Kind = "broadcast-network"
// etc....
)
```

saxon-milton (approved):

notify/notify.go
scruzin: I think we should leave this to clients and keep the notify package agnostic.

saxon-milton: Fair enough, it's a bit harder to enforce a particular scheme then however, but maybe that's OK.

PR #190, opened 2024-06-29
Merge broadcast man save and config update transaction

### LINKED ISSUE #177: broadcast: OceanBroadcastManager.SaveBroadcast should use transaction
Every other update of the broadcast config uses a transaction so that we can ensure things are synchronised properly, we should also do this with SaveBroadcast from the OceanBroadcastManager

saxon-milton:
closes #177 

We're having OceanBroadcastManager.Save wrap
updateConfigWithTransaction, which achieves several things.

Firstly, if everything uses OceanBroadcastManager.Save for updating broadcast config fields, then we'll have everything inherently synchronised propertly.

Secondly, this should greatly reduce boilerplate associated with using updateConfigWithTransaction for every modifification of the broadcast config.

### DIFF
--- a/cmd/oceantv/broadcast_manager.go
+++ b/cmd/oceantv/broadcast_manager.go
@@ -48,7 +48,7 @@ type BroadcastManager interface {
 		onSuccess func(),
 		onFailure func(error))
 	StopBroadcast(ctx Ctx, cfg *Cfg, store Store, svc BroadcastService) error
-	SaveBroadcast(ctx Ctx, cfg *Cfg, store Store) error
+	Save(ctx Ctx, update func(*BroadcastConfig)) error
 
 	// HandleStatus checks the status of a broadcast and would perform any
 	// necessary actions based on this status. For example, if the broadcast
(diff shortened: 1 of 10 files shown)


cmd/oceantv/broadcast_manager.go
```diff
@@ -177,11 +172,26 @@ func (m *OceanBroadcastManager) StopBroadcast(ctx Ctx, cfg *Cfg, store Store, sv
...
-	return saveBroadcast(ctx, cfg, store, m.log)
+// Save performs broadcast configuration update operations.
+// If ctx is nil, the background context will be used.
+//
+// update allows for the update of specific fields. After this update takes
+// place, the config we currently point at will be updated with any changes
+// that we have applied, and with anything from the store before the update.
+// If this is nil, the config currently in store will be replaced.
+func (m *OceanBroadcastManager) Save(ctx Ctx, update func(_cfg *Cfg)) error {
+	_ctx := context.Background()
```
scruzin, line 183: Why set _ctx to Background, only to overwrite it if ctx is nil? Why not just the following?

```
	if ctx != nil {
		ctx = context.Background()
	}
```

scruzin (approved):
One comment. Otherwise LGTM.

cmd/oceantv/broadcast_manager.go
saxon-milton: It gets overwritten if ctx is _not_ nil i.e. if the caller has provided a context then we use that, otherwise we just use context.Background().

scruzin: That's exactly what my code snippet attempt to do, i.e., use ctx unless it is nil, in which case use the background context.

scruzin: Actually, I meant this:

```
if ctx == nil {
		ctx = context.Background()
	}
```

saxon-milton: Oh yeah, you're right

PR #191, opened 2024-06-29
Use notify.Kind in logAndNotify

saxon-milton:
We're increasing the specificity of notifications by having the kind being passed into the logAndNotify method.

We've also introduced a kind enum with values relevant to broadcasting we can use across the oceantv functionality.


scruzin (approved):
LGTM

PR #199, opened 2024-07-01
Bench: Add function to autofill controller configuration (#195)

### LINKED ISSUE #195: Bench: Auto Configuration of Devices
Since we generally will always use the same base variables, and some other parts of the device configuration, we should be able to configure these automatically, reducing the time taken to set up a new device.

This could include variables, sensors, actuators, configuration parameters, as well as variable types.

ao-david:
This function takes in the minimal data which describes a unique controller and populates all the other default fields to their default values.

In the future this functionality will be extended to all device types.

closes #195

### DIFF
--- a/cmd/oceanbench/default.go
+++ b/cmd/oceanbench/default.go
@@ -1,0 +1,135 @@
+/*
+NAME
+  This file contains default values used for variables across different
+	device types.
+
+AUTHORS
+  David Sutton <davidsutton@ausocean.org
+
+LICENSE
+  Copyright (C) 2018-2024 the Australian Ocean Lab (AusOcean)
+
+  This file is part of Ocean Bench. Ocean Bench is free software: you can
+  redistribute it and/or modify it under the terms of the GNU
+  General Public License as published by the Free Software
+  Foundation, either version 3 of the License, or (at your option)
+  any later version.
+
+  Ocean Bench is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  in gpl.txt.  If not, see
+  <http://www.gnu.org/licenses/>.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ausocean/cloud/model"
+	"github.com/ausocean/openfish/datastore"
+)
+
+// setupDefaultController takes the minimum settings required to define a Device and the settings store
+// and adds all the default variables, sensors, and actuators for a controller.
+//
+// Args:
+// - skey: Site Key of the new controller.
+// - name: Name of the device (Mutable).
+// - MAC: MAC address of the device optionally with colons (Immutable).
+// - SSID: SSID of the network for the controller to connect.
+// - password: password of the network.
+// - lat, long: latitude and longitude of the device.
+func setupDefaultController(ctx context.Context, skey int64, name, MAC, SSID, password string, lat, long float64, settingsStore datastore.Store) error {
+	// Check that the passed MAC is valid.
+	if !model.IsMacAddress(MAC) {
+		return model.ErrInvalidMACAddress
+	}
+
+	// Create the Controller with default settings.
+	controller := model.Device{
+		Skey:          skey,
+		Mac:           model.MacEncode(MAC),
+		Name:          name,
+		Inputs:        "A0,X50,X51,X60,X10",
+		Outputs:       "D14,D15,D16",
+		Wifi:          SSID + "," + password,
+		MonitorPeriod: 60,
+		ActPeriod:     60,
+		Type:          "Controller",
+		Latitude:      lat,
+		Longitude:     long,
+		Enabled:       true,
+	}
+
+	// Save the device to the datastore.
+	err := model.PutDevice(ctx, settingsStore, &controller)
+	if err != nil {
+		return fmt.Errorf("unable to save controller to datastore: %w", err)
+	}
+
+	// Default variables used for a controller.
+	var variables = []model.Variable{
+		{Name: "AlarmNetwork", Value: "10"},
+		{Name: "AlarmPeriod", Value: "5"},
+		{Name: "AlarmRecoveryVoltage", Value: "840"},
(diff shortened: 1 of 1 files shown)


scruzin (commented):
Suggest calling this defaults.go (plural)

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+		sensor.Mac = model.MacEncode(MAC)
+		err = model.PutSensorV2(ctx, settingsStore, &sensor)
+		if err != nil {
+			return fmt.Errorf("unable to put sensor %s: %w", sensor.Name, err)
+		}
+	}
+
+	// Default Actuators for a controller.
+	var actuators = []model.ActuatorV2{
+		{Name: "Empty 1", Var: "Power1", Pin: "D16"},
```
scruzin, line 120: Name them Power xxx instead of Empty xxx,

scruzin (changes requested):
Minor changes.

trekhopton (changes requested):

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+	// Put all the sensors into the datastore.
+	for _, sensor := range sensors {
+		sensor.Mac = model.MacEncode(MAC)
+		err = model.PutSensorV2(ctx, settingsStore, &sensor)
+		if err != nil {
+			return fmt.Errorf("unable to put sensor %s: %w", sensor.Name, err)
+		}
+	}
+
+	// Default Actuators for a controller.
```
trekhopton, line 118: Don't need a capital for actuators.

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+	// Save the device to the datastore.
+	err := model.PutDevice(ctx, settingsStore, &controller)
+	if err != nil {
+		return fmt.Errorf("unable to save controller to datastore: %w", err)
+	}
+
+	// Default variables used for a controller.
+	var variables = []model.Variable{
+		{Name: "AlarmNetwork", Value: "10"},
+		{Name: "AlarmPeriod", Value: "5"},
```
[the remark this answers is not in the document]
trekhopton: If I'm not mistaken this is fine as 5 seconds because it determines how long the ESP will power off the router in case of a network alarm.
https://netreceiver.appspot.com/help/esp8266

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+	controller := model.Device{
+		Skey:          skey,
+		Mac:           model.MacEncode(MAC),
+		Name:          name,
+		Inputs:        "A0,X50,X51,X60,X10",
+		Outputs:       "D14,D15,D16",
+		Wifi:          SSID + "," + password,
+		MonitorPeriod: 60,
+		ActPeriod:     60,
+		Type:          "Controller",
```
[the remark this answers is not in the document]
ao-david: I think this is worse, because this doesn't really link the changes in that other file to this one anyway. Ie if the index changes then this will change. But perhaps they can be handled better in set.go

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
+/*
+NAME
```
saxon-milton, line 2: This is description, not name.

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Create the Controller with default settings.
+	controller := model.Device{
+		Skey:          skey,
+		Mac:           model.MacEncode(MAC),
+		Name:          name,
+		Inputs:        "A0,X50,X51,X60,X10",
+		Outputs:       "D14,D15,D16",
+		Wifi:          SSID + "," + password,
+		MonitorPeriod: 60,
```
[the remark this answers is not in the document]
saxon-milton: Agreed

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+// - password: password of the network.
+// - lat, long: latitude and longitude of the device.
+func setupDefaultController(ctx context.Context, skey int64, name, MAC, SSID, password string, lat, long float64, settingsStore datastore.Store) error {
+	// Check that the passed MAC is valid.
+	if !model.IsMacAddress(MAC) {
+		return model.ErrInvalidMACAddress
+	}
+
+	// Create the Controller with default settings.
+	controller := model.Device{
```
saxon-milton, line 55: Probably doesn't have to be in this change, but maybe we should add a "constructor" in model for Device.

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// Save the device to the datastore.
+	err := model.PutDevice(ctx, settingsStore, &controller)
+	if err != nil {
+		return fmt.Errorf("unable to save controller to datastore: %w", err)
+	}
+
+	// Default variables used for a controller.
+	var variables = []model.Variable{
```
saxon-milton, line 77: Now that I see this, maybe we should have a controller.go file under model that provides a factory to provide a controller model.Device. We can also have a function that returns these default variables for a "controller" device.

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+	// Put all the variables into the datastore.
+	for _, variable := range variables {
+		err = model.PutVariable(ctx, settingsStore, skey, MAC+"."+variable.Name, variable.Value)
+		if err != nil {
+			return fmt.Errorf("unable to put variable %s.%s: %w", MAC, variable.Name, err)
+		}
+	}
+
+	// Default sensors used for a controller.
+	var sensors = []model.SensorV2{
```
saxon-milton, line 101: Same deal as variables

cmd/oceanbench/default.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, sensor := range sensors {
+		sensor.Mac = model.MacEncode(MAC)
+		err = model.PutSensorV2(ctx, settingsStore, &sensor)
+		if err != nil {
+			return fmt.Errorf("unable to put sensor %s: %w", sensor.Name, err)
+		}
+	}
+
+	// Default Actuators for a controller.
+	var actuators = []model.ActuatorV2{
```
saxon-milton, line 119: Ditto

ao-david:
This has been closed in favour of a generalised version.

PR #202, opened 2024-07-02
Model: Move device types to model

ao-david:
Device type constants are now stored in model/device.go and exported.


trekhopton (approved):
Nice.

PR #203, opened 2024-07-02
Model: NewDevice() func

ao-david:
This change adds a NewDevice method which will accept function options (DeviceOption) to generate a new device using defaults.

This implementation relies heavily on the use of maps which should hopefully allow these values to be more loosely coupled from the implementation, allowing for easier and more clear modification, however, these maps could be replaced by the switch statement if that was preferable.

depends on #202 

Starts addressing issue #195

### DIFF
--- a/model/device.go
+++ b/model/device.go
@@ -93,6 +93,146 @@ const (
 	DevTypeTest       = "Test"
 )
 
+// Consts containing defaults for all device configs.
+const (
+	defaultActPeriod int64 = 60
+	defaultMonPeriod int64 = 60
+)
+
+// Consts containing defaults for a controller config.
+const (
+	DefaultControllerInputs  string = "A0,X50,X51,X60,X10"
+	DefaultControllerOutputs string = "D14,D15,D16" // Peripheral Power.
+)
+
+// ControllerDefaultVars is a map which relates controller variable names to their default values
+var DefaultControllerVars = map[string]string{
+	"AlarmNetwork":         "10",
+	"AlarmPeriod":          "5",     // 5 seconds.
+	"AlarmRecoveryVoltage": "840",   // 25.4545 V.
+	"AlarmVoltage":         "825",   // 25 V.
+	"AutoRestart":          "600",   // 10 minutes.
+	"Power1":               "false", // initialise to OFF.
+	"Power2":               "false", // initialise to OFF.
+	"Power3":               "false", // initialise to OFF.
+	"Pulses":               "3",     // 3 pulses per cycle.
+	"PulseWidth":           "2",     // 1s ON, 1s OFF.
+	"PulseCycle":           "30",    // 30s between flashes.
+	"PulseSuppress":        "false",
+}
+
+// ControllerDefaultSensors contains a slice of the default sensors for a controller device.
+var DefaultControllerSensors = []SensorV2{
+	{Name: "Battery Voltage", Pin: "A0", Quantity: "DCV", Func: "scale", Args: "0.0289", Units: "V", Format: "round1"},
+	{Name: "Analog Value", Pin: "X10", Quantity: "OTH", Func: "none", Format: "round1"},
+	{Name: "Air Temperature", Pin: "X50", Quantity: "MWH", Func: "linear", Args: "0.1,-273.15", Units: "C", Format: "round1"},
+	{Name: "Humidity", Pin: "X51", Quantity: "MMB", Func: "scale", Args: "0.1", Units: "%", Format: "round2"},
+	{Name: "Water Temperature", Pin: "X60", Quantity: "MTW", Func: "linear", Args: "0.1,-273.15", Units: "C", Format: "round1"},
+}
+
+// ControllerDefaultActs contains a slice of the default actuators for a controller device.
+var DefaultControllerActs = []ActuatorV2{
+	{Name: "Device 1", Var: "Power1", Pin: "D16"},
+	{Name: "Device 2", Var: "Power2", Pin: "D14"},
+	{Name: "Device 3", Var: "Power3", Pin: "D15"},
+}
+
+// DeviceOption is a functional option to be supplied to NewDevice.
+type DeviceOption func(context.Context, datastore.Store, *Device) error
+
+// WithInputs creates a device with the given inputs.
+func WithInputs(inputs string) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
+		dev.Inputs = inputs
+		return nil
+	}
+}
+
+// WithOutputs creates a device with the given outputs.
+func WithOutputs(outputs string) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
+		dev.Outputs = outputs
+		return nil
+	}
+}
+
+// WithVariables creates the given variables for the device.
+func WithVariables(variables map[string]string) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
+		// Put all variables into the datastore.
+		for varName, varVal := range variables {
+			err := PutVariable(ctx, store, dev.Skey, dev.MAC()+"."+varName, varVal)
+			if err != nil {
+				return fmt.Errorf("unable to put variable: %w", err)
+			}
+		}
+		return nil
+	}
+}
+
+// WithSensors creates the given sensors for the device.
+func WithSensors(sensors []SensorV2) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
+		// Put all sensors into the datastore.
+		for _, sensor := range sensors {
+			sensor.Mac = dev.Mac
+			err := PutSensorV2(ctx, store, &sensor)
+			if err != nil {
+				return fmt.Errorf("unable to put sensor %s: %w", sensor.Name, err)
+			}
+		}
+		return nil
+	}
+}
+
+// WithActuators creates the given actuators for the device.
+func WithActuators(acts []ActuatorV2) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
+		// Put all actuators into the datastore.
+		for _, act := range acts {
+			act.Mac = dev.Mac
+			err := PutActuatorV2(ctx, store, &act)
+			if err != nil {
+				return fmt.Errorf("unable to put actuator %s: %w", act.Name, err)
+			}
+		}
+		return nil
+	}
+}
+
+// NewDevice creates a new device with the given parameters, and options.
+//
+// NOTE: Options are applied in order and any order dependent options should be
+// passed in order of execution.
+func NewDevice(ctx context.Context, store datastore.Store, skey int64, name, MAC string, options ...DeviceOption) (*Device, error) {
+	// Create a device with common config parameters.
+	dev := &Device{
+		Skey:          skey,
+		Name:          name,
+		Mac:           MacEncode(MAC),
+		ActPeriod:     defaultActPeriod,
+		MonitorPeriod: defaultMonPeriod,
+		Enabled:       true,
+	}
+
+	// Apply the functional options.
+	var err error
+	for i, opt := range options {
+		err = opt(ctx, store, dev)
+		if err != nil {
+			return nil, fmt.Errorf("unable to apply option[%d]: %w", i, err)
+		}
+	}
+
+	// Put the device into the datastore.
+	err = PutDevice(ctx, store, dev)
+	if err != nil {
+		return nil, fmt.Errorf("unable to put device into settings store: %w", err)
+	}
+
+	return dev, nil
+}
+
 // Encode serializes a Device into tab-separated values.
 func (dev *Device) Encode() []byte {
 	return []byte(fmt.Sprintf("%d\t%d\t%d\t%s\t%s\t%s\t%s\t%d\t%d\t%d\t%s\t%s\t%s\t%f\t%f\t%t\t%d",


model/device.go
```diff
@@ -93,6 +94,138 @@ const (
...
+	defaultMonPeriod int64 = 60
+)
+
+// Consts containing defaults for a controller config.
+const (
+	defaultControllerInputs  string = "A0,X50,X51,X60,X10"
+	defaultControllerOutputs string = "D14,D15,D16" // Peripheral Power.
+)
+
+// controllerDefaultVars is a map which relates variable names to their values for
```
trekhopton, line 109: Perhaps "controllerDefaultVars is a map which relates controller variable names to their default values" would be more concise.

model/device.go
```diff
@@ -93,6 +94,138 @@ const (
...
+	{Name: "Battery Voltage", Pin: "A0", Quantity: "DCV", Func: "scale", Args: "0.0289", Units: "V", Format: "round1"},
+	{Name: "Analog Value", Pin: "X10", Quantity: "OTH", Func: "none", Format: "round1"},
+	{Name: "Air Temperature", Pin: "X50", Quantity: "MWH", Func: "linear", Args: "0.1,-273.15", Units: "C", Format: "round1"},
+	{Name: "Humidity", Pin: "X51", Quantity: "MMB", Func: "scale", Args: "0.1", Units: "%", Format: "round2"},
+	{Name: "Water Temperature", Pin: "X60", Quantity: "MTW", Func: "linear", Args: "0.1,-273.15", Units: "C", Format: "round1"},
+}
+
+// controllerDefaultActs contains a slice of the default actuators for a controller device.
+const controllerDefaultActs = []model.ActuatorV2{
+	{Name: "Empty 1", Var: "Power1", Pin: "D16"},
```
trekhopton, line 135: Please make sure you address the comment from the previous PR too. Alan suggested that these be named Power 1, Power 2 etc. instead of Empty 1, Empty 2 etc. which I think are better default names too.

model/device.go
```diff
@@ -93,6 +94,138 @@ const (
...
+var defaultVars = map[string](map[string]string){
+	DevTypeController: controllerDefaultVars,
+}
+
+// defaultSensors contains a map between the device type and the default sensors for that type.
+var defaultSensors = map[string]([]model.SensorV2){
+	DevTypeController: controllerDefaultSensors,
+}
+
+// defaultActs contains a map between the device type and the default sensors for that type.
```
trekhopton, line 150: sensors -> actuators.

model/device.go
```diff
@@ -93,6 +94,138 @@ const (
...
+	for _, actuator := range acts {
+		actuator.Mac = model.MacEncode(MAC)
+		err = model.PutActuatorV2(ctx, settingsStore, &actuator)
+		if err != nil {
+			return nil, fmt.Errorf("unable to put actuator %s: %w", actuator.Name, err)
+		}
+	}
+
+	return dev, nil
+
```
trekhopton, line 226: Remove blank line.

trekhopton (changes requested):
Nice. Few comments.

model/device.go
ao-david: I was using empty to indicate clearly that the power had nothing attached to it, and that it should probably be updated once a device has been connected to it. If it is labeled Power xxx, a user will likely not rename it, and then forget which device is connected to which power.

I think calling it something like empty by default is a good way to encourage users to rename them once they connect a peripheral.

scruzin: Suggest naming them "Device 1", "Device 2", so users know they correspond to devices.

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+)
+
+// controllerDefaultVars is a map which relates controller variable names to their default values
+var controllerDefaultVars = map[string]string{
+	"AlarmNetwork":         "10",
+	"AlarmPeriod":          "5",     // 5 seconds.
+	"AlarmRecoveryVoltage": "840",   // 25.4545 V.
+	"AlarmVoltage":         "825",   // 25 V.
+	"AutoRestart":          "600",   // 10 minutes.
+	"Power":                "false", // initialise to OFF.
```
scruzin, line 115: Power is an obsolete var. It has been replaced by Power1, Power2 and Power3.

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to apply option[%d]: %w", i, err)
+		}
+	}
+
+	switch dev.Type {
+	case DevTypeController:
+		dev.Inputs = defaultControllerInputs
+		dev.Outputs = defaultControllerOutputs
+	}
```
scruzin, line 185: Add default case with `panic("Unxpected dev.Type in NewDevice")`

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+
+	// Put the device into the datastore.
+	err = PutDevice(ctx, settingsStore, dev)
```
scruzin, line 188: `settingsStore` is at the application layer, not at the model layer. Pass in the `store` instead.

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+}
+
+// DeviceOption is a functional option to be supplied to NewDevice.
+type DeviceOption func(*Device) error
+
+// NewDevice creates a new device with the given parameters, and options.
+//
+// NOTE: Options are applied in order and any order dependent options should be
+// passed in order of execution.
+func NewDevice(ctx context.Context, skey int64, name, MAC string, settingsStore datastore.Store, options ...DeviceOption) (*Device, error) {
```
scruzin, line 160: Rename `settingsStore` to `store`, and make it the 2nd param for consistency with other model methods.

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+	// Create an empty device.
+	dev := &Device{}
```
scruzin, line 162: Initialize with common params, i.e.,

```
dev := &Device{
	Skey: skey,
	Name: name,
        ...
}
```

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+	for i, opt := range options {
+		err = opt(dev)
+		if err != nil {
+			return nil, fmt.Errorf("unable to apply option[%d]: %w", i, err)
+		}
+	}
+
+	switch dev.Type {
+	case DevTypeController:
+		dev.Inputs = defaultControllerInputs
```
scruzin, line 183: This also could be done with options, I.e, `WithInputs` and `WithOuputs`, and exporting the default inputs and outputs. There there is no device-specific code inside `NewDevice`.

For example:

```
devController, err = model.NewDevice(... model.WithInputs(model.DefaultControllerInputs) ...)

devCamera, err = model.NewDevice( ... model.WithInputs(model.DefaultCameraInputs) ...)

devHydrophone, err = model.NewDevice( ... model.WithInputs(model.DefaultHydrophoneInputs) ...)


```

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+	}
+
+	// Put the device into the datastore.
+	err = PutDevice(ctx, settingsStore, dev)
+	if err != nil {
+		return nil, fmt.Errorf("unable to put device into settings store: %w", err)
+	}
+
+	varMap := defaultVars[dev.Type]
+	sensors := defaultSensors[dev.Type]
```
scruzin, line 194: Ditto, sensors and actuators can be initialized with `WithSensors` and `WithActuators`.

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+	for _, sensor := range sensors {
+		sensor.Mac = MacEncode(MAC)
+		err = PutSensorV2(ctx, settingsStore, &sensor)
+		if err != nil {
+			return nil, fmt.Errorf("unable to put sensor %s: %w", sensor.Name, err)
+		}
+	}
+
+	// Put all the actuators into the datastore.
+	for _, actuator := range acts {
```
scruzin, line 215: actuator -> act

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
 
+// Consts containing defaults for all device configs.
+const (
+	defaultActPeriod int64 = 60
+	defaultMonPeriod int64 = 60
+)
+
+// Consts containing defaults for a controller config.
+const (
+	defaultControllerInputs  string = "A0,X50,X51,X60,X10"
```
scruzin, line 104: Export these. See below for explanation.

scruzin (changes requested):

model/device.go
```diff
@@ -93,6 +93,136 @@ const (
...
+	"AlarmNetwork":         "10",
+	"AlarmPeriod":          "5",     // 5 seconds.
+	"AlarmRecoveryVoltage": "840",   // 25.4545 V.
+	"AlarmVoltage":         "825",   // 25 V.
+	"AutoRestart":          "600",   // 10 minutes.
+	"Power":                "false", // initialise to OFF.
+	"Pulses":               "3",     // 3 pulses per cycle.
+	"PulseWidth":           "2",     // 1s ON, 1s OFF.
+	"PulseCycle":           "30",    // 30s between flashes.
+	"Suppress":             "false",
```
trekhopton, line 119: Suppress is an obsolete var too, it's PulseSuppress. The defaults are in the config guide: https://docs.google.com/document/d/1xdS0qk25hFhkAivB2wlzd5U0ZxGEyk_avg-NGPYXtc4/edit

model/device.go
trekhopton, line 115: Add Power1, Power2 and Power3 set to false.

ao-david:
All comments addressed. Now using WithInputs, WithOutputs, WithVariables, WithSensors, WithActuators

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+		err = opt(ctx, store, dev)
+		if err != nil {
+			return nil, fmt.Errorf("unable to apply option[%d]: %w", i, err)
+		}
+	}
+
+	// Put the device into the datastore.
+	err = PutDevice(ctx, store, dev)
+	if err != nil {
+		return nil, fmt.Errorf("unable to put device into settings store: %w", err)
```
scruzin, line 230: settings store -> store

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+		}
+		return nil
+	}
+}
+
+// NewDevice creates a new device with the given parameters, and options.
+//
+// NOTE: Options are applied in order and any order dependent options should be
+// passed in order of execution.
+func NewDevice(ctx context.Context, store datastore.Store, skey int64, name, MAC string, options ...DeviceOption) (*Device, error) {
```
scruzin, line 207: NewDevice should also take the device type as a required param.

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+	// Create a device with common config parameters.
+	dev := &Device{
+		Skey:          skey,
+		Name:          name,
+		Mac:           MacEncode(MAC),
+		ActPeriod:     defaultActPeriod,
```
scruzin, line 213: Include Type as passed in.

scruzin (changes requested):

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
 
+// Consts containing defaults for all device configs.
+const (
+	defaultActPeriod int64 = 60
+	defaultMonPeriod int64 = 60
+)
+
+// Consts containing defaults for a controller config.
+const (
+	DefaultControllerInputs  string = "A0,X50,X51,X60,X10"
```
saxon-milton, line 104: It might be worth commenting what exactly these are. 

It might even be worth having a "string enum" to define these, and then have a factory to provide the default set e.g.

```
type Input string

type ExternalInput Input

const (
    SurfaceTemperature ExternalInput = "X50"
    Humidity ExternalInput = "X51"
    WaterTemperature ExternalInput = "X60"
    // etc.... 
)

type AnalogInput Input

const (
    BatteryVoltage AnalogInput = "A0"
    // etc ...
)

func defaultInputs() []Input {
    return []Input{BatteryVoltage,SurfaceTemperature,Humidity,...}
}
```

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+	DefaultControllerOutputs string = "D14,D15,D16" // Peripheral Power.
```
saxon-milton, line 105: Ditto for outputs

model/device.go
saxon-milton, line 104: Also, where are these used ?

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+	"Power2":               "false", // initialise to OFF.
+	"Power3":               "false", // initialise to OFF.
+	"Pulses":               "3",     // 3 pulses per cycle.
+	"PulseWidth":           "2",     // 1s ON, 1s OFF.
+	"PulseCycle":           "30",    // 30s between flashes.
+	"PulseSuppress":        "false",
+}
+
+// ControllerDefaultSensors contains a slice of the default sensors for a controller device.
+var DefaultControllerSensors = []SensorV2{
```
saxon-milton, line 125: I almost think each one of these should have a factory function. e.g.

```
func batteryVoltageSensor() SensorV2{
    return SensorV2{Name: "Battery Voltage", Pin: BatteryVoltage, Quantity: dcv, Func: scale, Args: voltageScaleFactor, Units: voltage, Format: round1}
}

// other factories
````

Also notice that I'm making a bunch of stuff consts too (which could from a "string enum") this ensures we have a set of valid values for each field defined. We could have the set of valid Funcs for example, 
```
type Func string

const (
    Scale Func = "scale"
    Linear Func = "linear"
    // etc....
)
```


then have a factory for the default list of sensors,

````
func defaultControllerSensors() []SensorV2 {
    return []SensorV2{
        batteryVoltageSensor(),
        airTemperatureSensor(),
    }
```

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+)
+
+// Consts containing defaults for a controller config.
+const (
+	DefaultControllerInputs  string = "A0,X50,X51,X60,X10"
+	DefaultControllerOutputs string = "D14,D15,D16" // Peripheral Power.
+)
+
+// ControllerDefaultVars is a map which relates controller variable names to their default values
+var DefaultControllerVars = map[string]string{
```
saxon-milton, line 109: maybe factory too

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+var DefaultControllerSensors = []SensorV2{
+	{Name: "Battery Voltage", Pin: "A0", Quantity: "DCV", Func: "scale", Args: "0.0289", Units: "V", Format: "round1"},
+	{Name: "Analog Value", Pin: "X10", Quantity: "OTH", Func: "none", Format: "round1"},
+	{Name: "Air Temperature", Pin: "X50", Quantity: "MWH", Func: "linear", Args: "0.1,-273.15", Units: "C", Format: "round1"},
+	{Name: "Humidity", Pin: "X51", Quantity: "MMB", Func: "scale", Args: "0.1", Units: "%", Format: "round2"},
+	{Name: "Water Temperature", Pin: "X60", Quantity: "MTW", Func: "linear", Args: "0.1,-273.15", Units: "C", Format: "round1"},
+}
+
+// ControllerDefaultActs contains a slice of the default actuators for a controller device.
+var DefaultControllerActs = []ActuatorV2{
```
saxon-milton, line 134: ditto re factory

model/device.go
```diff
@@ -93,6 +93,146 @@ const (
...
+		return nil
+	}
+}
+
+// WithVariables creates the given variables for the device.
+func WithVariables(variables map[string]string) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
+		// Put all variables into the datastore.
+		for varName, varVal := range variables {
+			err := PutVariable(ctx, store, dev.Skey, dev.MAC()+"."+varName, varVal)
```
saxon-milton, line 164: I'm thinking we don't want to do this in here.

scruzin: Yes, I'm inclined to think that anything that does not directly involve configuring the device, should sit in a layer above the (device) model layer.

saxon-milton:
I'm wondering if we should reconsider the design of this slightly. It feels like we're conflating the concept of a Device, with a _system_ comprising the Device, as well as the Sensors and Actuators that connect up to the Device's inputs and outputs.

In other words, does it really make sense that our ```NewDevice``` constructor takes a bunch of options where we provide the Actuators and Sensors, and then the constructor spits out only a Device ? Also, we shouldn't need to have ```WithInputs``` options as well as ```WithSensors``` options right ? Shouldn't we be able to derive the inputs from the sensors; if that's exactly the reason that they're linked conceptually in the first place ?

Perhaps we need a new struct like ```ControllerSystem```, which has the ```Device```, ```Sensor``` and ```Actuator``` fields (not to be used as a datastore entity necessarily).  We'd then have a ```NewControllerSystem``` constructor that can take functional options, similarly to how we've done it, i.e. ```WithActuators```, ```WithSensors``` etc

This would then populate the ```ControllerSystem``` struct as appropriate.

Also, I'm not sure about how we've got the datastore operations built into the options and constructor, this seems inflexible/lacks modularity. I think we should have a couple of adapter functions instead; one for putting a ```ControllerSystem``` into the datastore (puts the individual entities into the datastore i.e. Device, Actuators etc) and on for getting a ```ControllerSystem``` (gets the required entities from the datastore and constructs a ```ControllerSystem```). 

We can use all this to create some factories for creating default controllers for example

```
func rigControllerSystem() *ControllerSystem {
    return NewControllerSystem(
                    WithActuators(defaultRigActuators()),
                    WithSensors(defaultRigSensors()),
                    WithVariables(defaultRigVariables()),
             ),
}

func jettyRigControllerSystem() *ControllerSystem {
    // blah
}

func reefTreatmentControllerSystem() *ControllerSystem {
    /// blah
}

```

I think this could be a really powerful way to start abstracting our systems.

model/device.go
ao-david: > type Input string
> 
> type ExternalInput Input
> 
> const (
>     SurfaceTemperature ExternalInput = "X50"
>     Humidity ExternalInput = "X51"
>     WaterTemperature ExternalInput = "X60"
>     // etc.... 
> )
> 
> type AnalogInput Input
> 
> const (
>     BatteryVoltage AnalogInput = "A0"
>     // etc ...
> )

I am going to make this a separate issue to add these values to the model/sensor to then be used in this change.

PR #204, opened 2024-07-02
Model: Add DeviceOptions (#195)

ao-david:
New DeviceOptions allow for the configuration of a new device with defaults relating to the specific type.
- WithLocation    - allows the user to set the latitude and longitude fields
- WithWifi        - allows the user to set the wifi username and password
- WithType        - specifies the type of the device
- WithoutDefaults - stops the NewDevice from using default values for configuration

Requires #203

### DIFF
--- a/model/device.go
+++ b/model/device.go
@@ -28,6 +28,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


model/device.go
```diff
@@ -200,6 +201,45 @@ func WithActuators(acts []ActuatorV2) DeviceOption {
 	}
 }
 
+// WithType sets the given type of the new device. If no type is passed, or the type is not a known type
+// this will create a device without any default settings (only MAC and name), and an empty device type.
+func WithType(deviceType string) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
```
saxon-milton, line 207: Why do these require context.Context and datastore.Store ? Generally with functional options you're just taking the thing you're applying the options to, i.e. Device.

model/device.go
```diff
@@ -200,6 +201,45 @@ func WithActuators(acts []ActuatorV2) DeviceOption {
...
+			deviceType = ""
+		}
+		dev.Type = deviceType
+		return nil
+	}
+}
+
+// WithWifi sets the WiFi SSID and password in a CSV format.
+func WithWifi(SSID, password string) DeviceOption {
+	return func(ctx context.Context, store datastore.Store, dev *Device) error {
```
saxon-milton, line 228: ditto

model/device.go
ao-david: This was to allow the options to create variables, sensors and actuators. Without this, these entities would have to be passed through the device somehow.

saxon-milton: That's why I think we should have a ```*System``` struct instead that can hold everything together.

ao-david: Yeah, I agree. I am in the process of making factories for sensors, variables, and actuators within model, which can be exported and used in a system layer

saxon-milton: I'm not sure there's much point doing this though; we know what direction we should going in with this, so why don't we just start with that ?

ao-david: Sorry if that was unclear, I am implementing the new model you suggested, as a starting point, I am going to make the factories for the sensors at the model layer (#207). These factories will be used in the system layer:
```golang
func defaultControllerSensors() []*SensorV2 {
  return []SensorV2{
    BateryVoltageSensor(defaultVoltageScaleFactor),
    AnalogSensor(),
    ...
  }
}

func NewDefaultControllerSystem() *ControllerSystem {
  return NewController(mac, name, 
    WithSensors(defaultControllerSensors),
    WithActuators(defaultControllerActurators),
  )
}
```
I will close this PR

scruzin: In retrospect, I think what Saxon is referring to as the `system` layer I was describing as the `manager` layer.

I'm okay with the `system` nomenclature, though. It may be worthy of its own package that logically sits above the `model` layer.

PR #205, opened 2024-07-02
Model: Add tests for NewDevice

ao-david:
This change adds tests to test the NewDevice function and some of the options.


saxon-milton:
This is likely to change right ?

ao-david:
> This is likely to change right ?
Yes, I will move to draft

Issue #206, opened 2024-07-03
model: Create consts for different models


ao-david:
When implementing #195 it was suggested that many of the types should be consts which can be referenced well across different implementations. 

This change should make consts for valid values that can be passed into entity creation, as well as define some factories to generate common instantiations.

PR #207, opened 2024-07-03
Model: Create Sensor factories (#206)

ao-david:
This change adds sensor factories for the following sensors:
- battery voltage
- analog value
- air temperature
- humidity sensor
- water temperature sensor

These factories use constant values defined in sensor.go, as well as exports from utils/nmea.

Requires https://github.com/ausocean/utils/pull/5


ao-david:
In regards to @saxon-milton's suggestion for a DefaultControllerSensors function, I think this should be included in a higher layer (system) which will only call the existing sensor factory functions, and does not require as many consts to be exported from model.

model/sensor.go
```diff
@@ -58,6 +58,33 @@ const (
 	keyCustom    = "custom"
 )
 
+// The keys for sensor formats.
+// See also SensorFormats below.
+const (
+	keyBool    string = "bool"
```
saxon-milton, line 64: Any reason why you're not having these typed ?

ao-david: This would require the sensor type itself to change, or mean that we would be calling ```string(keyBool)`` very often. I'm open to other suggestions if I have missed something

saxon-milton: I think we should do this and just use a shim between this and the device if we have to (or just string). Maybe in future we can change Device to use this type.

ao-david: I have taken a stab at implementing a shim.

Changing the actual fields of the sensorV2 struct is a very big change which certainly warrants a PR in itself (and lots of testing). I am not convinced that change should block this change.

scruzin: IMO, datastore entities typically use standard datastore types, which tend to be elemental GoLang types.

saxon-milton: Is that a technical limitation ? Or can it flatten things out to basic types e.g. would ``` type Input string``` just flatten out to string in the datastore ?

model/factories.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"github.com/ausocean/utils/nmea"
+)
+
+// Sensor factories //
+
+// Consts used by sensor factories.
+const (
+	tempLinearArgs      string = "0.1,-273.15"
```
saxon-milton, line 36: I reckon these should be const floats and then converted to strings when required.

model/factories.go
```diff
@@ -0,0 +1,127 @@
...
+	humidityScaleFactor string = "0.1"
+)
+
+// shimSensorV2 implements the fields of SensorV2 used here,
+// but replaces generic string fields with custom types.
+type shimSensorV2 struct {
```
saxon-milton, line 42: See also actuator PR regarding shim; I think we just need something like this
```
func sensorShim(n string, p pin, q nmea.Cod, f Func, a string, u unit, fmt Format) *SensorV2{
    return &SensorV2{n, string(p), string(q), string(f), a, string(u), string(fmt)}
}
```

model/factories.go
```diff
@@ -0,0 +1,127 @@
...
+		Quantity: string(s.Quantity),
+		Func:     string(s.Func),
+		Args:     s.Args,
+		Units:    string(s.Units),
+		Format:   string(s.Format),
+	}
+}
+
+// BatteryVoltageSensor returns a default calibrated battery sensor.
+func BatteryVoltageSensor(scaleFactor string) *SensorV2 {
```
saxon-milton, line 66: Should scale factor be float ?

model/sensor.go
```diff
@@ -48,14 +48,54 @@ const (
...
+
 // The keys for the function types.
 // See also SensorFuncs below.
 const (
-	keyNone      = "none"
-	keyScale     = "scale"
-	keyLinear    = "linear"
-	keyQuadratic = "quadratic"
-	keyCustom    = "custom"
+	keyNone      Func = "none"
```
saxon-milton, line 57: Do we need to prefix with ```key``` ?

model/sensor.go
```diff
@@ -48,14 +48,54 @@ const (
...
+	keyCustom    Func = "custom"
+)
+
+// Format defines the formatting on a sensor.
+type Format string
+
+// The keys for sensor formats.
+// See also SensorFormats below.
+const (
+	keyBool    Format = "bool"
```
saxon-milton, line 70: key ?

saxon-milton (changes requested):

model/sensor.go
ao-david: I was just using whatever was there, happy to change

model/factories.go
ao-david: Is there any detriment to the way that I implemented it? I started implementing the way that you have suggested, but it makes the code less clear IMO. My implementation looks like
```golang
// BatteryVoltageSensor returns a default calibrated battery sensor.
func BatteryVoltageSensor(scaleFactor string) *SensorV2 {
	return shimSensorV2{
		Name:     "Battery Voltage",
		Pin:      pinBatteryVoltage,
		Quantity: nmea.DCVoltage,
		Func:     keyScale,
		Args:     scaleFactor,
		Units:    unitVoltage,
		Format:   keyRound1,
	}.SensorV2()

}
```

The shim you suggested looks like
```golang
// BatteryVoltageSensor returns a default calibrated battery sensor.
func BatteryVoltageSensor(scaleFactor string) *SensorV2 {
	return shimSensorV2(
		"Battery Voltage",
		pinBatteryVoltage,
		nmea.DCVoltage,
		keyScale,
		scaleFactor,
		unitVoltage,
		keyRound1,
	)

}
```

With my shim, it makes it clear what values are being assigned to which field.

It's slightly more code overall (an extra struct definition), but I think it might be worth it. especially because not all sensors use all the fields, and so we would be passing null values into the function.

model/factories.go
```diff
@@ -0,0 +1,99 @@
...
+package model
+
+import (
+	"github.com/ausocean/utils/nmea"
+)
+
+// Consts/Pseudo used by sensor factories.
+const humidityScaleFactor Arg = 0.1
+
+var tempLinearArgs []Arg = []Arg{0.1, -273.15}
```
saxon-milton, line 35: IMO this should be returned from factory

saxon-milton (approved):

PR #209, opened 2024-07-03
Model: Add variable factories (#206)

ao-david:
This change adds factories to create the following variables (with passed values)
- AlarmNetwork
- AlarmPeriod
- AlarmRecoveryVoltage
- AlarmVoltage
- AutoRestart
- Power1
- Power2
- Power3
- Pulses
- PulseWidth
- PulseCycleVar
- PulseSuppress

## NOTE
It is up to the caller to define the value of the variable for most of these factories. I think this makes more sense, as the system layer will be making the decision about what the default values look like.

depends on #207

### DIFF
--- a/model/factories.go
+++ b/model/factories.go
@@ -26,6 +26,9 @@ LICENSE
 package model
 
 import (
+	"strconv"
+	"time"
+
 	"github.com/ausocean/utils/nmea"
 )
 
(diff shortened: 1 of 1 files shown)


model/factories.go
```diff
@@ -125,3 +125,113 @@ func WaterTemperatureSensor() *SensorV2 {
 		Format:   keyRound1,
 	}.SensorV2()
 }
+
+// Variable factories //
+
+// NewAlarmNetworkVar returns a new AlarmNetwork var with the
+// passed number of failures as its value.
+func NewAlarmNetworkVar(numberOfFailures string) *Variable {
```
scruzin, line 133: Although Variable values are all represented as strings, it would be preferable if the constructors took the underlying type, e.g., an `int` in the case of `numberOfFailures`, and performed the conversion.

model/factories.go
```diff
@@ -125,3 +125,113 @@ func WaterTemperatureSensor() *SensorV2 {
...
+func NewAutoRestartVar(duration string) *Variable {
+	return &Variable{
+		Name:  "AutoRestart",
+		Value: duration,
+	}
+}
+
+// NewPower1Var returns a new Power1 var with the
+// power initialised to OFF.
+func NewPower1Var() *Variable {
```
scruzin, line 178: Similarly, this should take a `bool` which is converted to either "true" or "false".

scruzin (changes requested):

model/factories.go
```diff
@@ -125,3 +125,113 @@ func WaterTemperatureSensor() *SensorV2 {
 		Format:   keyRound1,
 	}.SensorV2()
 }
+
+// Variable factories //
```
saxon-milton, line 129: don't think you need this

saxon-milton (changes requested):
Looks pretty good, but I tend to agree with Alan; these factories would be nicer if they took more natural types for the parameters and then made the necessary conversions.

ao-david:
> Looks pretty good, but I tend to agree with Alan; these factories would be nicer if they took more natural types for the parameters and then made the necessary conversions.

Done

saxon-milton (approved):

PR #210, opened 2024-07-03
Model: Add actuator factories (#206)

ao-david:
This change implements factories for the following actuators.
- Device 1 -> Power 1
- Device 2 -> Power 2
- Device 3 -> Power 3

### DIFF
--- a/model/actuator.go
+++ b/model/actuator.go
@@ -45,6 +45,13 @@ const (
 	numParts       = 4 // Number of parts for actuator in TSV.
 )
 
+// Pins used by actuators.
+const (
+	PinPower1 Pin = "D16"
+	PinPower2 Pin = "D14"
+	PinPower3 Pin = "D15"
+)
+
 // Actuator represents an actuator datastore type to link a device output with
 // a variable such that that variable value changes the physical device output value.
 // NB: This type is deprecated. See ActuatorV2.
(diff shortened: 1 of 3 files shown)


model/factories.go
```diff
@@ -235,3 +235,51 @@ func NewPulseSuppressVar(value string) *Variable {
 		Value: value,
 	}
 }
+
+// Actuator factories //
```
saxon-milton, line 239: don't need

model/factories.go
```diff
@@ -235,3 +235,51 @@ func NewPulseSuppressVar(value string) *Variable {
...
+
+// shimActuatorV2 provides a typed shim for a ActuatorV2.
+type shimActuatorV2 struct {
```
saxon-milton, line 242: Not sure if this is what I had in mind exactly for the shim. I was thinking more along the lines of a function that made the conversion. It could look something like this, 

```
func shim(n string, p Pin, v Var) *ActuatorV2 {
    return &ActuatorV2{Name: n, Pin: string(p), Var: v}
}

func NewDeviceActuator() *Actuator {
    return shim("Device 1", PinPower1,"Power1")
}
```

saxon-milton (changes requested):

model/actuator.go
```diff
@@ -45,6 +45,13 @@ const (
 	numParts       = 4 // Number of parts for actuator in TSV.
 )
 
+// Pins used by actuators.
+const (
+	PinPower1 Pin = "D16"
```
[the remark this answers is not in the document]
saxon-milton: I think we probably want to differentiate digital and analog pins, but I guess that can happen at a later point.

saxon-milton (approved):

PR #217, opened 2024-07-08
oceancenter/dl: use run_pi_client.sh v1.1.0

### LINKED ISSUE #216: oceancenter/dl: run_pi_client.sh version bump in releases
Now that ausocean.service is used for all clients, it is assuming that the run_pi_client.sh version is at least 1.1.0. However hydrophone v1.0.1 does not contain this version, and should be updated.

ao-david:
ausocean.service expects run_pi_client.sh so we should ensure that this version is being served.

closes #216


scruzin (approved):
LGTM

scruzin:
This is now obsolete as v2.0.0 is the latest version of run_pi_client.sh

scruzin:
From v2.0.0, run_pi_client.sh should be part of the image and not part of the device software package.

It _could_ be part of future packages if we need to upgrade the run script.

ao-david:
> This is now obsolete as v2.0.0 is the latest version of run_pi_client.sh

Should this PR be closed then?

Issue #188, opened 2024-06-29
model: Site entity should include YouTubeEmail


scruzin:
At present, the YouTube account email is associated with the broadcast config, which means it is handled every time 
a configuration is updated, usually unnecessarily. The principle of _data minimization_ suggests that we rethink this approach.

On the reasonable assumption that the given Site is only ever associated with one YouTube account, the YouTube account email should move to the Site entity, along with the `OwnerEmail` (the Google account associated with the site).

saxon-milton:
In what way is the account handled every time the configuration is updated ? I don't necessarily think this is the case, but maybe to some level it's unavoidable with how we're doing things.

I think there might still be circumstances where we want an AusOcean managed site, but have one of the cameras going to someone elses channel ?

scruzin:
The account is part of the broadcast config but, AFAIK, is only required to generate a YT access token. In contrast, all other properties of the broadcast config are inherent properties of the broadcast.

Further, a given premium customer, although they may have multiple broadcasts, will in all likelihood use the same YT account for all of them (just as we do).

We can still have an AusOcean-managed site, but specify a YT account email for that site that is different to our standard AusOcean YT channel.

saxon-milton:
The account field in the BroadcastConfig is also used to retrieve the token from a bucket object; not sure how much that changes things.

scruzin:
We could handle that case by creating an extra site for that purpose. Seems like a rare use case, though.

saxon-milton:
Opps, that close was unintentional...

scruzin:
Just to clarify my earlier comment about _data minimization_, the general idea with schema design is factor out common data to avoid duplication. For example, `Device` entities include neither the email address of the device's owner, nor the name of the site. Instead, this common data is part of the `Site` entity, which is shared by multiple devices. Similarly, there will typically be multiple broadcasts per site, all using the same account (enforced by policy, if not technically).

As a bonus, moving the YT account email to the Site means it comes accessible to usage outside of just the broadcast manager. For example, we could send emails to YT account owners to advise them of YT policy changes, etc.

saxon-milton:
> Just to clarify my earlier comment about _data minimization_, the general idea with schema design is factor out common data to avoid duplication. For example, `Device` entities include neither the email address of the device's owner, nor the name of the site. Instead, this common data is part of the `Site` entity, which is shared by multiple devices. Similarly, there will typically be multiple broadcasts per site, all using the same account (enforced by policy, if not technically).
[...]

PR #218, opened 2024-07-08
Support for Site V3

scruzin:
This adds `Description`, `OrgID`, `OpEmail` and `YouTubeEmail` to `Site` entity. 

Implements Issue #188.

### DIFF
--- a/model/site.go
+++ b/model/site.go
@@ -6,7 +6,7 @@ AUTHORS
   Alan Noble <alan@ausocean.org>
 
 LICENSE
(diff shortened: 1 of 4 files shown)


model/site.go
```diff
@@ -40,7 +40,11 @@ const typeSite = "Site"
 type Site struct {
 	Skey         int64
 	Name         string
+	Description  string
```
saxon-milton, line 43: Doesn't have to happen in this PR but i think these fields should have comments seeings as they're exported.

saxon-milton (approved):

PR #225, opened 2024-07-16
TV: Fix error matching for ErrNoSuchEntity

### LINKED ISSUE #213: Broadcast: New broadcast fails to save
Since this change #190 we can't create new broadcasts since we are now using Update for calls to the broadcast even if it doesn't exist. This results in the save failing.

ao-david:
When saving a new broadcast, errors coming from cloudstore are now properly matched with the correct types from cloud.google.com/go/datastore.

This change also changes the updateConfig function since there will be no config to unmarshal

NOTE:
This is currently deployed to fix a P0 level bug

closes #213

### DIFF
--- a/cmd/oceantv/utils.go
+++ b/cmd/oceantv/utils.go
@@ -37,6 +37,7 @@ import (
 	"strings"
 	"time"
 
+	googledatastore "cloud.google.com/go/datastore"
 	"github.com/ausocean/cloud/model"
 	"github.com/ausocean/openfish/datastore"
 )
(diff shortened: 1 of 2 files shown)


cmd/oceantv/utils.go
```diff
@@ -37,6 +37,7 @@ import (
 	"strings"
 	"time"
 
+	googledatastore "cloud.google.com/go/datastore"
```
scruzin, line 40: Using "cloud.google.com/go/datastore" is breaking the encapsulation provided by openfish/datastore.
Please do this another way.

PS Apart from breaking the encapsulation, it breaks standalone mode.

cmd/oceantv/utils.go
```diff
@@ -162,12 +163,39 @@ func updateConfigWithTransaction(ctx context.Context, store Store, skey int64, b
 	}
 
 	err := store.Update(ctx, key, updateConfig, &model.Variable{})
-	if errors.Is(err, datastore.ErrNoSuchEntity) {
+	// We need to check for both filestore and cloudstore error types here.
+	if errors.Is(err, datastore.ErrNoSuchEntity) || errors.Is(err, googledatastore.ErrNoSuchEntity) {
```
scruzin, line 167: Test for openfish/datastore errors, not cloud.google.com/go/datastore errors.

scruzin (changes requested):

ao-david:
After a little thinking I think what needs to happen is the openfish/datastore layer needs to check for the errNoExistError coming from google datastore, and replace it with a openfish error which we can then read from cloud.

cmd/oceantv/utils.go
trekhopton: This can now be left just as datastore.ErrNoSuchEntity since this PR:
https://github.com/ausocean/openfish/pull/147

scruzin (approved):
LGTM. Please update the version too.

Issue #232, opened 2024-07-17
tv: checkbroadcasts cron is not being added automatically anymore


trekhopton:
Since PR #190 OceanBroadcastManager's Save uses the updateConfigWithTransaction function, not the saveBroadcast function in broadcast.go. Therefore it doesn't check to see if the checkbroadcast cron exists. CreateBroadcast uses the Save function so the cron doesn't get checked upon creation. 
Perhaps we should add the cron check to the CreateBroadcast function instead of the save function.

PR #233, opened 2024-07-17
tv: move auto add checkbroadcast cron to CreateBroadcast

trekhopton:
This was done because CreateBroadcast no longer uses the saveBroadcast function that checks the cron. But we still want the cron to be checked whena broadcast is created.

Addresses issue #232.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -332,6 +332,12 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 			reportError(w, r, req, "could not save broadcast: %v", err)
 			return
 		}
+		// Ensure that the CheckBroadcast cron exists.
+		c := &model.Cron{Skey: cfg.SKey, ID: "Broadcast Check", TOD: "* * * * *", Action: "rpc", Var: tvURL + "/checkbroadcasts", Enabled: true}
+		err = model.PutCron(context.Background(), settingsStore, c)
+		if err != nil {
+			return reportError(w, r, req, "Warning: failed to verify checkbroadcasts cron: %v", err)
+		}
 
 	case broadcastDelete:
 		err = deleteBroadcast(ctx, &req, settingsStore)
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -455,13 +455,6 @@ func saveBroadcast(ctx context.Context, cfg *BroadcastConfig, store datastore.St
 		return fmt.Errorf("could not put broadcast data in store: %w", err)
 	}
 
-	// Ensure that the CheckBroadcast cron exists.
-	c := &model.Cron{Skey: cfg.SKey, ID: "Broadcast Check", TOD: "* * * * *", Action: "rpc", Var: projectURL + "/checkbroadcasts", Enabled: true}
-	err = model.PutCron(ctx, store, c)
-	if err != nil {
-		return fmt.Errorf("failure verifying check broadcast cron: %w", err)
-	}
-
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


cmd/oceantv/broadcast_manager.go
```diff
@@ -127,6 +127,14 @@ func (m *OceanBroadcastManager) CreateBroadcast(
 	if err != nil {
 		return fmt.Errorf("could not update config with transaction: %w", err)
 	}
+
+	// Ensure that the CheckBroadcast cron exists.
+	c := &model.Cron{Skey: cfg.SKey, ID: "Broadcast Check", TOD: "* * * * *", Action: "rpc", Var: projectURL + "/checkbroadcasts", Enabled: true}
```
saxon-milton, line 132: OceanBroadcastManager.CreateBroadcast isn't related to broadcast config creation (which is what creating a broadcast on bench does). OceanBroadcastManager.CreateBroadcast uses the BroadcastService to create a broadcast object on YouTube. In short, this shouldn't be in here but rather in the broadcast save handler for bench.

saxon-milton (changes requested):

cmd/oceantv/broadcast_manager.go
saxon-milton: To clarify, I think this should go in "saveBroadcast" in oceanbench

cmd/oceantv/broadcast.go
```diff
@@ -460,13 +460,6 @@ func saveBroadcast(ctx context.Context, cfg *BroadcastConfig, store datastore.St
 		return fmt.Errorf("could not put broadcast data in store: %w", err)
 	}
 
-	// Ensure that the CheckBroadcast cron exists.
```
saxon-milton, line 463: I'm thinking we could also remove this version of saveBroadcast (oceanbench has a different version that we keep); any operation on broadcast configs should use transactions now, so it's dangerous keeping this.

trekhopton: I agree, I tried to briefly but it's still used in SetupSecondary and I wasn't sure how to handle how it wants to create another new config, not just modify the existing one. Perhaps another issue/PR.

cmd/oceantv/broadcast_manager.go
trekhopton: In oceanbench, not oceantv's handler of /broadcast/save? Okay I suppose that makes most sense since it's sort of separate to saving a broadcast.

trekhopton: Done.

cmd/oceantv/broadcast.go
saxon-milton: Oh right, OceanBroadcastManager.Save should handle creation of a new broadcast. So we can replace that call with that. But year, another PR sounds good.

cmd/oceanbench/broadcast.go
```diff
@@ -343,6 +343,12 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 			reportError(w, r, req, "could not save broadcast: %v", err)
 			return
 		}
+		// Ensure that the CheckBroadcast cron exists.
```
saxon-milton, line 346: What's your rationale for putting it here rather than in https://github.com/ausocean/cloud/blob/beeff00faca7ec3738340c685701755d65c5d143/cmd/oceanbench/broadcast.go#L399
?

trekhopton: I didn't really have a rationale, but now after consideration, I think it actually makes more sense to be here because I think it's more intuitive that the cron only gets set when you click save. If we put it in the saveBroadcast function it will happen at other points that saveBroadcast is called eg. when generating a token.

saxon-milton: Yeah, that's a fair point.

saxon-milton (approved):
All good once rebased

PR #236, opened 2024-07-18
Remove unused or inappropriate config fields

saxon-milton:
This change performs some cleaning up of the BroadcastConfig struct. We remove some fields that are unused i.e. StartTime, EndTime, CamOn and CamOff, and also relocate LastStatusCheck, LastChatMsg and LastHealthCheck, given that these are state specific and can be part of the state data. Moving the lastBlah states was most easily done by embedding a couple of structs with common fields into states that require these, and have these implement an interface to allow access of these fields.


scruzin (approved):
LGTM

Issue #115, opened 2024-06-12
cmd: Common standalone filestore


ao-david:
When testing with standalone instances of the different services, it is useful to have them all using the same filestore. This way oceanbench is able to read the broadcasts that have been saved by oceanTV. In their current state, unless using appengine mode, each service uses a distinct filestore which each other service does not have access to.

scruzin:
At present, there's a pretty easy work around to this already, though. Just run all binaries from the same working directory and the standalone file store _is_ shared.

For example, say you wish to run oceanbench and oceancron.

From terminal 1:
```
cd ~/go/src/github.com/ausocean/cloud/cmd/oceanbench
./oceanbench --standalone
```

From terminal 2:
```
cd ~/go/src/github.com/ausocean/cloud/cmd/oceanbench
../oceancron/oceancron -- standalone
```

That said, we could add a `--standalone-store` flag that specifies the path to the shared file store.

ao-david:
This is probably fine behaviour, but it would be good to document this if it is not already. If it is already documented, maybe we need to think about how this documentation is being done as I missed this.

ao-david:
@scruzin After doing some more testing using this method today, I have discovered an issue with this.

When bench is accessing broadcasts from the store it uses the vidgrind store (store/vidgrind), whereas oceantv will store its broadcasts in store/oceantv. This means that you can successfully save a broadcast, but you won't be able to load that broadcast from the filestore.

scruzin:
Two solutions:

- Change oceantv to use store/vidgrind not store oceantv
- Add a `--standalone-store` flag that specifies the path to the shared file store.

scruzin:
Implemented in Pull Request #238

PR #238, opened 2024-07-19
Added --filestore command line flag to specify file store path

scruzin:
Also ensured that all of the default ports  in standalone mode are different:

- oceanbench 8080
- oceancron 8081
- oceantv 8082
- datablue 8083
- oceancenter 8084

Implements Issue #115


ao-david (approved):
LGTM. will be much nicer to run locally now.

PR #239, opened 2024-07-19
Device status var

scruzin:
With this change, it is possible to programmatically set the device status by setting a variable called `status` to a value between 0 and 7 (as defined in `device.go`).

This will be most useful in conjunction with cron jobs. For example, to force a device to reboot, set the variable `<device.mac-address>.status` to the value `2`.

### DIFF
--- a/model/device.go
+++ b/model/device.go
@@ -40,6 +40,7 @@ import (
 const typeDevice = "Device"
 
 // Device state statuses.
+// Keep StatusText and IsValidStatus is sync with this list.
 const (
 	DeviceStatusOK = iota
 	DeviceStatusUpdate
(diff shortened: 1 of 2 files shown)


model/device.go
```diff
@@ -431,3 +432,13 @@ func DeviceIsUp(ctx context.Context, store datastore.Store, mac string) (bool, e
 	}
 	return false, nil
 }
+
+// IsValidStatus returns true if the supplied number is a valid status, or false otherwise.
+func IsValidStatus(n int64) bool {
+	switch n {
+	case DeviceStatusOK, DeviceStatusUpdate, DeviceStatusReboot, DeviceStatusDebug, DeviceStatusUpgrade, DeviceStatusAlarm, DeviceStatusTest, DeviceStatusShutdown:
```
saxon-milton, line 439: Optional, but there's a pattern I've seen for iota based enums where you add an "end" value to the enum so that you can do this,
```
func IsValidStatus(n int64) bool {
    if 0 <= n && n < deviceStatusEnd {
        return true
    }
    return false
}
```

Then when you add any more values you don't have to update the validation function.

saxon-milton (approved):

PR #245, opened 2024-07-21
Start adding configuration sanitisation

saxon-milton:
Depends on PR #244 

We're starting with checking the hardware is defined for a given broadcast, otherwise we disable. We're doing this with an invalidConfigurationEvent. This can be used for various configuration issues, and for convenience also impelements the error interface.

We're adding a basic test case that highlights the ease of integration type testing using the new broadcastSystem abstraction added in the prior change.

### DIFF
--- a/cmd/oceantv/broadcast_states.go
+++ b/cmd/oceantv/broadcast_states.go
@@ -37,11 +37,12 @@ func (ctx *broadcastContext) log(msg string, args ...interface{}) {
 }
 
 const (
-	broadcastGeneric   notify.Kind = "broadcast-generic"   // Problems where cause is unknown.
-	broadcastForwarder notify.Kind = "broadcast-forwarder" // Problems related to out forwarding service i.e. can't stream slate.
-	broadcastHardware  notify.Kind = "broadcast-hardware"  // Problems related to streaming hardware i.e. controllers and cameras.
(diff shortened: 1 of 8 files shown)


cmd/oceantv/broadcast_states.go
```diff
@@ -37,11 +37,12 @@ func (ctx *broadcastContext) log(msg string, args ...interface{}) {
 }
 
 const (
-	broadcastGeneric   notify.Kind = "broadcast-generic"   // Problems where cause is unknown.
-	broadcastForwarder notify.Kind = "broadcast-forwarder" // Problems related to out forwarding service i.e. can't stream slate.
-	broadcastHardware  notify.Kind = "broadcast-hardware"  // Problems related to streaming hardware i.e. controllers and cameras.
-	broadcastNetwork   notify.Kind = "broadcast-network"   // Problems related to bad bandwidth, generally indicated by bad health events.
-	broadcastSoftware  notify.Kind = "broadcast-software"  // Problems related to the functioning of our broadcast software.
+	broadcastGeneric       notify.Kind = "broadcast-generic"       // Problems where cause is unknown.
+	broadcastForwarder     notify.Kind = "broadcast-forwarder"     // Problems related to out forwarding service i.e. can't stream slate.
```
scruzin, line 41: out -> our

scruzin (approved):

Issue #250, opened 2024-07-24
bench: app crashes if broadcast is selected that doesn't exist


trekhopton:
I deleted a broadcast and the broadcast page loaded fine but the corresponding secondary broadcast was shown in the drop-down menu. I clicked on the secondary Broadcast and then the page didn't load. Looking at the logs, it caused bench to crash. See the below logs:

`DEFAULT 2024-07-23T07:01:26.784059Z /admin/broadcast
DEFAULT 2024-07-23T07:01:28.058103Z 2024/07/23 07:01:28 existing broadcast selected: Broadcast.Willunga Test Rig 2(Secondary)
DEFAULT 2024-07-23T07:01:28.058163Z 2024/07/23 07:01:28 could not load existing settings for broadcast: could not get broadcast from vars: broadcast with name Broadcast.Willunga Test Rig 2(Secondary) doesn't exist
DEFAULT 2024-07-23T07:01:28.058918Z 2024/07/23 07:01:28 ExecuteTemplate failed on broadcast.html: template: broadcast.html:122:92: executing "broadcast.html" at <.Site.Timezone>: nil pointer evaluating *model.Site.Timezone
`
After navigating to the broadcast page again via the nav menu I could no longer see the secondary stream.

After briefly looking at the code, this could be caused from not returning after an error when getting the Site for the timezone, leaving the site to be nil.

PR #251, opened 2024-07-24
bench: check site in broadcast.html exists, return after GetSite error

trekhopton:
This was done to prevent crashing on nil Sites for broadcasts on broadcast.html.

Closes Issue #250


ao-david (approved):

ao-david:
Bump bench version before merging

Issue #253, opened 2024-07-24
bench: user should be able to tell if broadcasts are saved or deleted successfully.


trekhopton:
I propose we include a message upon saving and deleting broadcasts that tell the user if the action was performed successfully.

PR #254, opened 2024-07-24
bench: send user a message for saves and deletes on broadcast page.

trekhopton:
This was done because it is hard to tell is an operation has been performed successfully.

Closes issue #253 

messages have also been added for slate upload and authentication of a new channel too.


ao-david (commented):
LGTM. Just remember to bump main again

PR #258, opened 2024-07-24
Add RecoveryCallbackServeMux

### LINKED ISSUE #257: cloud: write a custom serve mux offering handler panic recovery email notification

saxon-milton:
closes #257 

This is a ServeMux that extends the http.ServeMux to allow the provision of custom handler panic recovery. This was motivated by the desire to have notification on recovery of handler panics so that things don't fail so silently.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -1,5 +1,9 @@
 package utils
 
+import (
+	"net/http"
+)
+
 // TokenURIFromAccount forms a Google Cloud Storage URI for a YouTube token
 // based on the provided account. If the account is empty, it's assumed that
 // the legacy token is being used. Otherwise, the account is used to form the
@@ -20,4 +24,37 @@ func TokenURIFromAccount(account string) string {
 	const tokenPostfix = ".youtube.token.json"
 
 	return bucket + account + tokenPostfix
+}
+
+// RecoveryCallbackServeMux extends the default http.ServeMux and accepts
+// a callback function to be called in case of handler panic recovery.
+type RecoveryCallbackServeMux struct {
+	*http.ServeMux
+	recover func(w http.ResponseWriter, err any)
+}
+
+// NewRecoveryCallbackServeMux creates a new RecoveryCallbackServeMux.
+// recover is the callback function to be called in case of handler panic recovery.
+func NewRecoveryCallbackServeMux(recover func(w http.ResponseWriter, err any)) *RecoveryCallbackServeMux {
+	return &RecoveryCallbackServeMux{http.NewServeMux(), recover}
+}
+
+// ServeHTTP applies the recovery middleware and serves the HTTP request.
+func (m *RecoveryCallbackServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	defer func() {
+		if err := recover(); err != nil {
+			m.recover(w, err)
+		}
+	}()
+	m.ServeMux.ServeHTTP(w, r)
+}
+
+// Handle registers the handler for the given pattern.
+func (m *RecoveryCallbackServeMux) Handle(pattern string, handler http.Handler) {
+	m.ServeMux.Handle(pattern, handler)
+}
+
+// HandleFunc registers the handler function for the given pattern.
+func (m *RecoveryCallbackServeMux) HandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request)) {
+	m.ServeMux.HandleFunc(pattern, handler)
 }
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -1,0 +1,46 @@
+package utils
+
+import (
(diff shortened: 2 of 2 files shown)


ao-david:
I would assume that this change makes no significant impact to performace and/or delay, but would be good to know for sure?

scruzin (approved):
I think this is good, but I think it would be even better if we adopted a common interface for all of our services and then these functions became methods on that interface.

That way, the recovery methods would have direct access to the service state and perform more complex recoveries, such as re-initializing the datastore or notifier.

For example, _Ocean Center_ has the following type. All but `totpSecret` is common to every AusOcean service.

```
// service defines the properties of our web service.
type service struct {
	setupMutex    sync.Mutex
	settingsStore datastore.Store
	debug         bool
	standalone    bool
	notifier      notify.Notifier
	totpSecret    []byte
	storePath     string
}

```

saxon-milton:
> I would assume that this change makes no significant impact to performace and/or delay, but would be good to know for sure?

I might be missing something, but what makes you think it would have any effect on performance ?

saxon-milton:
> I think this is good, but I think it would be even better if we adopted a common interface for all of our services and then these functions became methods on that interface.
> 
> That way, the recovery methods would have direct access to the service state and perform more complex recoveries, such as re-initializing the datastore or notifier.
> 
> For example, _Ocean Center_ has the following type. All but `totpSecret` is common to every AusOcean service.
> 
> ```
> // service defines the properties of our web service.
> type service struct {
> 	setupMutex    sync.Mutex
> 	settingsStore datastore.Store
> 	debug         bool
> 	standalone    bool
> 	notifier      notify.Notifier
> 	totpSecret    []byte
> 	storePath     string
> }
> ```

I had a similar thought but I'm thinking it would be good if we could provide multiple ```RecoveryHandler``` (which would be a function type) so that types like ```service``` could implement this interface multiple times for different panic conditions. It would look something like this,
```
func main(){
        /// ........
        mux := utils.NewRecoverableServeMux(
		utils.NewConfigurableRecoveryHandler(
			utils.WithHandlers(
                                svc.errNoGlobalNotifierHandler,
                                svc.errAnotherHandler,
                        ),
		),
	)
	// .....
}

func (s *service) errNoGlobalNotifierHandler(w http.ResponseWriter, panicErr any) bool {
        err, ok := panicErr.(error)
        if !ok {
                return false
        }
        if errors.Is(err, errNoGlobalNotifier){
               // ... remediation code
               return true
        }
        return false
}
```

The configurable recovery handler would loop through the registered handlers until it found one that handled the panic.

So, regarding commonality between services, we could have ```BaseService``` which would provide some common panic handlers. Other services would the "inherit" from the base service. This means they can either use the common handler or they can override with their own. In this example, we use register the common handler, but override the "errAnotherHandler" handler
```
type BaseService struct {
 	setupMutex    sync.Mutex
 	settingsStore datastore.Store
 	debug         bool
 	standalone    bool
 	notifier      notify.Notifier
 	totpSecret    []byte
 	storePath     string
 }
 
 func (s *BaseService) errNoGlobalNotifierHandler(w http.ResponseWriter, panicErr any) bool { // ...}

func (s *BaseService) errAnotherHandler(w http.ResponseWriter, panicErr any) bool { // ...}

type OceanTVService struct {
    BaseService
}

func (s *OceanTVService) errAnotherHandler(w http.ResponseWriter, panicErr any) bool {// ...}

func main(){
        /// ........
        mux := utils.NewRecoverableServeMux(
		utils.NewConfigurableRecoveryHandler(
			utils.WithHandlers(
                                svc.errNoGlobalNotifierHandler,
                                svc.errAnotherHandler,
                        ),
		),
	)
	// .....
}

```

ao-david:
> > I would assume that this change makes no significant impact to performace and/or delay, but would be good to know for sure?
> 
> I might be missing something, but what makes you think it would have any effect on performance ?

I don't think it would. I not familiar with defer and recover, but having an extra layer seems like it _could_ affect performance.

I'm in no way saying it would, but just wanted to make sure that we're thinking of performance given that we're already having issues.

saxon-milton:
> > > I would assume that this change makes no significant impact to performace and/or delay, but would be good to know for sure?
> > 
> > 
> > I might be missing something, but what makes you think it would have any effect on performance ?
> 
> I don't think it would. I not familiar with defer and recover, but having an extra layer seems like it _could_ affect performance.
> 
> I'm in no way saying it would, but just wanted to make sure that we're thinking of performance given that we're already having issues.

Sure! Personally I think that unless you have an obvious reason to think something would slow things down/ you actually see a non negligible performance hit when you run it, you don't have to be thinking about performance. Languages like go and C++ with their compilers are just so good at optimising these days! There's probably a distinction to be made between data transfer/handling performance (which is what our current problem is arguably?) vs raw code performance.

PR #259, opened 2024-07-24
Add Send to notify package

saxon-milton:
Depends on #258 

This is intended to be a bare-bones email send function i.e. it doesn't deal with filters and categorisation based on site and kind like the MailjetNotifier implementation offers.

It's intended to be used for critial notification e.g. panic recovery.


scruzin (approved):
LGTM

PR #260, opened 2024-07-24
Utilise RecoveryCallbackServeMux in oceantv

saxon-milton:
depends on #259 

This change updates oceantv to use the RecoveryCallbackServeMux. This means that on handler panic we no longer use the default recovery mechanism, which is relatively quiet, and instead now use a bare-bones email notification to make it clear that a panic has occured during serving.


ao-david (approved):
LGTM

PR #262, opened 2024-07-24
bench: ensure timezone is applied correctly when converting from timestamp to date

### LINKED ISSUE #261: bench: when a date is modified and saved for a broadcast, it loses 30 minutes
When a date for a broadcast is modified it loses 30 minutes signifying that the conversion between timestamp and date is not happening properly.

It seems that the timezone is not being interpreted correctly again. This was fixed for Date to timestamp conversion but not timestamp to date conversion.

trekhopton:
This was done because the time was losing 30 minutes due to a misinterpretation of the timezone.

Closes #261

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -72,7 +72,7 @@ import (
 )
 
 const (
-	version     = "v0.21.6"
+	version     = "v0.21.7"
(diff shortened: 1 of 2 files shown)


scruzin (approved):
LGTM

PR #265, opened 2024-07-27
Removed run_pi_client.sh from package

scruzin:
run_pi_client.sh is now part of the image and does not need to be part of the package.


cmd/oceancenter/dl/hydrophone/pkg/v1.0.2/pkg.json
```diff
@@ -0,0 +1,22 @@
+{
+  "device": "hydrophone",
+  "description": "Install hydrophone release. Must use raspios bullseye or later.",
```
trekhopton, line 3: This description doesn't need the "Must use raspios bullseye or later" anymore. Now it's just that the image needs to have the microphone drivers installed before the package can be installed, which happens when we prepare an image.

trekhopton (approved):

ao-david:
I think this can be closed @scruzin (not merged) superseded by #296

PR #266, opened 2024-07-27
Fixed tests

scruzin:
Also fixed a few string conversions in factories.go.


saxon-milton:
Not sure if in the scope of this PR but looks like ```TestRun``` under oceancron is panicking.

saxon-milton (approved):

PR #267, opened 2024-07-29
TV: Use onFailureClosure for request limits (#222)

### LINKED ISSUE #222: broadcast: on request limit excession we should disable broadcast
We don't currently handle the case when we don't have any more request tokens left i.e. we have exceeded the YT request limit. This means we get a start failed event, but try again the next time. Instead we should probably use the onFailureClosure which can disable the broadcast.

ao-david:
When a ErrRequestLimitExceeded error is received from svc.CreateBroadcast we will use the disable on first failure bool in the failure closure to ensure the broadcast is disabled before making any additional calls to youtube (which will also be rate limited). This will prevent us from extending our broadcast down time by needlessly making blocked calls, adding to our rate limit.

closes #222

### DIFF
--- a/cmd/oceantv/broadcast_test.go
+++ b/cmd/oceantv/broadcast_test.go
@@ -62,16 +62,32 @@ type dummyManager struct {
 	cfg                                                                *Cfg
 	startDone                                                          chan struct{}
 	saved, started, stopped, healthHandled, statusHandled, chatHandled bool
+	Limiter                                                            RateLimiter
 	t                                                                  *testing.T
 }
 
-func newDummyManager(t *testing.T, cfg *Cfg) *dummyManager {
+type dummyManagerOption func(interface{}) error
+
+func withRateLimiter(l RateLimiter) dummyManagerOption {
+	return func(i interface{}) error {
+		if s, ok := i.(*dummyManager); ok {
+			s.Limiter = l
+		}
+		return nil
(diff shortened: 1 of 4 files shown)


trekhopton (approved):

saxon-milton:
Only thing I'd say about this one is that we could try and add testing with
it (think we should get into this habit with broadcast stuff). Not sure
what exactly this would look like though. Might have to add a mock limiter
to the mock broadcast service.

Cheers,
Saxon

davidsutts:
> Only thing I'd say about this one is that we could try and add testing with it (think we should get into this habit with broadcast stuff). Not sure what exactly this would look like though. Might have to add a mock limiter to the mock broadcast service.

Yeah I did briefly have a look at this. I will try and implement a mock limiter.

ao-david:
@saxon-milton added testing (and a mock limiter)

cmd/oceantv/broadcast_test.go
```diff
@@ -388,3 +407,15 @@ func (bus *mockEventBus) checkEvents(want []event) error {
 	}
 	return nil
 }
+
+type mockLimiter struct {
```
saxon-milton, line 411: I think there's a mock.go file that can house all mock stuff.

saxon-milton (approved):

PR #268, opened 2024-07-29
ignore .git and .github folders

trekhopton:
This was done because they were increasing the size of gcloud deployments.


ao-david (approved):
LGTM

scruzin (approved):

Issue #104, opened 2024-06-05
bench: broadcast page shouldn't resubmit form on reload.


trekhopton:
When a form is submitted, it's generally good practice to either go to another URL, or clear the form and display a success message. 
We should implement one of these options on the broadcast page.

I propose we initially have the broadcast config displayed as a view only UI with labels, then when a user clicks "edit" it loads admin/broadcast/edit and turns the config labels into config form inputs and they can be edited. A user can then either click "save" which submits the new config and loads admin/broadcast, or cancel, which takes them back to admin/broadcast without saving.

scruzin:
Is this fixed @trekhopton

trekhopton:
No, it's not although it's more of a desirable change than a bug, not very high priority.

deborah-baker:
This caused an issue where multiple streams were started because the browser refresh was used @ao-david

PR #269, opened 2024-07-30
bench: add refresh button to broadcast page

trekhopton:
This was done because it was inconvenient to use the navigation menu then have to reselect a broadcast to refresh the selection.

image

### DIFF
--- a/cmd/oceanbench/s/broadcast.js
+++ b/cmd/oceanbench/s/broadcast.js
@@ -76,6 +76,9 @@ function buttonClick(button) {
 }
 
 function submitSelect(select) {
+  if (!select){
+    select = document.getElementById("broadcast-select");
+  }
   select.form.querySelector("input[name='action']").value = "broadcast-select";
   select.form.submit();
 }
--- a/cmd/oceanbench/t/broadcast.html
+++ b/cmd/oceanbench/t/broadcast.html
@@ -41,8 +41,11 @@
     <div class="border rounded p-4 container-md bg-white">
       <form action="/admin/broadcast" enctype="multipart/form-data" method="post">
           <div class="d-flex align-items-center gap-1 mb-1">
+            <button class="advanced btn btn-primary btn-sm" onclick="submitSelect()">Refresh</button>
+          </div>
+          <div class="d-flex align-items-center gap-1 mb-1">
             <label for="broadcast-select" class="w-25 text-end">Select Broadcast:</label>
-            <select name="broadcast-select" class="form-select w-50 h-auto align-items-center" onchange="submitSelect(this)" id="broadcast-name-select">
+            <select id="broadcast-select" name="broadcast-select" class="form-select w-50 h-auto align-items-center" onchange="submitSelect(this)" id="broadcast-name-select">
               <option value="">-- New Broadcast --</option>
               {{range .BroadcastVars}}
                 <option value="{{ .Name }}" {{if eq (part (split .Name ".") 1) $.CurrentBroadcast.Name}}selected{{end -}}>{{ .Name }}</option>


saxon-milton:
I'm not sure I understand what the problem is and what this does; might you be able to elaborate ?

trekhopton:
> I'm not sure I understand what the problem is and what this does; might you be able to elaborate ?

I just made an issue, see #275.

ao-david:
I understand the issue, however the fix seems more like a bandaid over a symptom rather than a fix. I've been advocating for a while to consider the way that we are submitting forms, to make them consistent and logical. I think that this is probably fine for now, but long term we need to make sure all our forms work the same and are not broken by refreshing, changing sites etc...

saxon-milton:
> I understand the issue, however the fix seems more like a bandaid over a symptom rather than a fix. I've been advocating for a while to consider the way that we are submitting forms, to make them consistent and logical. I think that this is probably fine for now, but long term we need to make sure all our forms work the same and are not broken by refreshing, changing sites etc...

@trekhopton thanks for writing the issue, it's clearer now. @ao-david I tend to agree; I'm not sure this is the best solution. ISTM we could just try to clear the action after it's been performed, then refresh would resubmit it again ?

ao-david:
> Just because we clear the form values doesn't mean we'll get rid of that prompt asking to resubmit a form.

There is a js method to change the history of the page, which can remove any requests made to reach the page. It might be worth looking into quickly. I'm not sure that many people will default to using the page refresh button over the browser refresh button, but if it makes your life easier, I don't think there is anything particularly wrong with it for a bandaid solution

trekhopton:
> > Just because we clear the form values doesn't mean we'll get rid of that prompt asking to resubmit a form.
> 
> There is a js method to change the history of the page, which can remove any requests made to reach the page. It might be worth looking into quickly. I'm not sure that many people will default to using the page refresh button over the browser refresh button, but if it makes your life easier, I don't think there is anything particularly wrong with it for a bandaid solution

I just looked into that quickly and it doesn't seem straight forward. It would make my life easier until we change the way we load this form pages. I think the solution will be something like redirecting to another page when the form has been submitted. If you're happy with it, can you approve it?

ao-david (approved):

ao-david:
> > > Just because we clear the form values doesn't mean we'll get rid of that prompt asking to resubmit a form.
> > 
> > 
> > There is a js method to change the history of the page, which can remove any requests made to reach the page. It might be worth looking into quickly. I'm not sure that many people will default to using the page refresh button over the browser refresh button, but if it makes your life easier, I don't think there is anything particularly wrong with it for a bandaid solution
> 
> I just looked into that quickly and it doesn't seem straight forward. It would make my life easier until we change the way we load this form pages. I think the solution will be something like redirecting to another page when the form has been submitted. If you're happy with it, can you approve it please?

Maybe don't close the issue, or write a new issue that describes all of the non-ideal behaviour that we have with forms, and how we could avoid it

trekhopton:
> > > > Just because we clear the form values doesn't mean we'll get rid of that prompt asking to resubmit a form.
> > > 
> > > 
> > > There is a js method to change the history of the page, which can remove any requests made to reach the page. It might be worth looking into quickly. I'm not sure that many people will default to using the page refresh button over the browser refresh button, but if it makes your life easier, I don't think there is anything particularly wrong with it for a bandaid solution
> > 
> > 
> > I just looked into that quickly and it doesn't seem straight forward. It would make my life easier until we change the way we load this form pages. I think the solution will be something like redirecting to another page when the form has been submitted. If you're happy with it, can you approve it please?
> 
> Maybe don't close the issue, or write a new issue that describes all of the non-ideal behaviour that we have with forms, and how we could avoid it

There's already an issue for it. See #104.

PR #271, opened 2024-07-30
tv: only create a new broadcast when one doesn't already exist

### LINKED ISSUE #42: broadcast: reuse the same broadcast object when there's a failure.
Currently when a broadcast fails, another one is created; for various failure reasons, this can happen multiple times, and to the YouTube subscribers, it appears as different streams in their feed. This is undesirable and we should figure out how to reuse the same broadcast to retry. Or at the very least delete the old one programmatically so that two don’t exist at the same time.

trekhopton:
closes #42

I have tested this on my machine and it will only create and use one broadcast, when the hardware is finally operational it successfully transitions to live.


cmd/oceantv/broadcast/youtube.go
```diff
@@ -68,6 +68,7 @@ var (
 	ErrNoBroadcastItems = errors.New("no broadcast items")
 )
 
+// Broadcast ID, Stream ID and Chat ID.
```
saxon-milton, line 71: Thanks for adding this, but remember comment structure is ``` // IDs contains the Stream ID ... ```

cmd/oceantv/broadcast_manager.go
```diff
@@ -88,44 +88,58 @@ func (m *OceanBroadcastManager) CreateBroadcast(
 	store Store,
 	svc BroadcastService,
 ) error {
-	// We're going to add the date to the broadcast's name, so get this and format.
-	loc, err := time.LoadLocation(locationID)
+	// If the broadcast doesn't have an ID or is revoked or completed, create a new broadcast.
+	ctx := context.Background()
```
saxon-milton, line 92: Can probably just probide context.Background() as the argument below.

cmd/oceantv/broadcast_manager.go
```diff
@@ -88,44 +88,58 @@ func (m *OceanBroadcastManager) CreateBroadcast(
...
+	status, err := svc.BroadcastStatus(ctx, cfg.ID)
 	if err != nil {
-		return fmt.Errorf("could not load location: %w", err)
+		return fmt.Errorf("could not get broadcast status: %v", err)
 	}
+	if cfg.ID == "" || status == "" || status == broadcast.StatusRevoked || status == broadcast.StatusComplete {
```
saxon-milton, line 97: Could we check for the negative case instead i.e. conditions for using same broadcast. Then we don't need an else, and we don't need to nest this whole block e.g.
``` 
if cfg.ID != "" && status != "" && status != broadcast.StatusRevoked && status != broadcast.StatusComplete {
    err = m.Save ....
    // err check
    return nil 
}

// Create broadcast

Comment (saxon-milton) [reply cmd/oceantv/broadcast_manager.go:97]: Also, should we have any sort of time based check i.e. if the start time is like yesterday then should we really use this broadcast ?

**Comment by saxon-milton** [comment]
Seems like a pretty good solution. Might you be able to tests for this ?

**Comment by saxon-milton** [comment]
Don't forget to bump oceantv version

**Comment by trekhopton** [comment]
> Seems like a pretty good solution. Might you be able to tests for this ?

I'm not sure what that would look like. I'd rather not at this stage.

File: cmd/oceantv/broadcast_manager.go (hunk 3 above)
Comment (trekhopton) [reply cmd/oceantv/broadcast_manager.go]: The idea is that no uncompleted broadcasts from yesterday should exist to use.

**Comment by saxon-milton** [comment]
> > Seems like a pretty good solution. Might you be able to tests for this ?
> 
> I'm not sure what that would look like. I'd rather not at this stage.

I think it's worth getting into the habit of adding testing for any new functionality though ? We should have a great testing infrastructure to perform broadcast testing.

File: cmd/oceantv/broadcast_manager.go (hunk 3 above)
Comment (saxon-milton) [reply cmd/oceantv/broadcast_manager.go]: How can we be sure that assumption is true ?

Comment (trekhopton) [reply cmd/oceantv/broadcast_manager.go]: If it's not true, then that would be a bug that we should fix. I suppose it may be helpful to check it here too. I just prefer solving the problem at the source.

**Comment by trekhopton** [comment]
> > > Seems like a pretty good solution. Might you be able to tests for this ?
> > 
> > 
> > I'm not sure what that would look like. I'd rather not at this stage.
> 
> I think it's worth getting into the habit of adding testing for any new functionality though ? We have a great testing infrastructure to perform broadcast testing.

I guess I'm just not familiar with our testing infrastructure. Where would a test like this go? And what might it look like?

**Comment by saxon-milton** [comment]
> > > > Seems like a pretty good solution. Might you be able to tests for this ?
> > > 
> > > 
> > > I'm not sure what that would look like. I'd rather not at this stage.
> > 
> > 
> > I think it's worth getting into the habit of adding testing for any new functionality though ? We have a great testing infrastructure to perform broadcast testing.
> 
> I guess I'm just not familiar with our testing infrastructure. Where would a test like this go? And what might it look like?

Part of your job is to figure that out :stuck_out_tongue_winking_eye: 

I noted an example using the new broadcast system in slack :smile:

File: cmd/oceantv/broadcast_manager.go (hunk 3 above)
Comment (scruzin) [reply cmd/oceantv/broadcast_manager.go]: Is it worth encapsulating the if test condition as a function, e.g., `IsValidBroadcast`?

Comment (saxon-milton) [reply cmd/oceantv/broadcast_manager.go]: @trekhopton there's the possibility that we create a broadcast but can't get the hardware properly started or get the broadcast to transition to a live state (even after retries), so we'd be left with an unused broadcast. I don't think we currently have any mechanism that removes broadcasts that don't eventuate (we could have this though). We could check for these sorts of conditions (especially unexpected conditions) and log accordingly. 

Also, agree with @scruzin; we could put this logic in a IsValidBroadcast function :smile:

File: cmd/oceantv/broadcast_manager.go [hunk 4]
```diff
@@ -88,6 +88,20 @@ func (m *OceanBroadcastManager) CreateBroadcast(
 	store Store,
 	svc BroadcastService,
 ) error {
+	// If the broadcast doesn't have an ID or is revoked or completed, create a new broadcast.
```
saxon-milton, line 91: Should this comment be updated to reflect what we're doing in this check ?

cmd/oceantv/broadcast_manager.go
trekhopton: Okay, I think we should remove broadcasts that don't eventuate or at least give the user that option. But yeah I think I agree it would be good to check that we're using a broadcast from today. I've added that now.

Also added the function.

trekhopton:
> > > > > Seems like a pretty good solution. Might you be able to tests for this ?
> > > > 
> > > > 
> > > > I'm not sure what that would look like. I'd rather not at this stage.
> > > 
> > > 
> > > I think it's worth getting into the habit of adding testing for any new functionality though ? We have a great testing infrastructure to perform broadcast testing.
> > 
> > 
> > I guess I'm just not familiar with our testing infrastructure. Where would a test like this go? And what might it look like?
> 
> Part of your job is to figure that out 😜
> 
> I noted an example using the new broadcast system in slack 😄

I've added testing.

cmd/oceantv/broadcast_manager_test.go
```diff
@@ -0,0 +1,114 @@
...
+			if b != tt.expectedReuse {
+				t.Errorf("broadcastCanBeReused() test failed for %s: expected %v, got %v", tt.name, tt.expectedReuse, b)
+			}
+		})
+	}
+}
+
+// dummyGoodService is a dummy implementation of the BroadcastService interface.
+// It mostly does nothing like dummyService except its BroadcastStatus method returns "upcoming" for test purposes.
+type dummyGoodService struct{}
```
saxon-milton, line 86: You could just use the dummyService (which should become "mockService") and provide an option to configure it to have it return "upcoming".

cmd/oceantv/broadcast_manager_test.go
```diff
@@ -0,0 +1,114 @@
+package main
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ausocean/cloud/cmd/oceantv/broadcast"
+)
+
+func TestBroadcastCanBeReused(t *testing.T) {
```
saxon-milton, line 10: Might you also be able to add an integration test using the broadcastSystem ?

cmd/oceanbench/broadcast.go
```diff
@@ -145,8 +145,10 @@ type BroadcastConfig struct {
 	StartFailures     int           // The number of times the broadcast has failed to start.
 	Transitioning     bool          // If the broadcast is transition from live to slate, or vice versa.
 	StateData         []byte        // States will be marshalled and their data stored here.
+	HardwareStateData []byte        // Hardware states will be marshalled and their data stored here.
 	Account           string        // The YouTube account email that this broadcast is associated with.
 	InFailure         bool          // True if the broadcast is in a failure state.
+	TimeCreated       time.Time     // The date and time the broadcast was created.
```
saxon-milton, line 151: Couldn't we just use the scheduled start time from the broadcast object data ?

cmd/oceantv/broadcast_manager_test.go
saxon-milton: Here's an example

https://github.com/ausocean/cloud/blob/main/cmd/oceantv/broadcast_machine_test.go#L1348

trekhopton: I did have a look at writing an integration test but for CreateBroadcast, it needs a token bucket limiter in the store or something and I wasn't sure how to imitate that.
I get this: could not get token bucket limiter: could not unmarshal token bucket limiter: unexpected end of JSON input

cmd/oceanbench/broadcast.go
trekhopton: Do you mean the Start field of the BroadcastConfig struct? It's my understanding that that doesn't get updated to reflect the date of the stream, we just use the time portion of that field for start and stop times.

Are you suggesting that we update the date in the Start field to be current when the broadcast is created, then just use that.. That could work.

cmd/oceantv/broadcast_manager_test.go
trekhopton: David has mentioned that he did something with options on this PR: https://github.com/ausocean/cloud/pull/267/files

I'll have a look and see what I can do with options because I'm not quite familiar with how I can modify the test system with options.

cmd/oceanbench/broadcast.go
saxon-milton: I mean more like the youtube broadcast object.

trekhopton: Done.

cmd/oceantv/broadcast_manager_test.go
trekhopton: Done.

cmd/oceantv/broadcast_manager_test.go
trekhopton: I think I'll leave this for now since it's somewhat of an urgent fix and it's taking some time for me to figure out the system testing stuff. That can happen in another PR if need be.

cmd/oceanbench/main.go
```diff
@@ -72,7 +72,7 @@ import (
 )
 
 const (
-	version     = "v0.21.8"
+	version     = "v0.21.9"
```
saxon-milton, line 75: We also need an oceantv version bump.

cmd/oceantv/broadcast/youtube.go
```diff
@@ -353,6 +354,22 @@ func getBroadcastStatus(svc *youtube.LiveBroadcastsService, id string) (string,
 	return resp.Items[0].Status.LifeCycleStatus, nil
 }
 
+// GetBroadcastScheduledStart gets the status of the broadcast with the provided ID.
+func GetBroadcastScheduledStart(svc *youtube.Service, id string) (string, error) {
```
saxon-milton, line 358: Should this be ```GetBroadcastStatus``` ?

cmd/oceantv/broadcast/youtube.go
```diff
@@ -353,6 +354,22 @@ func getBroadcastStatus(svc *youtube.LiveBroadcastsService, id string) (string,
...
+	return getBroadcastStatus(youtube.NewLiveBroadcastsService(svc), id)
+}
+
+func getBroadcastScheduledStart(svc *youtube.LiveBroadcastsService, id string) (string, error) {
+	resp, err := svc.List([]string{"snippet"}).Id(id).Do()
+	if err != nil {
+		return "", fmt.Errorf("could not list broadcasts: %w", err)
+	}
+	if len(resp.Items) == 0 {
```
saxon-milton, line 367: Should we also check for more than one ? It's a really unexpected condition because you'd think that IDs are unique, but this code only works if this assumption holds. I would even panic under that condition.

cmd/oceantv/broadcast_manager.go
```diff
@@ -381,3 +397,24 @@ func opsHealthNotifyFunc(ctx context.Context, cfg *BroadcastConfig) func(string)
 		return notifier.Send(ctx, cfg.SKey, broadcastGeneric, msg)
 	}
 }
+
+// broadcastCanBeReused checks if a broadcast can be reused based on how old it
+// is, if it has been revoked or completed, and if its IDs have been set.
+func (m *OceanBroadcastManager) broadcastCanBeReused(cfg *BroadcastConfig, svc BroadcastService) bool {
+	startTime, err := svc.BroadcastScheduledStartTime(context.Background(), cfg.ID)
```
saxon-milton, line 404: Not checking error ?

cmd/oceantv/broadcast/youtube.go
trekhopton: I think it's okay to proceed even if multiple are listed. We'll only use the first one right?

cmd/oceanbench/main.go
trekhopton: done.

cmd/oceantv/broadcast/youtube.go
trekhopton: No, changed.

cmd/oceantv/broadcast/youtube.go
saxon-milton: Isn't that weird though ?

saxon-milton: I think we should at least log.

cmd/oceantv/broadcast_manager_test.go
saxon-milton: As discussed I think we can deploy fixes before they are merged. I think this change deserves some integration testing to go along with it.

cmd/oceantv/broadcast/youtube.go
trekhopton: Yeah that would be weird, a log sounds good so we can at least tell if there's weirdness going on.

cmd/oceantv/broadcast_manager_test.go
trekhopton: See TestCreateBroadcast, it uses an actual broadcast manager so it runs the CreateBroadcast code.

saxon-milton (approved):
I think we can get this merged once rebased and conflicts resolved :)

ao-david:
I'm excited for this. When this is merged and working, does this mean that we shouldn't have to disable broadcasts anymore?

PR #272, opened 2024-07-30
bench: add cloudblue favicon

trekhopton:


scruzin (commented):
So you want to have a single favicon?

Normally, each binary should have its own favicon under cmd/xxx.

Just have to make sure app.yaml handles them.

trekhopton:
> So you want to have a single favicon?
> 
> Normally, each binary should have its own favicon under cmd/xxx.
> 
> Just have to make sure app.yaml handles them.

@scruzin This is really for oceanbench, but I don't think it matters if they all have the same favicon seeing that they're all under CloudBlue.
I only put it at the root directory because that's where the handler expects it to be. I can move it and change the handler if you like?

scruzin:
> > So you want to have a single favicon?
> > Normally, each binary should have its own favicon under cmd/xxx.
> > Just have to make sure app.yaml handles them.
> 
> @scruzin This is really for oceanbench, but I don't think it matters if they all have the same favicon seeing that they're all under CloudBlue. I only put it at the root directory because that's where the handler expects it to be. I can move it and change the handler if you like?

I think it should be with the binary, for consistency with other content (such as templates).

trekhopton:
This now uses the logo that David made.

trekhopton:
> > > So you want to have a single favicon?
> > > Normally, each binary should have its own favicon under cmd/xxx.
> > > Just have to make sure app.yaml handles them.
> > 
> > 
> > @scruzin This is really for oceanbench, but I don't think it matters if they all have the same favicon seeing that they're all under CloudBlue. I only put it at the root directory because that's where the handler expects it to be. I can move it and change the handler if you like?
> 
> I think it should be with the binary, for consistency with other content (such as templates).

I moved it to cmd/oceanbench and added a handler in the app yaml. I understand what you mean now about app.yaml specifying the static files for appengine and the handlers in main just being for standalone.

ao-david (approved):

PR #273, opened 2024-07-30
tv: check for youtube credentials env var in standalone

trekhopton:
This was done so we can stream to our youtube channel from our local machines.

### DIFF
--- a/cmd/oceantv/broadcast/storage.go
+++ b/cmd/oceantv/broadcast/storage.go
@@ -130,14 +130,9 @@ func objTok(ctx context.Context, url string) (*oauth2.Token, error) {
 // fileTok extracts an oauth2.Token from a file with name specified by the given
 // URL.
 func fileTok(uri string) (*oauth2.Token, error) {
-	var name string
+	name := os.Getenv("YOUTUBE_API_CREDENTIALS")
 	var err error
-	if standalone {
-		name = os.Getenv("YOUTUBE_API_CREDENTIALS")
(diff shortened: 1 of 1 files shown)


cmd/oceantv/broadcast/storage.go
```diff
@@ -130,9 +130,16 @@ func objTok(ctx context.Context, url string) (*oauth2.Token, error) {
 // fileTok extracts an oauth2.Token from a file with name specified by the given
 // URL.
 func fileTok(uri string) (*oauth2.Token, error) {
-	_, name, err := googleStorageAddr(uri)
-	if err != nil {
-		return nil, fmt.Errorf("could not parse uri: %w", err)
+	c := os.Getenv("YOUTUBE_API_CREDENTIALS")
+	var name string
```
scruzin, line 134: You could simplify by defaulting name, then overriding when it is empty, i.e,

```
	name := os.Getenv("YOUTUBE_API_CREDENTIALS")
	var err error
	if name == "" {
		_, name, err = googleStorageAddr(uri)
		if err != nil {
			return nil, fmt.Errorf("could not parse uri: %w", err)
		}
	}

```

scruzin (changes requested):

cmd/oceantv/broadcast/storage.go
trekhopton: Done.

scruzin (approved):

PR #274, opened 2024-08-01
Provide  NewConfigurableRecoveryHandler

saxon-milton:
This provides a nice interface for creating a recovery handler. It works on the assumption that there's some common things we might like to do for panic recover i.e. first decide whether we should handle the particular panic, format a message in response to the panic, log that message, notify through email/sms of that message, produce a HTTP status and finally maybe ask for help from further, more custom handlers.

We provide options to customize each of these steps using the functional options pattern.

We're now using this handler factory to set up a handler for oceantv. We've added an additional handler that is invoked for errNoGlobalNotifier type panics, in which case we initialise the global notifier in attempt to remediate the problem.


utils/utils.go
```diff
@@ -26,35 +28,225 @@ func TokenURIFromAccount(account string) string {
...
+				cfg.logOutput(fmt.Sprintf("could not notify of panic: %v", err))
+			}
+		}
+
+		for _, handle := range cfg.handlers {
+			if handle(w, panicErr) {
+				handled = true
+			}
+		}
+		return handled
```
scruzin, line 213: It would be good to log when there is no panic handler (handled == false).

scruzin (approved):
One suggestion, but otherwise LGTM.

PR #276, opened 2024-08-05
system: More Options and add IO automatically (#195)

ao-david:
The followig options were added:
- WithDefaults
- WithWifi
- WithLocation

Acutators and sensors now add to the input and output list of the controller.

### DIFF
--- a/system/rig_system.go
+++ b/system/rig_system.go
@@ -3,7 +3,9 @@ package system
 import (
 	"context"
 	"fmt"
+	"log"
 	"strings"
+	"time"
 
 	"github.com/ausocean/cloud/model"
 	"github.com/ausocean/openfish/datastore"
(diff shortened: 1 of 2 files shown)


trekhopton:
@ao-david I'm sure you're aware but this isn't building.

system/rig_system.go
```diff
@@ -89,10 +161,16 @@ func NewRigSystem(skey int64, MAC, name string, options ...Option) (*RigSystem,
 	}
 	for _, sensor := range sys.Sensors {
 		sensor.Mac = sys.Controller.Mac
+		sys.Controller.Inputs += sensor.Pin + ","
```
trekhopton, line 164: This will result in a trailing comma after the last entry. We don't want that. Ditto below.

trekhopton (approved):
LGTM other than the err const not building. Fix that and it's all g.

PR #289, opened 2024-08-13
Ensure that utils form is not resubmitted (#288)

### LINKED ISSUE #288: Bench: Moving a device breaks variables and actuators
When using /utilities to move a device, variables are not carried across to the new site. Additionally, actuators do not properly move either, resulting in issues with ```processActuators``` hence resulting in the status not showing correctly. As the uptime variable is updated _after_ responding to the client

ao-david:
Previously, if the site was changed on the utils page, it would resubmit the form. Since moving a device puts a variable, then deletes it, if the user switches to the site the device was moved to, it will delete the variables associated with that device.

This change also update the error reporting in datablue which can hide the root issue.

closes #288

### DIFF
--- a/cmd/datablue/handlers.go
+++ b/cmd/datablue/handlers.go
@@ -363,7 +363,7 @@ func processActuators(ctx context.Context, dev *model.Device, respMap map[string
 		// Actuator var names are relative to their device.
 		val, err := model.GetVariable(ctx, settingsStore, dev.Skey, dev.Hex()+"."+act.Var)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


scruzin (approved):
LGTM but don't forget to bump the version too.

scruzin (changes requested):
LGTM but don't forget to bump the version too.

Also, add a check to Ocean Bench `utilsHandler` to ensure `targetSkey != skey`. .

trekhopton (approved):

scruzin (approved):

PR #290, opened 2024-08-16
TV: Add back incrementing of failures (#231)

### LINKED ISSUE #231: TV: Broadcasts not disabling
Broadcasts should disable after a given number of attempts to start, which will also prevent youtube from being spammed with requests to start new broadcasts. Currently this feature doesn't seem to be working (at least in some instances)

ao-david:
The number of start failures was not being incremented correctly, meaning that broadcasts were not disabling after failing 3 times.

closes #231

### DIFF
--- a/cmd/oceantv/broadcast_states.go
+++ b/cmd/oceantv/broadcast_states.go
@@ -717,6 +717,7 @@ func onFailureClosure(ctx *broadcastContext, cfg *BroadcastConfig, disableOnFirs
 		ctx.bus.publish(startFailedEvent{})
 		try(ctx.man.Save(nil, func(_cfg *BroadcastConfig) {
 			const maxStartFailures = 3
+			_cfg.StartFailures++
 			if disableOnFirstFail || _cfg.StartFailures >= maxStartFailures {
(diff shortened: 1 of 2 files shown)


cmd/oceantv/broadcast_states.go
```diff
@@ -716,6 +716,7 @@ func onFailureClosure(ctx *broadcastContext, cfg *BroadcastConfig, disableOnFirs
 		ctx.log("failed to start broadcast: %v", err)
 		ctx.bus.publish(startFailedEvent{})
 		try(ctx.man.Save(nil, func(_cfg *BroadcastConfig) {
+			_cfg.StartFailures++
 			const maxStartFailures = 3
```
scruzin, line 720: IMO, it is slightly easier to read if the const declarations are at the start of a block, i.e., before `_cfg.StartFailures++`

scruzin (approved):

cmd/oceantv/broadcast_states.go
ao-david: fixed

PR #300, opened 2024-08-20
tv: handle stop and start requests to prevent getting stuck in restarting

trekhopton:
I noticed the hardware machine could get into a state of restarting but ignoring stop and start requests. Since it was ignoring start requests it never transitioned out of that state. Now it will transition, same for a stop request while restarting.

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -226,11 +226,11 @@ func (sm *hardwareStateMachine) handleHardwareStartedEvent(event hardwareStarted
 func (sm *hardwareStateMachine) handleHardwareStartRequestEvent(event hardwareStartRequestEvent) {
 	sm.log("handling hardware start request event")
 	switch sm.currentState.(type) {
-	case *hardwareOff, *hardwareStopping:
+	case *hardwareOff, *hardwareStopping, *hardwareRestarting:
 		sm.transition(newHardwareStarting(sm.ctx))
 	case *hardwareStarting:
 		sm.ctx.camera.publishEventIfStatus(hardwareStartedEvent{}, true, sm.ctx.cfg.CameraMac, sm.ctx.store, sm.log, sm.ctx.bus.publish)
-	case *hardwareOn, *hardwareRestarting:
+	case *hardwareOn:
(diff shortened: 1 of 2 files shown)


scruzin (approved):
LGTM.

Out of curiosity, is the "hardware restarting" state a late addition, which is why it was previously overlooked?

saxon-milton:
Are you merging into the right branch ?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -226,11 +226,11 @@ func (sm *hardwareStateMachine) handleHardwareStartedEvent(event hardwareStarted
 func (sm *hardwareStateMachine) handleHardwareStartRequestEvent(event hardwareStartRequestEvent) {
 	sm.log("handling hardware start request event")
 	switch sm.currentState.(type) {
-	case *hardwareOff, *hardwareStopping:
+	case *hardwareOff, *hardwareStopping, *hardwareRestarting:
```
saxon-milton, line 229: I'm wondering if I had it the way it was because I thought a restart should complete (or fail) before we tried starting/stopping. Is the problem really that we're not coming out of the restarting state ?

saxon-milton (changes requested):

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: OK I've reminded myself of how this works; it does come out of the state using a hardwareStartRequestEvent.

saxon-milton (approved):

saxon-milton:
> LGTM.
> 
> Out of curiosity, is the "hardware restarting" state a late addition, which is why it was previously overlooked?

Not a late addition, but the behaviour change and forgot to add some transitions.

saxon-milton:
I almost think while we're at it, we should have the hardwareRestarting state implement stateWithTimeout so that it's not restarting forever. I'm also thinking this is the kind of thing that should have testing i.e. hardware issues and then restart.

trekhopton:
> Are you merging into the right branch ?

I branched off dup-broadcasts because I wanted to deploy both changes at once. I might have to rebase to merge into main now.

PR #301, opened 2024-08-20
Limit YouTubeEmail notifications

scruzin:
Only notify YouTubeEmail for hardware, network, and configuration issues.

### DIFF
--- a/cmd/oceantv/main.go
+++ b/cmd/oceantv/main.go
@@ -45,7 +45,7 @@ import (
 
 const (
 	projectID          = "oceantv"
-	version            = "v0.2.2"
+	version            = "v0.2.3"
 	projectURL         = "https://oceantv.appspot.com"
 	cronServiceAccount = "oceancron@appspot.gserviceaccount.com"
 	locationID         = "Australia/Adelaide" // TODO: Use site location.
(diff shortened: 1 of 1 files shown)


cmd/oceantv/main.go
```diff
@@ -223,16 +223,25 @@ func setup(ctx context.Context) {
...
 func tvRecipients(skey int64, kind notify.Kind) ([]string, time.Duration, error) {
 	ctx := context.Background()
 	site, err := model.GetSite(ctx, settingsStore, skey)
 	if err != nil {
 		return nil, 0, fmt.Errorf("error getting site: %w", err)
 	}
+	if site.OpsEmail == "" {
+		log.Printf("OpsEmail not defined for site %s", site.Name)
+	}
 	recipients := []string{site.OpsEmail}
```
trekhopton, line 235: site.OpsEmail will still be added if it is empty here. Recipients should be an empty slice if OpsEmail is empty.

trekhopton (changes requested):

cmd/oceantv/main.go
scruzin: Empty strings are ignored by the notifier.

trekhopton (approved):

PR #302, opened 2024-08-22
Bench: Add new device endpoint to create new device with defaults (#195)

ao-david:
an endpoint has been added to /set/devices/new that reads the form fields for
- name
- mac
- wifi
- location
- device type and creates a new device based on these settings. At the moment the only supported device type is a controller.

### DIFF
--- a/cmd/oceanbench/set.go
+++ b/cmd/oceanbench/set.go
@@ -26,6 +26,7 @@ LICENSE
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log"
@@ -37,6 +38,7 @@ import (
 
 	"github.com/ausocean/cloud/gauth"
 	"github.com/ausocean/cloud/model"
+	"github.com/ausocean/cloud/system"
 	"github.com/ausocean/openfish/datastore"
 	"github.com/ausocean/utils/nmea"
 )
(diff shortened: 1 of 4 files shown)


cmd/oceanbench/set.go
```diff
@@ -374,6 +376,99 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+	pass := r.FormValue("pass")
+	sLat := r.FormValue("lat")
+	sLong := r.FormValue("long")
+	r.ParseForm()
+
+	var lat, long float64
+	if sLat != "" && sLong != "" {
+		lat, err = strconv.ParseFloat(sLat, 64)
+		if err != nil {
+			writeError(w, fmt.Errorf("unable to parse float64 from: %s, err:", err))
```
trekhopton, line 417: Looks like you were intending to include the variable in the log message but instead you're including the error in the wrong spot.

cmd/oceanbench/set.go
```diff
@@ -374,6 +376,99 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		long, err = strconv.ParseFloat(sLong, 64)
+		if err != nil {
+			writeError(w, fmt.Errorf("unable to parse float64 from: %s, err:", err))
```
trekhopton, line 422: Message needs to be restructure as above.

trekhopton (changes requested):

cmd/oceanbench/set.go
```diff
@@ -374,6 +376,99 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, "/set/devices?ma="+ma, http.StatusFound)
 }
 
+// configDevicesHandler handles configuration of new devices.
+//
+// Form Fields:
+//
+//	name = Name of the new device
+//	ma = MAC address
```
scruzin, line 384: For consistency, please use standard two-letter query params, i.e., `ma` for MAC address, `ct` for client type, `wi` for the command separated string ssid-name, ssid-password, `ll` for comma-separated latitude,longitude, etc.

scruzin (changes requested):

cmd/oceanbench/set.go
scruzin: PS These query params form part of our REST API, that is why it's important to keep them consistent.

ao-david: done

cmd/oceanbench/set.go
ao-david: done

trekhopton (approved):

PR #303, opened 2024-08-22
Bench: Add unique devices page for sandbbox site

ao-david:
This change alters the default devices page for the sandbox site, tailoring the experience towards new device configuring.

image
image

requires #302

### DIFF
--- a/cmd/oceanbench/set.go
+++ b/cmd/oceanbench/set.go
@@ -32,6 +32,7 @@ import (
 	"log"
 	"math/rand"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -106,6 +107,12 @@ func writeDevices(w http.ResponseWriter, r *http.Request, msg string, args ...in
 		return
 	}
 	skey, _ := profileData(profile)
+	sandbox := false
+	file := "set/device.html"
+	if skey == model.SandboxSkey {
+		sandbox = true
+		file = "sandbox.html"
+	}
 
 	data := devicesData{
 		commonData: commonData{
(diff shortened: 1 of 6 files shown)


cmd/oceanbench/set.go
```diff
@@ -189,6 +201,35 @@ func writeDevices(w http.ResponseWriter, r *http.Request, msg string, args ...in
 		return
 	}
 
+	// Provide uptime and device status information.
```
scruzin, line 204: Why has this chunk of code moved?

cmd/oceanbench/set.go
```diff
@@ -106,6 +107,12 @@ func writeDevices(w http.ResponseWriter, r *http.Request, msg string, args ...in
 		return
 	}
 	skey, _ := profileData(profile)
+	sandbox := false
+	file := "set/device.html"
+	if skey == 3 {
```
scruzin, line 112: Use a const, e.g. _sanboxSiteKey_

scruzin (changes requested):
Since the sandbox device settings page is used entirely for configuring newly detected devices, we should consider a completely different endpoint to /set/device.

Also, note that new devices must be explicitly enabled. This gives the operator of choosing to provision a device, but not enabled it.

cmd/oceanbench/set.go
ao-david: Maybe this should be exported somewhere, since it is also used by oceancenter?

scruzin: Yes. I think it could live in `model/site.go` in lieu of another common cloud package.

cmd/oceanbench/set.go
ao-david: since the sandbox site page does not need all the device variables, moving this code here prevents calls to the database for unused variables, and instead only gets the ones that matter first.

cmd/oceanbench/set.go
ao-david: done

trekhopton (approved):

ao-david:
> Since the sandbox device settings page is used entirely for configuring newly detected devices, we should consider a completely different endpoint to /set/device.

I am going to make this a separate issue to handle in a separate PR

PR #304, opened 2024-08-22
Bench: Add configuration page to apply defaults (#195)

ao-david:
The configuration page lets a user simply add the fields that are required and then configures and saves the device accordingly.

image

Currently this shows all the sites to the user, but in the future this should use a getAllUserSites() (which would also be used by the site menu)

requires #303 #302


trekhopton (approved):

ao-david:
This PR has had some substantial changes since the last review which were required when rebasing onto the changes made in previous PRs.

cmd/oceanbench/main.go
```diff
@@ -720,7 +720,7 @@ func logRequest(r *http.Request) {
 // writeError writes an error in JSON format.
 func writeError(w http.ResponseWriter, err error) {
 	w.Header().Add("Content-Type", "application/json")
-	fmt.Fprint(w, `{"er":"`+err.Error()+`}`)
+	fmt.Fprint(w, `{"er":"`+err.Error()+`"}`)
```
saxon-milton, line 723: I think we should do 
```json.NewEncoder(w).Encode(map[string]string{"er": err.Error()})```
that way issues like the one you found couldn't happen

File: cmd/oceanbench/main.go [hunk 2]
```diff
@@ -72,7 +72,7 @@ import (
 )
 
 const (
-	version     = "v0.22.11"
+	version     = "v0.22.12"
```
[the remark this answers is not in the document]
saxon-milton: Maybe not quite a new feature, but i think big enough that it's a minor bump.

cmd/oceanbench/set.go
```diff
@@ -493,10 +505,43 @@ func configDevicesHandler(w http.ResponseWriter, r *http.Request) {
 		writeError(w, fmt.Errorf("unable to put rig system: %w", err))
 		return
 	}
+	site, err := model.GetSite(ctx, settingsStore, int64(skey))
```
saxon-milton, line 508: We should check the error, otherwise we end up overwriting it at the ```putProfileData``` call.

saxon-milton (approved):
LGTM after addressing comments

PR #306, opened 2024-08-26
Bench: Use api/get/sites/user to populate site menu

ao-david:
This change removes the need to call get sites before every page load, allowing the loading to happen completely async.

This should limit the number of database calls before returning the page to the user. This should make cloudblue feel like it loads slightly faster

### DIFF
--- a/cmd/oceanbench/admin.go
+++ b/cmd/oceanbench/admin.go
@@ -100,12 +100,6 @@ func adminHandler(w http.ResponseWriter, r *http.Request) {
 			},
 		}
 
-		data.Users, err = getUsersForSiteMenu(w, r, ctx, p, data)
-		if err != nil {
-			writeTemplate(w, r, "register.html", &data, fmt.Sprintf("could not populate site menu: %v", err.Error()))
-			return
-		}
-
 		switch r.Method {
 		case "GET":
 			writeTemplate(w, r, "register.html", &data, "")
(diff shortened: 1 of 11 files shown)


scruzin (approved):
LGTM, except `getUsersForSiteMenu` should probably just be renamed `getUserSites` (since it is now used via the API, and not necessarily for UI purposes).

Similarly, a function such as `getUserSites` should probably live in the `model` package.

ao-david:
> LGTM, except `getUsersForSiteMenu` should probably just be renamed `getUserSites` (since it is now used via the API, and not necessarily for UI purposes).
> 
> Similarly, a function such as `getUserSites` should probably live in the `model` package.

Should I move the function to model, called `getUserSites` then call this from /api?

scruzin:
> > LGTM, except `getUsersForSiteMenu` should probably just be renamed `getUserSites` (since it is now used via the API, and not necessarily for UI purposes).
> > Similarly, a function such as `getUserSites` should probably live in the `model` package.
> 
> Should I move the function to model, called `getUserSites` then call this from /api?

Yes, might as well do it now.

ao-david:
> > > LGTM, except `getUsersForSiteMenu` should probably just be renamed `getUserSites` (since it is now used via the API, and not necessarily for UI purposes).
> > > Similarly, a function such as `getUserSites` should probably live in the `model` package.
> > 
> > 
> > Should I move the function to model, called `getUserSites` then call this from /api?
> 
> Yes, might as well do it now.

Done

scruzin (changes requested):
Just use the email instead of the profile in GetUserSites.

ao-david:
> Just use the email instead of the profile in GetUserSites.

Done

PR #316, opened 2024-09-25
Add GetLatestScalar to model package

saxon-milton:
This was used in several places, so it made sense to generalise it. This will also be used for the broadcast voltage handling.


ao-david (approved):
LGTM

PR #319, opened 2024-09-30
Bench: Add last reported time to monitor page (#281)

### LINKED ISSUE #281: Bench: controller last reported feature on monitor page
while looking into the rapid bay upgrade situation, I sought to first establish that the status of our existing infrastructure was still consistent with our last understanding. Unfortunately it was not reporting when I checked, and I had to go through the search function to determine when it last reported data. it turned out to be July 19th, which was over a week ago at the time of writing, and so obviously  the 'last 7 days' search function didn't display any kind of data, but I wasn't sure if this was a bug with the search function or if there was just no data reported in that time period. it took a bit of searching to determine when it last reported data, and although it's not something we deal with regularly, it would be really useful if on the monitor page there was a 'last reported on...' feature. it may also be useful to factor that into some kind of alert if a controller hasn't reported in a few days or so, which would almost certainly be indicative of a critical failure

ao-david:
This change adds the last reported time to the monitor page for any devices that are not currently reporting (ie. red).

This change also sorts the monitor page so that the most recently reported devices are first in the list.

image

closes #281

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -72,7 +72,7 @@ import (
 )
 
 const (
-	version     = "v0.23.1"
+	version     = "v0.24.0"
 	localSite   = "localhost"
 	localDevice = "localdevice"
 	localEmail  = "localuser@localhost"
--- a/cmd/oceanbench/monitor.go
+++ b/cmd/oceanbench/monitor.go
@@ -34,7 +34,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


cmd/oceanbench/main.go
```diff
@@ -72,7 +72,7 @@ import (
 )
 
 const (
-	version     = "v0.22.11"
+	version     = "v0.22.12"
```
saxon-milton, line 75: Minor bump for this one

cmd/oceanbench/monitor.go
```diff
@@ -64,15 +65,16 @@ type sensorData struct {
...
-	Count      int // Number of scalars sent in the monitor period.
-	MaxCount   int // Max number of scalars that could be sent.
-	Throughput int // Percentage of successful scalars.
-	Sensors    []sensorData
+	Device       model.Device
+	Address      string
+	Sending      string
+	StatusText   string
+	Uptime       string
+	LastReported string
```
saxon-milton, line 73: I kind of think this should be a ```Time```

saxon-milton (changes requested):

trekhopton (approved):
I agree with Saxon's comments, otherwise approved.

cmd/oceanbench/monitor.go
ao-david: This is only used to populate the template. Having it as a string means that we can control the formatting better. Unless there's a better way of doing this in the template?

saxon-milton: You can have a method that formats it and call that in the template :nerd_face:

ao-david: I have used unix timestamps instead, as these are already compatible with our exported template methods (localdatetime)

saxon-milton (approved):

PR #320, opened 2024-09-30
Bench: Redirect all cron pages to base path (#214)

### LINKED ISSUE #214: bench: Saving a cron the navigating to another site saves the cron there too
If you save a cron then change site, bench saves the cron at the newly selected site too. This is highly undesirable and could case lots of issues if the user isn't aware of this.

I think this is because the site menu reloads the page but because the cron page submits a form on reload, the new cron form is submitted to the new site as soon as it loads.

ao-david:
This change redirects any requests to /set/crons/* back to /set/crons/ on a call to writeCrons. This prevents resubmission of forms when changing sites.

closes #214

### DIFF
--- a/cmd/oceanbench/set.go
+++ b/cmd/oceanbench/set.go
@@ -762,6 +762,11 @@ func writeCrons(w http.ResponseWriter, r *http.Request, msg string) {
 		http.Redirect(w, r, "/", http.StatusUnauthorized)
 		return
 	}
+	if r.URL.Path != "/set/crons/" {
+		// Redirect all requests to the cron base path to clear any actions.
+		http.Redirect(w, r, "/set/crons/", http.StatusFound)
+		return
+	}
 
 	ctx := r.Context()
 	setup(ctx)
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/set.go
```diff
@@ -713,6 +713,11 @@ func writeCrons(w http.ResponseWriter, r *http.Request, msg string) {
 		http.Redirect(w, r, "/", http.StatusUnauthorized)
 		return
 	}
+	if r.URL.Path != "/set/crons/" {
+		// Redirect all requests to the cron base path to clear any actions.
+		http.Redirect(w, r, "/set/crons/	", http.StatusFound)
```
saxon-milton, line 718: There's a big space in the path ?

saxon-milton (changes requested):

cmd/oceanbench/set.go
ao-david: Hmmm... Not sure how that happened. Will fix

saxon-milton (approved):

Issue #113, opened 2024-06-12
OceanBench: Use dropdown for cron variables.


ao-david:
When adding a cron via the cron interface, the user should be able to select a variable from a range of variables on the site page. This can be limited to only the set type to ensure that other cron types are not limited to just these variables.

PR #321, opened 2024-09-30
Bench: Add /get/devices/site and /get/variables/site endpoints (#113)

ao-david:
This change adds two endpoints to the API.

- /get/variables/site returns all of the variables for the site that are tied to a device.
- /get/devices/site returns all of the devices for a given site.

These changes allow us to make more async calls after a page has done an initial load.

### DIFF
--- a/cmd/oceanbench/api.go
+++ b/cmd/oceanbench/api.go
@@ -91,6 +91,43 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, string(enc))
 			return
 
+		case "devices":
+			switch val {
+			case "site":
+				// Check that the user has at least read access.
+				if len(strings.Split(p.Data, ":")) != 2 {
+					writeHttpError(w, http.StatusBadRequest, "no site data in profile")
+					return
+				}
+				skey, err := strconv.ParseInt(strings.Split(p.Data, ":")[0], 10, 64)
+				if err != nil {
+					writeHttpError(w, http.StatusBadRequest, "invalid site data in profile data: %s", p.Data)
+					return
+				}
+				user, err := model.GetUser(ctx, settingsStore, skey, p.Email)
+				if err != nil {
+					writeHttpError(w, http.StatusInternalServerError, "unable to get user: %v", err)
+					return
+				}
+				if user.Perm&model.ReadPermission == 0 {
+					writeHttpError(w, http.StatusUnauthorized, "profile does not have read permissions")
+					return
+				}
+
+				devs, err := model.GetDevicesBySite(ctx, settingsStore, skey)
+				if err != nil {
+					writeHttpError(w, http.StatusInternalServerError, "unable to get devices by site: %v", err)
+					return
+				}
+				data, err := json.Marshal(devs)
+				if err != nil {
+					writeHttpError(w, http.StatusInternalServerError, "unable to marshal devs into json: %v", err)
+					return
+				}
+				w.Write(data)
+				return
+			}
+
 		case "sites":
 			if val == "user" {
 			}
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/api.go
```diff
@@ -91,6 +91,40 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, string(enc))
 			return
 
+		case "devices":
+			switch val {
+			case "site":
+				// Check that the user has at least write access.
+				if len(strings.Split(p.Data, ":")) != 2 {
+					fmt.Fprint(w, "no site data in profile")
```
trekhopton, line 99: Perhaps use writeHttpError instead of just writing a string as a response. Here and in all places an error is sent back below.

cmd/oceanbench/api.go
```diff
@@ -91,6 +91,40 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, string(enc))
 			return
 
+		case "devices":
+			switch val {
+			case "site":
+				// Check that the user has at least write access.
```
trekhopton, line 97: You mean _read_ permission?

cmd/oceanbench/api.go
```diff
@@ -143,6 +177,57 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 				fmt.Fprint(w, p.Data)
 				return
 			}
+
+		case "vars":
+			switch val {
+			case "site":
+				// Check that the user has at least write access.
```
trekhopton, line 184: *read

trekhopton (approved):
Small comments to address. Approved in advance.

cmd/oceanbench/api.go
```diff
@@ -91,6 +91,40 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, string(enc))
 			return
 
+		case "devices":
+			switch val {
```
saxon-milton, line 95: Should we use a switch for a single case ? Should we just do
```
if val != "site" {
    break
}
```

and then we can reduce the amount of nested code.

cmd/oceanbench/api.go
```diff
@@ -143,6 +177,57 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 				fmt.Fprint(w, p.Data)
 				return
 			}
+
+		case "vars":
+			switch val {
```
saxon-milton, line 182: Similar to above

cmd/oceanbench/api.go
```diff
@@ -143,6 +177,57 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
+				if err != nil {
+					writeHttpError(w, http.StatusBadRequest, "invalid site data in profile data: %s", p.Data)
+					return
+				}
+				user, err := model.GetUser(ctx, settingsStore, skey, p.Email)
+				if err != nil {
+					writeHttpError(w, http.StatusInternalServerError, "unable to get user: %v", err)
+					return
+				}
+				if user.Perm&model.ReadPermission != 0 {
```
saxon-milton, line 199: Just check for ```user.Perm&model.ReadPermission == 0``` and leave early so we don't have to nest.

cmd/oceanbench/api.go
```diff
@@ -91,6 +91,40 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
```
saxon-milton, line 112: Just check for user.Perm&model.ReadPermission == 0 and leave early so we don't have to nest.

saxon-milton (changes requested):
Just a few things on code structure

cmd/oceanbench/api.go
ao-david: I've done it like this with the intention that the api will grow over time and there will be more endpoints which can just be added as cases to the existing switch statement without having to adjust the current format. This also keeps it more inline with other endpoints.

ao-david:
> Just a few things on code structure

All addressed. I think that leaving them as switch statements is better for future extensibility however.

cmd/oceanbench/api.go
saxon-milton: IMO we should just change the structure when that happens. I don't think code needs to be extendable to the point that you choose a ```switch``` over ```if``` :laughing: .

trekhopton: I don't think the level of nesting is that bad, it might actually help follow the API endpoints better. I would probably do the same as David to save changing it if it's extended in the future aha

PR #322, opened 2024-09-30
Bench: Use dropdowns for cron variables (#113)

### LINKED ISSUE #113: OceanBench: Use dropdown for cron variables.

ao-david:
This changes allows the user to select from all of the variables on the site when setting a cron.

This also shows a human readable variable name, using the name of the device rather than the mac address. (the value is unchanged, and so all functionality is the same)

image

requires #321 
closes #113

### DIFF
--- a/cmd/oceanbench/s/cron.js
+++ b/cmd/oceanbench/s/cron.js
@@ -1,0 +1,49 @@
+window.addEventListener("load", init());
+
+async function init() {
+  // Get all variables for the site.
+  let devicesPromise = fetch("/api/get/devices/site");
+  let varsPromise = await fetch("/api/get/vars/site");
+
+  // Get select elements.
+  let varSelects = document.getElementsByName("cv");
+
+  let [devicesRaw, varsRaw] = await Promise.all([devicesPromise, varsPromise]);
+
+  const vars = await varsRaw.json();
+  const devs = await devicesRaw.json();
+
+  vars.forEach((v) => {
+    let mac = v.Name.split(".")[0];
+    let device = devs.find((d) => {
+      return d.Mac === parseInt(mac, 16);
+    });
+
+    // Add the variables to each dropdown.
+    varSelects.forEach((element) => {
+      if (v.Name == element.value) {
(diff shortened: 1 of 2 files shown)


trekhopton (approved):
Great work!

cmd/oceanbench/api.go
```diff
@@ -91,6 +91,40 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, string(enc))
 			return
 
+		case "devices":
+			switch val {
```
saxon-milton, line 95: Same feedback as other PR

cmd/oceanbench/api.go
```diff
@@ -143,6 +177,57 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 				fmt.Fprint(w, p.Data)
 				return
 			}
+
+		case "vars":
+			switch val {
```
saxon-milton, line 182: ditto

cmd/oceanbench/api.go
```diff
@@ -143,6 +177,57 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
+				if err != nil {
+					fmt.Fprintf(w, "invalid site data in profile data: %s", p.Data)
+					return
+				}
+				user, err := model.GetUser(ctx, settingsStore, skey, p.Email)
+				if err != nil {
+					fmt.Fprintf(w, "unable to get user: %v", err)
+					return
+				}
+				if user.Perm&model.ReadPermission != 0 {
```
saxon-milton, line 199: ditto

saxon-milton (changes requested):

cmd/oceanbench/api.go
ao-david: Not quite sure why they appear in this PR as well. The base branch of this PR should be that other PR and these changes shouldn't show here. I will make sure to fix before merging.

Issue #324, opened 2024-10-04
tv / bench: Support registering youtube streams with OpenFish automatically


scott97:
Add an option to oceanbench / tv to configure a broadcast to automatically register with OpenFish once the broadcast completes.

PR #325, opened 2024-10-04
tv / bench: Support registering youtube streams with OpenFish automatically

scott97:
Closes https://github.com/ausocean/cloud/issues/324

Automatic stream registration can be configured under advanced settings of broadcasts:

image

When a stream finishes, oceantv will make a request to OpenFish to add the videostream.

Authentication is done using oauth2, much the same way as we authenticate for datastore. Registering the stream is done as the vidgrind service account, which has been added as a user and given permissions in OpenFish.

A new env variable has been added: OPENFISH_OAUTH2_CLIENT_ID, which is needed because the JWT token needs its audience ("aud") to be set. OpenFish / IAP will reject any requests without an audience.


cmd/oceantv/openfish/openfish.go
```diff
@@ -0,0 +1,131 @@
+/*
+DESCRIPTION
+  openfish.go provides functions for registering a completed stream with
+  Openfish.
```
trekhopton, line 4: Openfish -> OpenFish*

trekhopton (approved):
Nice work! LGTM

PR #328, opened 2024-10-10
Extend hardwareManager interface

saxon-milton:
Depends on PR #327

This change extends the hardwareManager interface by adding voltage, alarmVoltage and isUp methods. These additions have been motivated by voltage handling in the broadcast functionality.

PR #329, opened 2024-10-11
Add slate options

saxon-milton:
Depends on PR #328 

This change modifies the ForwardingService interface's Slate method to accept SlateOptions. We've added an option WithType that will allow us to specify the type of slate. This has been motivated by the fact that we'll have a permanent broadcast low voltage state and may wish to indicate this by a particular slate.


scruzin (approved):
LGTM.

trekhopton (approved):

PR #331, opened 2024-10-12
Add missing mapping for invalidConfigurationEvent

saxon-milton:
There was no mapping for invalidConfigurationEvent in the stringToEvent method.

PR #332, opened 2024-10-12
Add hardware voltage logic to broadcasting

saxon-milton:
Depends on PR #331 

Up until this point the broadcasting logic has had no awareness of hardware voltage. Therefore failures to broadcast, or issues during broadcasting that were due to low voltage, could not be handled gracefully.

This change attempts to improve matters by checking the hardware voltage when starting a broadcast, or in the case of permanent broadcasts, when transitioning from slate to live.

In the event that the check indicates we have low voltage, we enter a hardware voltage recovery state. In the case of permanent broadcasting, we can also indicate a low voltage state through setting a low voltage slate to display.

During recovery, timeEvents invoke voltage checks. Once we reach the required voltage we assume normal function. If a timeout is exceeded, we enter a failure state.

The behaviour for controller or charging fault could be improved for permanent broadcasts. For example, slate modes could be added for these conditions as well. We also need more testing for these paths too.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -110,44 +110,47 @@ type Settings struct {
 
 // BroadcastConfig holds configuration data for a YouTube broadcast.
 type BroadcastConfig struct {
-	SKey              int64         // The key of the site this broadcast belongs to.
-	Name              string        // The name of the broadcast.
-	ID                string        // Broadcast identification.
-	SID               string        // Stream ID for any currently associated stream.
-	CID               string        // ID of associated chat.
-	StreamName        string        // The name of the stream we'll bind to the broadcast.
-	Description       string        // The broadcast description shown below viewing window.
-	Privacy           string        // Privacy of the broadcast i.e. public, private or unlisted.
-	Resolution        string        // Resolution of the stream e.g. 1080p.
-	StartTimestamp    string        // Start time of the broadcast in unix format.
-	Start             time.Time     // Start time in native go format for easy operations.
-	EndTimestamp      string        // End time of the broadcast in unix format.
-	End               time.Time     // End time in native go format for easy operations.
-	VidforwardHost    string        // Host address of vidforward service.
-	CameraMac         int64         // Camera hardware's MAC address.
-	ControllerMAC     int64         // Controller hardware's MAC adress (controller used to power camera).
-	OnActions         string        // A series of actions to be used for power up of camera hardware.
-	OffActions        string        // A series of actions to be used for power down of camera hardware.
-	RTMPVar           string        // The variable name that holds the RTMP URL and key.
-	Active            bool          // This is true if the broadcast is currently active i.e. waiting for data or currently streaming.
-	Slate             bool          // This is true if the broadcast is currently in slate mode i.e. no camera.
-	Issues            int           // The number of successive stream issues currently experienced. Reset when good health seen.
-	SendMsg           bool          // True if sensor data will be sent to the YouTube live chat.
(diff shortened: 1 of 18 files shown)


cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -34,20 +35,67 @@ func newHardwareStarting(ctx *broadcastContext) *hardwareStarting {
...
-	if s.cfg.ControllerMAC != 0 {
-		controllerIsOn, err := model.DeviceIsUp(context.Background(), settingsStore, model.MacDecode(s.cfg.ControllerMAC))
-		if err != nil {
-			s.log("error getting controller status: %v", err)
-			s.bus.publish(controllerFailureEvent{})
-			return
-		} else if !controllerIsOn {
-			s.log("controller not responding")
-			s.bus.publish(controllerFailureEvent{})
+	if s.cfg.ControllerMAC == 0 {
```
saxon-milton, line 38: Should comment this

cmd/oceanbench/broadcast.go
```diff
@@ -110,44 +110,47 @@ type Settings struct {
 
 // BroadcastConfig holds configuration data for a YouTube broadcast.
 type BroadcastConfig struct {
```
scruzin, line 112: Any reason why we cannot define BroadcastConfig in one place, and share it between Bench and TV?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -61,6 +109,36 @@ func (s *hardwareStarting) timedOut(t time.Time) bool {
...
+	}
+}
+
+func (s *hardwareRecoveringVoltage) enter() {
+	s.LastEntered = time.Now()
+}
+
+func sanatisedVoltageRecoveryTimeout(ctx *broadcastContext) int {
+	// If VoltageRecoveryTimeout is not set, default to 4 hours.
+	if ctx.cfg.VoltageRecoveryTimeout == 0 {
```
scruzin, line 131: Where is VoltageRecoveryTimeout set?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -166,7 +250,37 @@ func (sm *hardwareStateMachine) handleTimeEvent(t timeEvent) {
...
+		voltage, err := sm.ctx.camera.voltage(sm.ctx)
+		if err != nil {
+			msg := fmt.Sprintf("could not get hardware voltage: %v", err)
+			sm.log(msg)
+			sm.ctx.bus.publish(invalidConfigurationEvent{msg})
+			return
+		}
+
+		// If RequiredStreamingVoltage is not set, default to 24.5.
+		if sm.ctx.cfg.RequiredStreamingVoltage == 0 {
```
scruzin, line 271: Ditto. where is RequiredStreamingVoltage set?

scruzin (approved):
AFAICT, LGTM.

cmd/oceanbench/broadcast.go
saxon-milton: Can't really think of good reason. It has been annoying keeping them in sync.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: This will be available under the advanced broadcast settings.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Also will be available under the advanced broadcast settings.

trekhopton (approved):

Issue #143, opened 2024-06-18
bench: Use mode=Shutdown to power down Pi devices


scruzin:
It is preferable to always shutdown cameras before turning off the power, which is accomplished by setting the **mode** to value **Shutdown** instead of **Paused**.

For example, the broadcast off action should be:

`<camera-mac>.mode=Shutdown, <controller-mac>.Power=false`

Issue #31, opened 2024-05-18
broadcast.go: modify how device shutdown is performed at broadcast end


saxon-milton:
We’d like to start safely shutting down devices i.e. perform an OS shutdown before cutting power. To do this, we’ll need to evolve the methodology currently in use by the broadcast functionality for dealing with shutdown of devices, that is, a list of variable-value pairs that indicate what variables to set and to what values on shutdown time. 

Up until this point, this has generally been setting of the ESP’s pin for device power control to false i.e. a hard powering off of the device. In doing so, we risk corrupting the device. It’s therefore proposed that the broadcast functionality instead uses a “Shutdown” response code, as available through the shutdown button on the devices page. We may need to modify this functionality so that not only does an OS shutdown happen, but additionally powering off of the peripheral power supply through the ESP. For this purpose, it’s proposed that we add a variable to devices (unused by the device) that contains the ID of the ESP pin that controls power for the device. This variable we’ll be used to set the ESP pin to false after the OS shutdown has occurred.

If we enclose this functionality into a function, the crons functionality could also adopt this i.e. add as a “call” action option. Then, even crons can safely power down devices.

scruzin:
Setting the `mode` var to `Shutdown` can already be used to perform a graceful shutdown. The missing functionality is waiting for the shutdown to take effect before turning off the relay that powers the device. Therefore, introducing the notion of a **pause** between camera on/off actions might be sufficient.

That said, having a 'shutdown' function that could be called either by the broadcast manager or crons alike is also attractive.

trekhopton:
Fixed in #427 . Camera's syncreboot command must be executable. This is the case in the latest camera image.

PR #336, opened 2024-10-31
tv: perform off actions for camera first, then controller

trekhopton:
This was done to allow the camera to shutdown gracefully before the controller powers the device off.

See #31 and #143.

### DIFF
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -30,6 +30,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
@@ -307,14 +308,67 @@ func extStart(ctx context.Context, cfg *BroadcastConfig, log func(string, ...int
 
 // extStop uses the OffActions in the provided broadcast config to perform
 // external streaming hardware shutdown.
+// If an OffAction is a controller action, we don't perform it yet.
 func extStop(ctx context.Context, cfg *BroadcastConfig, log func(string, ...interface{})) error {
 	if cfg.OffActions == "" {
 		return nil
 	}
 
-	err := setActionVars(ctx, cfg.SKey, cfg.OffActions, settingsStore, log)
-	if err != nil {
-		return fmt.Errorf("could not set device variables to end stream: %w", err)
+	vars := strings.Split(cfg.OffActions, ",")
+	if len(vars) == 0 {
+		return errors.New("no off actions to perform")
+	}
+
+	for _, v := range vars {
+		// Get the off action MAC to see if it's for a controller, if so don't perform it yet.
+		parts := strings.Split(v, "=")
+		if len(parts) != 2 {
+			return fmt.Errorf("bad off action format: %s", v)
+		}
+		parts = strings.Split(parts[0], ".")
+		if len(parts) != 2 {
+			return fmt.Errorf("bad off action variable name format: %s", v)
+		}
+		if model.MacEncode(parts[0]) != cfg.ControllerMAC {
+			err := setActionVars(ctx, cfg.SKey, v, settingsStore, log)
+			if err != nil {
+				return fmt.Errorf("could not perform off actions: %w", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+// extStopController uses the OffActions in the provided broadcast config to perform
+// controller off actions.
+// This is intended to be run after the camera has had a chance to shutdown properly.
+func extStopController(ctx context.Context, cfg *BroadcastConfig, log func(string, ...interface{})) error {
+	if cfg.OffActions == "" {
+		return nil
+	}
+
+	vars := strings.Split(cfg.OffActions, ",")
+	if len(vars) == 0 {
+		return errors.New("no off actions to perform")
+	}
+
+	for _, v := range vars {
+		// Get the off action MAC to see if it's for a controller, if so perform the action.
+		parts := strings.Split(v, "=")
+		if len(parts) != 2 {
+			return fmt.Errorf("bad off action format: %s", v)
+		}
+		parts = strings.Split(parts[0], ".")
+		if len(parts) != 2 {
+			return fmt.Errorf("bad off action variable name format: %s", v)
+		}
+		if model.MacEncode(parts[0]) == cfg.ControllerMAC {
+			err := setActionVars(ctx, cfg.SKey, v, settingsStore, log)
+			if err != nil {
+				return fmt.Errorf("could not perform off actions for controller: %w", err)
+			}
+		}
 	}
 
 	return nil
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -143,9 +143,13 @@ func sanatisedVoltageRecoveryTimeout(ctx *broadcastContext) int {
 
 type hardwareStopping struct {
 	*broadcastContext `json:"-"`
+	stateWithTimeoutFields
+	PreviouslyTimedOut bool
 }
 
-func newHardwareStopping(ctx *broadcastContext) *hardwareStopping { return &hardwareStopping{ctx} }
+func newHardwareStopping(ctx *broadcastContext) *hardwareStopping {
+	return &hardwareStopping{broadcastContext: ctx}
+}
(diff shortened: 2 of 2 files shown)


cmd/oceantv/broadcast.go
```diff
@@ -307,14 +308,67 @@ func extStart(ctx context.Context, cfg *BroadcastConfig, log func(string, ...int
...
-	err := setActionVars(ctx, cfg.SKey, cfg.OffActions, settingsStore, log)
-	if err != nil {
-		return fmt.Errorf("could not set device variables to end stream: %w", err)
+	vars := strings.Split(cfg.OffActions, ",")
+	if len(vars) == 0 {
+		return errors.New("no off actions to perform")
+	}
+
+	for _, v := range vars {
+		// Get the off action MAC to see if it's for a controller, if so don't perform it yet.
```
ao-david, line 323: Small detail, but its checking if its for _THE_ controller registered to a broadcast, not just any controller.

cmd/oceantv/broadcast.go
```diff
@@ -307,14 +308,67 @@ func extStart(ctx context.Context, cfg *BroadcastConfig, log func(string, ...int
...
 	if cfg.OffActions == "" {
 		return nil
 	}
 
-	err := setActionVars(ctx, cfg.SKey, cfg.OffActions, settingsStore, log)
-	if err != nil {
-		return fmt.Errorf("could not set device variables to end stream: %w", err)
+	vars := strings.Split(cfg.OffActions, ",")
+	if len(vars) == 0 {
+		return errors.New("no off actions to perform")
```
ao-david, line 319: Since we already check that OffActions are not nil :311, its not that there are no off actions, but there are no valid off actions.

ao-david: Actually, is it even possible for this to fail if the actions != ""?

cmd/oceantv/broadcast.go
```diff
@@ -307,14 +308,67 @@ func extStart(ctx context.Context, cfg *BroadcastConfig, log func(string, ...int
...
+// extStopController uses the OffActions in the provided broadcast config to perform
+// controller off actions.
+// This is intended to be run after the camera has had a chance to shutdown properly.
+func extStopController(ctx context.Context, cfg *BroadcastConfig, log func(string, ...interface{})) error {
+	if cfg.OffActions == "" {
+		return nil
+	}
+
+	vars := strings.Split(cfg.OffActions, ",")
+	if len(vars) == 0 {
```
ao-david, line 352: Again, not sure that this can ever fail?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -238,22 +242,39 @@ func (sm *hardwareStateMachine) handleTimeEvent(t timeEvent) {
...
 			sm.transition(newHardwareOff())
 			return
 		}
 		eventIfStatus(hardwareStartedEvent{}, true)
 	case *hardwareStopping:
+		if state.timedOut(t.Time) {
+			if state.PreviouslyTimedOut {
+				sm.log("hardwareStopping state timed out for the second time")
+				sm.ctx.bus.publish(hardwareStopFailedEvent{})
+				sm.transition(newHardwareOn())
```
ao-david, line 258: Why are we transitioning back to HardwareOn?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -238,22 +242,39 @@ func (sm *hardwareStateMachine) handleTimeEvent(t timeEvent) {
...
+			sm.log("hardwareStopping state timed out, performing controller off actions")
+			err := extStopController(context.Background(), sm.ctx.cfg, sm.ctx.log)
+			if err != nil {
+				sm.log("could not perform controller off actions: %v", err)
+				sm.ctx.bus.publish(hardwareStopFailedEvent{})
+				sm.transition(newHardwareOn())
+				return
+			}
+			state.PreviouslyTimedOut = true
+			state.reset(5 * time.Minute) // 5 minutes is the initial timeout default, so we just reset it here.
```
ao-david, line 270: If this is a default somewhere, is there a constant that this can reference?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -238,22 +242,39 @@ func (sm *hardwareStateMachine) handleTimeEvent(t timeEvent) {
...
 	case *hardwareStopping:
+		if state.timedOut(t.Time) {
+			if state.PreviouslyTimedOut {
+				sm.log("hardwareStopping state timed out for the second time")
+				sm.ctx.bus.publish(hardwareStopFailedEvent{})
+				sm.transition(newHardwareOn())
+				return
+			}
+			sm.log("hardwareStopping state timed out, performing controller off actions")
+			err := extStopController(context.Background(), sm.ctx.cfg, sm.ctx.log)
```
ao-david, line 262: Is this supposed to run after the first timeout or the second timeout?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -238,22 +242,39 @@ func (sm *hardwareStateMachine) handleTimeEvent(t timeEvent) {
...
+			if err != nil {
+				sm.log("could not perform controller off actions: %v", err)
+				sm.ctx.bus.publish(hardwareStopFailedEvent{})
+				sm.transition(newHardwareOn())
```
ao-david, line 266: Why are we transitioning to hardwareon?

cmd/oceantv/broadcast.go
saxon-milton, line 323: Maybe it would be better if we had two separate fields. One for the shutdown of actions and one for power off, then we wouldn't have to do these gymnastics to find out what's for the controller vs shutdown.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -143,9 +143,13 @@ func sanatisedVoltageRecoveryTimeout(ctx *broadcastContext) int {
 
 type hardwareStopping struct {
 	*broadcastContext `json:"-"`
+	stateWithTimeoutFields
+	PreviouslyTimedOut bool
```
saxon-milton, line 147: I'm not a massive fan of this addition. It's basically creating a substate of hardwareStopping. We could instead look at splitting hardwareStopping into two states e.g. hardwareShuttingDown and hardwarePoweringOff, then you wouldn't need to have the weird logic around whether this previously timed out or not.

saxon-milton (changes requested):
Basically doing what we need it do, but I'm wondering if we could make it less awkward by differentiating shutdown actions from power off actions, then naturally we could split hardwareStopping into shutdown and power off states.

cmd/oceantv/broadcast.go
trekhopton: It's not possible, no. Will fix.

cmd/oceantv/broadcast.go
trekhopton: That could be good, but this should work for now.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: First timeout.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Good point. You're suggesting we stay in the stopping state, even if it fails. I think I agree. I'll have a think about this.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: There isn't. Could make it so though.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: True. I'll have a look into that.

cmd/oceantv/broadcast.go
saxon-milton: I think this would dovetail nicely with the two state change FYI

ao-david:
Has this been made redundant by #427?
@saxon-milton

trekhopton:
> Has this been made redundant by #427? @saxon-milton

Yeah it has, I'll close it.

trekhopton:
Implemented by PR #427 instead.

Issue #342, opened 2024-11-08
ausoceantv: Setup basic subscriber flow


ao-david:
This should add a simple set of pages to mimic how the flow will be to a user signing up to a subscription.

PR #343, opened 2024-11-09
AusOceanTV: Initialise Stripe (#342)

ao-david:
This change adds the most basic Stripe payment integration to the /checkout.html.

If anyone has feedback regarding how to properly handle CORS, and addresses used between Vite and Go, please let me know. It currently works in my local environment, but not sure how it will work in production.

EDIT: I am now using a Vite Proxy which should handle most of the CORS issues on advice from @scott97 

# TODO
This change will not be functional until Stripe keys are added. There will need to be one key stored as an env var ```AUSOCEAN_STRIPE_SECRET_KEY``` as well as the public  key added to the checkout.js file.

Additionally, the flow will need to create a subscription to attach to the paymentIntent object, so that the flow works correctly.

# Reference
Most of this code is boilerplate code supplied by Stripe, view the documentation [here](https://docs.stripe.com/payments/quickstart)

Part of #342

image


cmd/ausoceantv/main.go
```diff
@@ -66,6 +71,14 @@ func main() {
 		}
 	}
 
+	// Get stripe secret key.
+	v = os.Getenv("AUSOCEAN_STRIPE_SECRET_KEY")
+	if v == "" {
+		log.Println("AUSOCEAN_STRIPE_SECRET_KEY not found, cannot take payments")
```
ao-david, line 77: I suspect that we will probably use a secret file instead of a env var here

cmd/ausoceantv/main.go
```diff
@@ -120,6 +136,75 @@ func (svc *service) setup(ctx context.Context) {
...
+	}{
+		ClientSecret: pi.ClientSecret,
+		// [DEV]: For demo purposes only, you should avoid exposing the PaymentIntent ID in the client-side code.
+		// DpmCheckerLink: fmt.Sprintf("https://dashboard.stripe.com/settings/payment_methods/review?transaction_id=%s", pi.ID),
+	})
+}
+
+// enableCors allows the vite server to read responses from this webserver.
+func enableCors(w *http.ResponseWriter) {
+	(*w).Header().Set("Access-Control-Allow-Origin", "http://localhost:5173")
```
ao-david, line 187: This should also reflect the correct address of vite, which I think may be :4173 in production?

cmd/ausoceantv/src/checkout.css
```diff
@@ -0,0 +1,257 @@
+/* Variables */
```
ao-david, line 1: This is all straight from Stripe

cmd/ausoceantv/src/checkout.js
```diff
@@ -0,0 +1,102 @@
+// This is your test publishable API key.
+const stripe =
+  Stripe();
+  // TODO: Add Ausocean stripe public key.
```
ao-david, line 4: Need to add the public key.

cmd/ausoceantv/src/checkout.js
```diff
@@ -0,0 +1,102 @@
+// This is your test publishable API key.
```
ao-david, line 1: Again, all from Stripe

cmd/ausoceantv/main.go
saxon-milton: no need to dereference in Go FYI, it happens for you :)

cmd/ausoceantv/main.go
```diff
@@ -79,11 +92,14 @@ func main() {
 	ctx := context.Background()
 	app.setup(ctx)
 
-	// Serve static files when running locally
+	// Serve static files when running locally.
 	http.Handle("/s/", http.StripPrefix("/s", http.FileServer(http.Dir("s"))))
```
scott97, line 96: By serving the static files from the go program, we lose hot reloading which is very handy during development. Instead I suggest you serve your static files from vite, and use vite's proxy feature to proxy the go server, so that they can all be served from the same origin during development.

scott97 (changes requested):

cmd/ausoceantv/main.go
ao-david: I was only changing the comment here to put a fullstop, but I agree. I think that the static file serving was implemented by @scruzin before Trek implemented Vite on this service, I imagine that it is mostly a copy of our other services.

I have made an issue to address this in a separate PR: #352 

In terms of the Vite proxy, does this allow for calls to ```/stripe``` or ```/api``` to be redirected to golang without having issues with CORS?

scott97: Yes, sorry I just glanced at the changes briefly, I think I thought it was the line below you added. By proxying `/stripe` and `/api`  they will all be served from the same origin so CORS doesn't need to be considered during development. During production they either need to all be served from the same origin or you need to use the CORS headers

scott97 (approved):

trekhopton:
Since you have this marked as a draft, do you want me to review it yet?

cmd/ausoceantv/src/checkout.js
```diff
@@ -0,0 +1,93 @@
...
+const stripe = Stripe(import.meta.env.VITE_STRIPE_PUBLIC_KEY);
```
ao-david, line 2: This requires a .env file with the stripe public key

ao-david: Should this be committed? It is a public key, and will be different between development and production

ao-david:
> Since you have this marked as a draft, do you want me to review it yet?

I have now made the changes that I was waiting on. If you could review now that would be good :)

trekhopton:
A screenshot of the page would be good.

trekhopton:
Might need a go mod tidy, my IDE was complaining until I did that and it did this:

`-       github.com/stripe/stripe-go v70.15.0+incompatible`

cmd/ausoceantv/checkout.html
```diff
@@ -0,0 +1,25 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/png" href="/src/assets/ausocean_logo.png" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>AusOcean.TV</title>
```
trekhopton, line 7: Perhaps should we go with "AusOcean TV"?

PR #350, opened 2024-11-11
AusOceanTV:Use tailwind with lit

ao-david:
This change adds support for tailwind styling in lit elements using the Vite Dev tools and PostCSS.

See #351 for a developer readme to explain how this works.

### DIFF
--- a/cmd/ausoceantv/src/shared/tailwind.element.ts
+++ b/cmd/ausoceantv/src/shared/tailwind.element.ts
@@ -6,9 +6,9 @@ import style from "./tailwind.global.css?inline";
 
 const tailwindElement = unsafeCSS(style);
 
-export const TailwindElement = (style) =>
+export const TailwindElement = (style? : string) =>
     class extends LitElement {
 
-        static styles = [tailwindElement, unsafeCSS(style)];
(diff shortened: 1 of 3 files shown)


cmd/ausoceantv/src/shared/tailwind.element.ts
```diff
@@ -0,0 +1,14 @@
+// See Also: https://github.com/butopen/web-components-tailwind-starter-kit
+
+import {LitElement, unsafeCSS} from "lit";
+
+import style from "./tailwind.global.css?inline";
+
+const tailwindElement = unsafeCSS(style);
+
+export const TailwindElement = (style) =>
```
scott97, line 9: It would be good if TailwindElement took style as an optional argument, because for most of the time we would be able to do all the styling using tailwind and not need additional css written.

scott97 (changes requested):

cmd/ausoceantv/src/shared/tailwind.element.ts
ao-david: Good point. Done.

trekhopton (approved):

scott97 (approved):

Issue #263, opened 2024-07-26
bench: speed up graphing over long time periods


scruzin:
Search periods greater than a week long take excessively long to graph. We should discard more data (i.e., sample less) over long periods in order to speed up graphing.

ao-david:
There are 2 parts to this load time:
- data fetching period
- rendering period

Considering the data fetch period, there is already the concept of lower resolution, however, the backend code still gets all of the data, and only returns some of it, so this is basically the same between a resolution of 1pt per hour, and 60pts per hour. To improve the data fetch time, we could pipeline all of the requests, rather than send them serially, this would likely be more performant since golang is very multithreaded and so this could handle multiple requests quicker than it could handling 1 at a time.

In terms of the rendering period, it does take a long time to render a month of data at 60pts/hr, but decreasing the resolution to 1/hr keeps enough of the detail, but makes the render basically instant. Its also possible that there is a new version of the graphing package that we could use that would make this faster if we decided it was necessary.

PR #354, opened 2024-11-12
Bench: Pipeline requests for faster graphing (#263)

ao-david:
This change makes graphing requests batched and sent all at once, rather than sending them serially and waiting for a response before sending the next request.

This improves graphing time significantly.

TODO: This change does not change the resolution of the request, and so rendering the results can still take a long time. To make the rendering faster you must still request a lower resolution manually. It is proposed to use some form of algorithm (either on the frontend or backend) to automatically determine a reasonable resolution and only graph that many points.

Addresses #263


trekhopton (approved):

saxon-milton (approved):
nice

PR #358, opened 2024-11-15
ausoceantv: make make front end and API stub deploy-able to appengine

trekhopton:
ausoceantv with a vite dist and the API served by go now deploys to appengine.


scott97 (approved):
looks good

Issue #353, opened 2024-11-12
AusOceanTV: Initialise OAuth


ao-david:
users should be prompted to signup from the home page, and then have a cookie attached to their session to indicate their authentication

ao-david:
We have now decided to use go fiber to make it easier to reuse code from openfish, so this will take the implementation of middleware.

PR #363, opened 2024-11-20
AusOceanTV: Implement auth pipeline (#353)

ao-david:
This change uses the gauth package to implement login, logout, and callbacks for google authentication.

This change requires a secret file which holds the client secret, as well as a session key.

@scruzin I'm not quite sure what the session key needs to be? can it just be a random 64 character base64 string?


ao-david:
This has been closed in favour of using openfishes implementation with fiber.

PR #369, opened 2024-11-22
AOTV: Convert server to use Fiber (#366)

### LINKED ISSUE #366: AusOceanTV: Use go fiber
We want to use fiber to reduce boiler plate, and allow us to reuse code from openfish easier.

ao-david:
This change implements fiber to allow us to more easily share code with openfish.

I am not sure if implementing the service middleware is the best way of handling the service global, since this would require calling
```golang
svc := ctx.Locals("service").(*service)
```
in every handler which needs the service struct. Since the svc struct is already global, we _could_ revert to calling it from the global context.

closes #366


cmd/ausoceantv/main.go
```diff
@@ -53,8 +57,30 @@ type service struct {
...
+var svc *service = &service{}
+
+func registerAPIRoutes(app *fiber.App) {
+	v1 := app.Group("/api/v1")
+	v1.Get("version", versionHandler)
+}
+
+// errorHandler creates a HTTP response with the given status code or 500 by default.
+// The response body is JSON: {"message": "<error message here>"}
+func errorHandler(ctx *fiber.Ctx, err error) error {
```
trekhopton, line 70: I tagged openfish with version v0.1.6 that provides api.ErrorHandler

cmd/ausoceantv/main.go
```diff
@@ -68,27 +94,70 @@ func main() {
...
-	http.HandleFunc("/api/", app.apiHandler)
+	// Register routes.
+	registerAPIRoutes(app)
 
-	log.Printf("Listening on %s:%d", host, port)
-	log.Fatal(http.ListenAndServe(fmt.Sprintf("%s:%d", host, port), nil))
+	// Start web server.
+	listenOn := fmt.Sprintf(":%d", port)
+	fmt.Printf("starting web server on %s\n", listenOn)
+	app.Listen(listenOn)
```
trekhopton, line 144: We need to handle the error for Listen, perhaps wrap it in log.Fatal like it was before.

cmd/ausoceantv/main.go
```diff
@@ -68,27 +94,70 @@ func main() {
...
+}
+
+// ServiceMiddleware attaches the global service pointer to
```
trekhopton, line 147: lowercase func name in comment.

trekhopton (changes requested):
Few changes, otherwise, looks good.

trekhopton (approved):

PR #375, opened 2024-11-25
ausoceantv: add store to fiber's context.Locals

trekhopton:
This was done so the api package can use the store to create, get etc. from the store.


trekhopton:
Closed since it was working with code that no longer exists. The store can now be accessed in a different way.

PR #381, opened 2024-11-25
ausoceantv: add required code and structure to use datastore

trekhopton:
This was done because it is a prerequisite for interacting with the store entities via the API.


cmd/ausoceantv/main.go
```diff
@@ -144,23 +146,14 @@ func (svc *service) setup(ctx context.Context) {
...
-	if svc.standalone {
-		log.Info("Running in standalone mode")
-		svc.settingsStore, err = datastore.NewStore(ctx, "file", "vidgrind", svc.storePath)
-	} else {
-		log.Info("Running in App Engine mode")
-		svc.settingsStore, err = datastore.NewStore(ctx, "cloud", "netreceiver", "")
-	}
-	if err != nil {
-		log.Fatalf("could not set up datastore: %v", err)
+	if svc.settingsStore == nil {
```
ao-david, line 149: Could consider 
```go
if svc.settingsStore != nil {
    return
}
err := dsclient...

ao-david (approved):

cmd/ausoceantv/main.go
trekhopton: Changed it back.

Issue #385, opened 2024-11-26
Create Subscriber when new user logs in


ao-david:
When a new user logs in, we want to create a subscriber instance to store in the datastore.

PR #387, opened 2024-11-26
Model: Add subscriber create and get methods (#385)

ao-david:
This change adds the CreateSubscriber method, and the GetSubscriberByEmail methods.

### DIFF
--- a/model/subscriber.go
+++ b/model/subscriber.go
@@ -22,14 +22,20 @@ LICENSE
 package model
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"time"
 
+	"github.com/ausocean/cloud/utils"
 	"github.com/ausocean/openfish/datastore"
 )
 
 const (
 	typeSubscriber = "Subscriber" // Subscriber datastore type.
 )
+
+var errDuplicateSubscriberEmails = errors.New("more than one subscriber exists for a given email")
 
 // Subscriber is an entity in the datastore representing a user who subscribes to AusOcean.TV.
 type Subscriber struct {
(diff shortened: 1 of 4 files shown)


model/subscriber.go
```diff
@@ -64,3 +70,55 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+
+	return &subs[0], err
+}
+
+// GetSubscriber gets the subscriber with the given ID.
+func GetSubscriber(ctx context.Context, store datastore.Store, id int64) (*Subscriber, error) {
+	key := store.IDKey(typeSubscriber, id)
+	sub := &Subscriber{}
+	err := store.Get(ctx, key, sub)
+	return sub, err
```
trekhopton, line 108: Perhaps wrap the error with added context.

model/subscriber.go
```diff
@@ -64,3 +70,55 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+
+// NewSubscriberID returns a new unique in64 ID which does not already exist in
+// the datastore.
+func NewSubscriberID(ctx context.Context, store datastore.Store) int64 {
+	for {
+		// Create a random 31-bit number that is at least 10 digits.
+		id := rand.Int63n((1<<31)-1000000000) + 1000000000
+
+		// Check if the ID is unique.
+		_, err := GetSubscriber(ctx, store, id)
```
trekhopton, line 119: As mentioned above, this isn't concurrency safe.

trekhopton (changes requested):

model/subscriber.go
scruzin: You want to call `CreateSubscriber` until it is successful, as we do for sites, e.g.:

```
	for {
		// Create a random 31-bit number that is at least 10 digits.
		skey = rand.Int63n((1<<31)-1000000000) + 1000000000
		site := model.Site{Skey: skey, Name: sn, Public: false, Enabled: true}
		err := model.CreateSite(ctx, settingsStore, &site)
		if err == nil {
			break
		}
		if err != datastore.ErrEntityExists {
			return fmt.Errorf("cannot create site: %w", err)
		}
	}
```

Note that his still not concurrency safe in that it could result in two sites with the same name but different site keys. One site would be a zombie site that never gets used, though.

scruzin (changes requested):

model/subscriber.go
ao-david: I have now addressed this by having the Create function optionally generate a unique ID for the new subscriber through this method.

trekhopton (approved):

model/subscriber.go
```diff
@@ -64,3 +70,68 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
 func (s *Subscriber) GetCache() datastore.Cache {
 	return nil
 }
+
+// CreateSubscriber creates a new subscriber from the passed subscriber (s).
```
saxon-milton, line 74: Why ```subscriber (s)``` ? It seems like you can only provide 1 ?

model/subscriber.go
```diff
@@ -64,3 +70,68 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+	// Set the Created time.
+	s.Created = time.Now()
+
+	// If the subscriber has an ID, use that.
+	if s.ID != 0 {
+		key := store.IDKey(typeSubscriber, s.ID)
+		return store.Create(ctx, key, s)
+	}
+
+	// Otherwise generate and use a unique ID.
```
saxon-milton, line 92: Why not use https://pkg.go.dev/github.com/google/uuid ?

model/subscriber.go
```diff
@@ -64,3 +70,68 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+	}
+}
+
+// GetSubscriberByEmail returns the subscriber with the given email if it exists.
+func GetSubscriberByEmail(ctx context.Context, store datastore.Store, email string) (*Subscriber, error) {
+	q := store.NewQuery(typeSubscriber, false)
+	q.FilterField("Email", "=", email)
+	var subs []Subscriber
+	_, err := store.GetAll(ctx, q, &subs)
+
```
saxon-milton, line 111: Remove blank line

saxon-milton (changes requested):

model/subscriber.go
ao-david: The bracket is just indicating the variable name

PR #388, opened 2024-11-26
AOTV: Create new subscriber on signin (#385)

### LINKED ISSUE #385: Create Subscriber when new user logs in

ao-david:
When a new user signs in, check if there is an existing subscriber with the given email otherwise create a new subscriber.

closes #385

### DIFF
--- a/cmd/ausoceantv/auth.go
+++ b/cmd/ausoceantv/auth.go
@@ -25,10 +25,13 @@ LICENSE
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 
+	"cloud.google.com/go/datastore"
+	"github.com/ausocean/cloud/model"
 	"github.com/gofiber/fiber/v2"
 )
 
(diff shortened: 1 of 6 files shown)


cmd/ausoceantv/auth.go
```diff
@@ -44,7 +48,35 @@ func (svc *service) logoutHandler(c *fiber.Ctx) error {
...
+	p, err := svc.GetProfile(c)
+	if errors.Is(err, SessionNotFound) || errors.Is(err, TokenNotFound) {
+		return fiber.ErrUnauthorized
+	} else if err != nil {
+		return fmt.Errorf("unable to get profile: %w", err)
+	}
+
+	ctx := context.Background()
+
+	ID := model.NewSubscriberID(ctx, svc.settingsStore)
```
trekhopton, line 66: This will look different if ID is generated later by CreateSubscriber.

trekhopton (approved):

cmd/ausoceantv/auth.go
```diff
@@ -44,7 +47,33 @@ func (svc *service) logoutHandler(c *fiber.Ctx) error {
 
 // callbackHandler handles callbacks from google's oauth2 flow.
 func (svc *service) callbackHandler(c *fiber.Ctx) error {
-	return svc.CallbackHandler(c)
+	err := svc.CallbackHandler(c)
+	if err != nil {
+		return err
```
saxon-milton, line 52: wrap error ?

cmd/ausoceantv/auth.go
```diff
@@ -44,7 +47,33 @@ func (svc *service) logoutHandler(c *fiber.Ctx) error {
...
+	}
+
+	// Check if a user already exists.
+	p, err := svc.GetProfile(c)
+	if errors.Is(err, SessionNotFound) || errors.Is(err, TokenNotFound) {
```
saxon-milton, line 57: Would it be useful to have separate cases for these and then you can wrap the fiber error with more information i.e. whether it was a ```SessionNotFound``` or a ```TokenNotFound``` ?

Also, why exactly are we using the fiber error ?

cmd/ausoceantv/auth.go
```diff
@@ -44,7 +47,33 @@ func (svc *service) logoutHandler(c *fiber.Ctx) error {
 
 // callbackHandler handles callbacks from google's oauth2 flow.
 func (svc *service) callbackHandler(c *fiber.Ctx) error {
```
saxon-milton, line 49: I think the fiber context should be named like fCtx or something.

cmd/ausoceantv/auth.go
```diff
@@ -44,7 +47,33 @@ func (svc *service) logoutHandler(c *fiber.Ctx) error {
...
+
+	// Check if a user already exists.
+	p, err := svc.GetProfile(c)
+	if errors.Is(err, SessionNotFound) || errors.Is(err, TokenNotFound) {
+		return fiber.ErrUnauthorized
+	} else if err != nil {
+		return fmt.Errorf("unable to get profile: %w", err)
+	}
+
+	ctx := context.Background()
```
saxon-milton, line 63: So are there no ```context.Context``` in scope where ```callbackHandler``` is called that we could pass to this ?

saxon-milton (changes requested):

Issue #391, opened 2024-11-27
Add methods to ausoceantv entities


ao-david:
This change should add the methods for the ausoceantv types as defined [here](https://docs.google.com/document/d/12TYm5Al0G3T-R0wiWtX7-vrB0Yn5dH8VYRo8yWTujjY/edit?tab=t.0)

PR #392, opened 2024-11-28
model: Add methods for subscription and subscriber (#391)

ao-david:
This change implements the methods required for getting, creating and updating subscribers and subscriptions.

Also adds tests for the methods.

See the schema doc for details on these methods and types: https://docs.google.com/document/d/12TYm5Al0G3T-R0wiWtX7-vrB0Yn5dH8VYRo8yWTujjY/edit?tab=t.0#heading=h.mj06vfrpcpf5

requires #387

### DIFF
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -27,6 +27,7 @@ package model
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


model/subscription.go
```diff
@@ -62,3 +73,60 @@ func (s *Subscription) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+// CreateSubscription creates a subscription for a given subscriber ID and feed ID.
+func CreateSubscription(ctx context.Context, store datastore.Store, sid, fid int64, class, prefs string, renew bool) error {
+	// Calculate characteristics of the subscription.
+	start := time.Now().Truncate(time.Hour * 24) // Start the subscription at the start of the current day.
+	var end time.Time
+	switch class {
+	case SubscriptionDay:
+		end = start.AddDate(0, 0, 1)
+	case SubscriptionMonth:
+		end = start.AddDate(0, 1, 0)
```
trekhopton, line 106: Worth noting that this normalises the date, so it won't always be the same day in the following month. I don't think that's an issue.

trekhopton (approved):

scott97 (approved):

model/model_test.go
```diff
@@ -1324,6 +1331,144 @@ func testCron(t *testing.T, kind string) {
 	}
 }
 
+// testSubscriber tests Subscriber methods.
+func testSubscriber(t *testing.T, kind string) {
+	ctx := context.Background()
+	store, err := datastore.NewStore(ctx, kind, "vidgrind", "")
+	if err != nil {
+		t.Fatalf("could not create new store: %v", err)
```
saxon-milton, line 1339: Fatal _should_ be used here because you can't really do much else if you can't create the store, but with how ```TestNetreceiverFileAccess``` and ```TestNetreceiverCloudAccess``` call the subtests, it means that all subtest calls after a Fatal call won't execute. 

I think either each subtest needs to have its own test, or we actually use subtests i.e. call t.Run() for each test (https://pkg.go.dev/testing#T.Run). It may look something like this:
```
func getFuncName(i interface{}) string{
    return runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
}

for _, test := range []func(*testing.T, string){
    ...
    testSubscriber,
    testSubscription,
}{
    r.Run(getFuncName(test), func(t *testing.T){ test(t, "cloud") })
}
```

model/subscriber.go
```diff
@@ -64,3 +70,63 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+	return store.Create(ctx, key, s)
+}
+
+// GetSubscriberByEmail returns the subscriber with the given email if it exists.
+func GetSubscriberByEmail(ctx context.Context, store datastore.Store, email string) (*Subscriber, error) {
+	q := store.NewQuery(typeSubscriber, false)
+	q.FilterField("Email", "=", email)
+	var subs []Subscriber
+	_, err := store.GetAll(ctx, q, &subs)
+
```
saxon-milton, line 87: Remove blank line

model/subscriber.go
```diff
@@ -64,3 +70,63 @@ func (s *Subscriber) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+// of the passed subscriber.
+func UpdateSubscriber(ctx context.Context, store datastore.Store, subscriber *Subscriber) error {
+	key := store.IDKey(typeSubscriber, subscriber.ID)
+	_, err := store.Put(ctx, key, subscriber)
+	return err
+}
+
+// NewSubscriberID returns a new unique in64 ID which does not already exist in
+// the datastore.
+func NewSubscriberID(ctx context.Context, store datastore.Store) int64 {
```
saxon-milton, line 121: Is there a reason why we do it like this ? We could just use https://pkg.go.dev/github.com/google/uuid ?

model/subscription.go
```diff
@@ -62,3 +73,60 @@ func (s *Subscription) Copy(dst datastore.Entity) (datastore.Entity, error) {
...
+
+// GetSubscription gets a subscription for a given subscriberID (sid) and feedID (fid).
+func GetSubscription(ctx context.Context, store datastore.Store, sid, fid int64) (*Subscription, error) {
+	q := store.NewQuery(typeSubscription, false, "SubscriptionID", "FeedID")
+	q.FilterField("SubscriberID", "=", sid)
+	q.FilterField("FeedID", "=", fid)
+
+	var subscriptions []Subscription
+	_, err := store.GetAll(ctx, q, &subscriptions)
+
```
saxon-milton, line 85: Remove blank line

saxon-milton (changes requested):

model/model_test.go
ao-david: I think that these changes should be in a different PR. I have made the tests reflect the structure of the existing tests.

model/model_test.go
```diff
@@ -43,49 +44,51 @@ import (
 )
 
 const (
```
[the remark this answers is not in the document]
ao-david: Again, I think this can be a separate change.

model/model_test.go
saxon-milton: Yeah that's fair

PR #393, opened 2024-11-28
Bench: Fix Navmenu on configure page (#344)

### LINKED ISSUE #344: bench: init is undefined on configure page.
init is undefined on the device's configure page which means the navigation menu won't work.

Work around is edit the url to go where you want to go.

ao-david:
This change fixes the nav menu by ensuring the common data Page field is initialised correctly. This change also adds field validation for the configure page.

closes #344

### DIFF
--- a/cmd/oceanbench/s/configure.js
+++ b/cmd/oceanbench/s/configure.js
@@ -1,0 +1,19 @@
+function init() {
+    'use strict'
+
+    // Fetch all the forms we want to apply custom Bootstrap validation styles to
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/s/configure.js
```diff
@@ -0,0 +1,19 @@
+function init() {
```
ao-david, line 1: This is for the bootstrap form validation: https://getbootstrap.com/docs/5.0/forms/validation/

trekhopton (approved):

Issue #359, opened 2024-11-18
oceantv: broadcast doesn't perform off actions on disable when start fails 3 times


trekhopton:
After the broadcast directStarting state times out 3 times it is supposed to transition to directIdle and disable the broadcast. 
In broadcast_state.go line: 740, OnFailureClosure publishes the event "StartFailed" and disables the broadcast, however, startFailed event doesn't get handled, therefore the transition to directIdle doesn't happen and the off actions don't get performed. Instead startFailed gets stored in the broadcast config's events:
```
DEFAULT 2024-11-18T02:47:20.031849Z 2024/11/18 02:47:20 (name: Wagonga Inlet, id: 7NdYlI_T3ZE) publishing event: startFailedEvent
DEFAULT 2024-11-18T02:47:20.031853Z 2024/11/18 02:47:20 (name: Wagonga Inlet, id: 7NdYlI_T3ZE) storing event after cancel: startFailedEvent
DEFAULT 2024-11-18T02:47:21.048042Z 2024/11/18 02:47:21 (name: Wagonga Inlet, id: 7NdYlI_T3ZE) failed to start 3 times, disabling (disabled on first start: false, error: could not start broadcast: broadcast: Wagonga Inlet, ID: 7NdYlI_T3ZE, could not do status actions: failed to go live, could not perform status action: 1: status wait timeout exceeded for active status)
DEFAULT 2024-11-18T02:47:21.048054Z 2024/11/18 02:47:21 (name: Wagonga Inlet, id: 7NdYlI_T3ZE) broadcast context notifier is nil, setting to global notifier
DEFAULT 2024-11-18T02:47:22.526848Z 2024/11/18 02:47:22 (name: Wagonga Inlet, id: 7NdYlI_T3ZE) could not send health notification: could not send mail: could not send mail: {"Messages":[{"Errors":[{"ErrorClass":"","ErrorMessage":"You must provide at least one item in the collection.","ErrorRelatedTo":["To"],"StatusCode":400}]}]}
DEFAULT 2024-11-18T02:48:01.984500Z 2024/11/18 02:48:01 (name: Wagonga Inlet, id: 7NdYlI_T3ZE) (broadcast sm) got broadcast state machine; initial state: main.directStarting, start: 2024-11-17 22:30:00 +0000 UTC, end: 2024-11-18 06:30:00 +0000 UTC, cfg: (config logging disabled) Events: [startFailedEvent], HardwareState: hardwareOn
[...]

PR #397, opened 2024-12-02
Go to failure state instead of disabling broadcast

saxon-milton:
Resolves issue #359 

Up until this point, critical failures i.e. those that cause repeated broadcast failure, have resulted in disabling of the broadcast. Unfortunately, this meant that any stored events at the time of disabling would not get processed, such as hardware shutdown requests. This is because the state machines of disabled broadcasts do not spin up.

Furthermore, there should probably be a differentiation between a broadcast that is in a critical failure state, as opposed to one that is just inactive.

For the aforementioned reasons, this commit changes the behaviour of the broadcast state machine when it struggles to start a broadcast. If we exceed the allocated number of attempts, then a critical failure event is emitted. In response to this event, the broadcasts will move to a failure state, which is indicated in the broadcast UI via a tick box. This must then be unchecked by admin to put the broadcast into a startable state.

Direct broadcasts did not have a failure state, so this has been added.

### DIFF
--- a/cmd/oceantv/broadcast_machine.go
+++ b/cmd/oceantv/broadcast_machine.go
@@ -61,6 +61,8 @@ func (sm *broadcastStateMachine) handleEvent(event event) error {
 		sm.handleStartedEvent(event.(startedEvent))
 	case startFailedEvent:
 		sm.handleStartFailedEvent(event.(startFailedEvent))
+	case criticalFailureEvent:
+		sm.handleCriticalFailureEvent(event.(criticalFailureEvent))
 	case hardwareStartFailedEvent:
 		sm.handleHardwareStartFailedEvent(event.(hardwareStartFailedEvent))
 	case badHealthEvent:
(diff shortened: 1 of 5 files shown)


cmd/oceantv/broadcast_states.go
```diff
@@ -481,6 +481,19 @@ func (s *directIdle) enter() {
...
+type directFailure struct {
+	stateFields
+	*broadcastContext `json: "-"`
+}
+
+func newDirectFailure(ctx *broadcastContext) *directFailure {
+	return &directFailure{broadcastContext: ctx}
+}
+func (s *directFailure) enter() {
+	try(s.man.StopBroadcast(context.Background(), s.cfg, s.store, s.svc), "could not stop broadcast on direct failure entry", s.log)
```
ao-david, line 493: Does broadcastContext not implement context.Context?

cmd/oceantv/broadcast_machine.go
```diff
@@ -212,6 +214,25 @@ func (sm *broadcastStateMachine) handleStartFailedEvent(event startFailedEvent)
 	return nil
 }
 
+func (sm *broadcastStateMachine) handleCriticalFailureEvent(event criticalFailureEvent) error {
+	sm.log("handling critical failure event")
+	switch sm.currentState.(type) {
+	case *vidforwardPermanentStarting:
```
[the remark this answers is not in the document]
saxon-milton: Yes, at least at the moment. There's definitely the option to go into a failure state if we fail to fix broadcast issues through hardware restart or similar.

cmd/oceantv/broadcast_states.go
saxon-milton: nup

ao-david (approved):
LGTM

cmd/oceantv/broadcast_states.go
```diff
@@ -637,50 +675,52 @@ func broadcastCfgToState(ctx *broadcastContext) state {
 	)
 	var newState state
 	switch {
```
trekhopton, line 677: This switch is not very pretty haha

trekhopton (approved):
Looks good. Just to be clear, the way to get out of failure mode (This name cracks me up :laughing: ) is to go to the broadcast page and untick the failure mode box and save. Then it will try again three times to start the stream.

saxon-milton:
> Looks good. Just to be clear, the way to get out of failure mode (This name cracks me up 😆 ) is to go to the broadcast page and untick the failure mode box and save. Then it will try again three times to start the stream.

Yep that's it. How this looks on the UI is another question.

cmd/oceantv/broadcast_states.go
saxon-milton: Agreed. All ears if you have a better idea on how to do it hehe

Issue #162, opened 2024-06-25
gauth: Generalize to support different kinds of OAuth2 authentication


scruzin:

ao-david:
We could use an interface that implements at least the following methods:
```golang
type OAuther interface {
  AuthHandler(w http.ResponseWriter, r *http.Request) error
  CallbackHandler(w http.ResponseWriter, r *http.Request) error
}
```
These handlers would be passed the arguments from the generic endpoint, depending on the oauth implementation being used.

scruzin:
I think we can share a great deal of the implementation too, not just the interface.

I think functional options might be the way to go.

PR #398, opened 2024-12-02
Make Gauth framework Agnostic (#162)

ao-david:
This change implements a set of new interfaces (backend/) which can be implemented by either net/http types or fiber types.

This change also updates oceanbench to use these new changes.

I know that this is a massive PR, and I will probably break it up, but I thought it could be helpful to show all the pieces together.


saxon-milton (changes requested):
On a first pass it seems like an improvement! Just a few questions on the design etc. Also, is this really the "gauth framework", more just http handling framework?

ao-david:
> On a first pass it seems like an improvement! Just a few questions on the design etc. Also, is this really the "gauth framework", more just http handling framework?

Its a http framework, but it was designed with gauth in mind, and has been applied to gauth.

backend/httphandler.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		fh.ctx = c
+		return nil
+	}
+}
+
+// NetHTTPHandler creates a new HTTPHandler of the specified type, applying the passed options
+// to the newly created handler.
+func NewHTTPHandler(handlerType int, opts ...HTTPHandlerOption) (HTTPHandler, error) {
```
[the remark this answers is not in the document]
ao-david: I don't think that it is currently necessary, but maybe could useful in the future. I'll possibly leave it out then

backend/sessionstore.go
```diff
@@ -0,0 +1,157 @@
...
+	"reflect"
+
+	"github.com/gofiber/fiber/v2"
+	"github.com/gorilla/sessions"
+)
+
+// SessionStore defines an interface which manages the handling of Session management.
+type SessionStore interface {
+	// Get returns a Session based on the given id.
+	Get(string, ...SessionStoreOption) (Session, error)
```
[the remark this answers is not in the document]
ao-david: Because the option is providing the request/response interface (through either fiber.Ctx or net/http w/r), these could change with each call to Get or Save, and the values set for the New store will likely no longer be current.

What may be easier is to pass a HTTPHandler, which will have the required methods anyway? Similar to calling sessions.Get(r, ID)

backend/sessionstore.go
```diff
@@ -0,0 +1,157 @@
...
+	if s.ctx == nil {
+		return nil, fmt.Errorf("cannot get session with nil fiber context")
+	}
+
+	return NewFiberSession(id, s.ctx.Cookies(id)), nil
+}
+
+// Save implements the SessionStore interface for the FiberSessionStore type.
+//
+// NOTE: The WithFiberCtx option must be used to provide the fiber context for the
```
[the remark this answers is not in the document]
ao-david: see above

trekhopton:
It's quite a large pull request, if all all possible in the future, break these up into smaller PRS. Eg. Implementing the package, then using it with oceanbench.

cmd/oceanbench/tmp/main
trekhopton, line 1: WHat is this? Accidentally added to the commit?

go.sum
trekhopton, line 1: Do you do `go mod tidy`? Because I would expect more changes in go.sum looking at your go.mod

trekhopton (changes requested):

backend/backend_test.go
```diff
@@ -0,0 +1,188 @@
...
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/ausocean/cloud/backend"
```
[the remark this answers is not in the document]
ao-david: The main difference is whether or not the tests can call unexported methods. I also wonder whether it would make the compilation of binaries which use the backend package slightly smaller, although I don't know if it really matters.

backend/sessionstore.go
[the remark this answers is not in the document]
ao-david: Sorry, I explained that poorly, it wouldn't be changing between net/http and fiber.

Consider the following 2 handlers (in net/http)
```go
// handles requests to /first
func firstHandler(w, r) {
    h, _ := newHTTPhandler(backend.NetHTTP, WithHTTPWriterAndRequest(w,r))
    // Here h has the request to /first, and will write responses via w to that request client.
    sessionStore.Get(sessionID, withNetHTTPHandler(w,r))
    // This will get the sessions from the cookies for the request to /first
}

// handles requests to /second
func secondHandler(w, r) {
    h, _ := newHTTPhandler(backend.NetHTTP)
    // Without passing the request and response writer, this either has no context, 
    // or depending on how the handler may get reused with a global service, it may
    // have the request and response writer from the request to /first.
    sessionStore.Get(sessionID)
    // Assuming that the w and r from the first request are still attached, this will not 
    // get the sessions for the current request to /second
}
```

Hopefully this is a little more clear. Basically the fiber.Ctx or responseWriter and Request will change on every new call to a handler, and need to be passed in to calls to the session store to make sure that the correct sessions are actually being gotten. 

This may not be the best way, but consider that the gorilla/sessions implementation does the same thing, but without using functional options.
```golang
func (s *CookieStore) Get(r *http.Request, name string) (*Session, error)
```
We are just doing the same thing, but with functional options to make the interface generalisable over different http handler frameworks.

cmd/oceanbench/tmp/main
ao-david: Yes accidentally added.

go.sum
ao-david: Yes :relaxed:

ao-david:
All comments have been addressed.

SessionStore has now been removed and its functionality moved into the Handler interface, which removes the need to pass functional options.

backend/session.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// FiberSession implements the Session interface using a Fiber Cookie based
+// storage method.
+type FiberSession struct {
+	cookie *fiber.Cookie              // Cookie used to store the session.
+	values map[string]json.RawMessage // Map of the key value pairs to be encoded into the session.
+}
+
+// NewFiberSession creates a new empty FiberSession with the given id.
```
saxon-milton, line 60: what's value ?

backend/session.go
```diff
@@ -0,0 +1,173 @@
...
+// the maximum age of the cookie.
+func (s *FiberSession) SetMaxAge(age time.Duration) error {
+	s.cookie.MaxAge = int(age.Seconds())
+	return nil
+}
+
+// Set implements the Set method of the Session interface by encoding a query escaped
+// map in JSON format to the cookie value.
+func (s *FiberSession) Set(key string, value any) error {
+	// Convert the value into a json RawMessage.
```
saxon-milton, line 91: Probably don't need this.

backend/session.go
```diff
@@ -0,0 +1,173 @@
...
+
+// Invalidate implements the Invalidate method of the Session interface by setting
+// the Max Age of the cookie to -1.
+func (s *FiberSession) Invalidate() error {
+	s.cookie.MaxAge = -1
+	return nil
+}
+
+// getCookie is a helper function which returns the fiber Cookie used to store the Fiber Session.
+func (s *FiberSession) getCookie() *fiber.Cookie {
```
saxon-milton, line 120: This indirection doesn't seem that useful ?

saxon-milton (changes requested):

backend/session.go
ao-david: I have updated the comment.

backend/session.go
ao-david: Which bit sorry? just the comment or some of the logic?

backend/session.go
ao-david: I would agree, I have to assume this was carried over from the previous iteration where this made more sense. I will remove.

backend/backend_test.go
```diff
@@ -0,0 +1,170 @@
...
+// Oauth token used for testing.
+var testTok = &oauth2.Token{
+	AccessToken:  "example_access_token_12345",
+	TokenType:    "Bearer",
+	RefreshToken: "example_refresh_token_67890",
+	Expiry:       time.Now().AddDate(0, 0, 7),
+}
+
+// TestFiberHandler tests the implementation of the FiberHandler using FiberSessions.
+func TestFiberHandler(t *testing.T) {
```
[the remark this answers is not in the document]
saxon-milton: Less characters is almost always better; it's much easier to traverse the code and see what's happening at a glance (less visual input to process). The exceptions are obviously where by removing characters you're making the code harder to understand e.g. excessively abbreviated or short naming, more obscure pieces of code that deserves explanatory comments, or public/exported stuff.

A good rule of thumb is that in code commenting (not doc comments) should almost always be a "why" and not a "what"; "what" comments can be a code smell i.e. indicates poor naming/design.

backend/session.go
saxon-milton: Just the comment

ao-david:
All comments addressed.

backend/backend_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (svc *testService) set(h Handler) error {
+	sess, err := h.LoadSession(sessionID)
+	if err != nil {
+		return fmt.Errorf("error getting session: %w", err)
+	}
+
+	err = sess.Set("oauth2_token", testTok)
+	if err != nil {
+		return fmt.Errorf("unable to set seesion value: %w", err)
```
trekhopton, line 136: *session

trekhopton (approved):

saxon-milton (approved):

PR #399, opened 2024-12-03
Fix tests

ao-david:
This change fixes issues that were causing the tests to fail.
- Removes a bad import which was causing builds to fail.
- Removes tests for functions which no longer exist (newSubscriberID). This was kept due to bad merging when rebasing changes
- Adds UpdateSubscriber method which was lost due to bad merging when rebasing changes
- deals with times in UTC

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ toolchain go1.23.3
 
 require (
 	bou.ke/monkey v1.0.2
+	cloud.google.com/go/datastore v1.11.0
 	cloud.google.com/go/storage v1.30.1
 	github.com/Comcast/gots/v2 v2.2.1
 	github.com/Knetic/govaluate v3.0.0+incompatible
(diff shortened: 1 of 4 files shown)


saxon-milton:
There seems to be more going on than just removing a bad import ?

ao-david:
> There seems to be more going on than just removing a bad import ?

Yeah, It started out as removing a bad import to fix the tests, but then that revealed a test that was actually failing.
This should probably be 2 PRs?

trekhopton:
> > There seems to be more going on than just removing a bad import ?
> 
> Yeah, It started out as removing a bad import to fix the tests, but then that revealed a test that was actually failing. This should probably be 2 PRs?

It's not so big that it needs to be 2 PRs, just update the PR title and description to tell us what is being changed and why.

trekhopton (approved):

PR #402, opened 2024-12-09
bench: keep track of livestream visits with URL

trekhopton:
This was done because it should give us insights into how many people are visiting our streams through the URL redirect. The more user analytics the better.

I also remove saveLinkFunc and updateConfigWithTransaction from bench since they were unused duplicates.


model/variable.go
```diff
@@ -132,6 +132,59 @@ func PutVariable(ctx context.Context, store datastore.Store, skey int64, name, v
...
 }
 
+// PutVariableInTransaction updates or creates a variable in the datastore atomically.
+// If the variable doesn't exist, it is created with an empty value string, then the update function is applied. If it exists, the
+// provided update function is applied to modify its value.
+//
+// Argument updateFunc is a function that takes the current value of the variable and returns a new value.
+// It is used to atomically update the variable's value in the datastore. The function should return both the updated value and any error
+// encountered during the update logic.
+func PutVariableInTransaction(ctx context.Context, store datastore.Store, skey int64, name string, updateFunc func(currentValue string) string) error {
```
saxon-milton, line 142: I think you can improve the structure of this and remove the store.Get

```
func PutVariableInTransaction(ctx context.Context, store datastore.Store, skey int64, name string, updateFunc func(currentValue string) string) error {
        // If a dot exists in the name, that indicates scope, remove any colons from the scope.
	sep := strings.Index(name, ".")
	scope := ""
	if sep >= 0 {
		scope = strings.ReplaceAll(name[:sep], ":", "")
		name = scope + name[sep:]
	}

	key := store.NameKey(typeVariable, strconv.FormatInt(skey, 10)+"."+name)

update:
	err = store.Update(ctx, key, func(entity datastore.Entity) {
		if v, ok := entity.(*Variable); ok {
			v.Value = updateFunc(v.Value)
			v.Updated = time.Now()
		}
	}, &variable)
        if errors.Is(err, ErrNoSuchEntity) {
              // The variable doesn't exist, initialize it with an empty value.
		variable = Variable{
			Skey:    skey,
			Name:    name,
			Scope:   scope,
			Value:   "",
			Updated: time.Now(),
		}

		// Create the variable in the datastore.
                err := store.Create(ctx, key, &variable)
                if errors.Is(err, ErrEntityAlreadyExists) {
                      // do nothing.
                }
                if err != nil {
                       return fmt.Errorf("failed to create variable: %w", createErr)
                }
                goto update
        }
	if err != nil {
		return fmt.Errorf("failed to update variable: %w", err)
	}

	if cache := variable.GetCache(); cache != nil {
		cache.Set(key, &variable)
	}

	return nil
```

cmd/oceanbench/broadcast.go
```diff
@@ -574,12 +567,36 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
+	variableName := fmt.Sprintf("visits.%s", streamName)
+
+	// Put the incremented count. A site key of -1 indicates a global variable.
+	return model.PutVariableInTransaction(ctx, store, -1, variableName, func(currentValue string) string {
+		// Parse the current count or default to 0 if the variable doesn't exist.
+		visitCount := 0
+		if currentValue != "" {
+			var err error
+			visitCount, err = strconv.Atoi(currentValue)
+			if err != nil {
```
saxon-milton, line 593: Maybe log ?

saxon-milton (changes requested):

cmd/oceanbench/broadcast.go
```diff
@@ -574,12 +567,36 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Increment the link visit count.
+	go func() {
```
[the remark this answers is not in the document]
trekhopton: Because we can return quickly, which is preferable for a redirect handler, and update the count concurrently which may take longer.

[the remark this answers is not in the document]
saxon-milton: yeah fair enough

model/variable.go
trekhopton: Good suggestion, done.

cmd/oceanbench/utils.go
```diff
@@ -106,58 +104,6 @@ func broadcastByName(sKey int64, name string) (*BroadcastConfig, error) {
 	return cfg, nil
 }
 
-func updateConfigWithTransaction(ctx context.Context, store Store, skey int64, broadcast string, update func(cfg *BroadcastConfig) error) error {
```
ao-david, line 109: Is this supposed to be in this change?

cmd/oceanbench/broadcast.go
```diff
@@ -596,8 +596,25 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, redirectURL, http.StatusFound)
 }
 
-// writeHttpErrorAndLog is a wrapper for writeHttpError that adds logging.
```
ao-david, line 599: Is this supposed to be in this change?

ao-david (approved):
LGTM. Has it been tested? And bump the version number

model/variable.go
```diff
@@ -132,6 +133,64 @@ func PutVariable(ctx context.Context, store datastore.Store, skey int64, name, v
 	return err
 }
 
+// PutVariableInTransaction updates or creates a variable in the datastore atomically.
+// If the variable doesn't exist, it is created with an empty value string, then the update function is applied. If it exists, the
+// provided update function is applied to modify its value.
+//
+// Argument updateFunc is a function that takes the current value of the variable and returns a new value.
+// It is used to atomically update the variable's value in the datastore. The function should return both the updated value and any error
```
saxon-milton, line 141: This is wrong now ?

model/variable.go
```diff
@@ -132,6 +133,64 @@ func PutVariable(ctx context.Context, store datastore.Store, skey int64, name, v
...
+			Value:   "",
+			Updated: time.Now(),
+		}
+
+		// Create the variable in the datastore.
+		err := store.Create(ctx, key, &variable)
+		if errors.Is(err, datastore.ErrEntityExists) {
+			// do nothing.
+		}
+		if err != nil {
```
saxon-milton, line 177: does this need to be else if ? Otherwise, datastore.ErrEntityExists will trigger this and not do nothing ?

saxon-milton (changes requested):
Looks good once errors are corrected

saxon-milton:
Don't forget a minor version bump

cmd/oceanbench/utils.go
trekhopton: Yeah, if you read the PR description, I mention that I'm taking the opportunity to remove unused code. Hope you don't mind.

cmd/oceanbench/broadcast.go
trekhopton: Yeah removing unused code. See description of PR.

model/variable.go
trekhopton: Updated.

model/variable.go
trekhopton: Yeah, good catch. Changed.

PR #403, opened 2024-12-09
scripts: add build_all.sh and run_all_standalone.sh

trekhopton:
This was done to automate testing all the services in standalone. A scripts directory was made to house these scripts.


scripts/build_all.sh
```diff
@@ -0,0 +1,62 @@
...
+for dir in "${PROJECT_DIRS[@]}"; do
+    echo "Building binary for $dir..."
+    
+    # Ensure the directory exists.
+    if [ -d "$dir" ]; then
+        # Navigate to the directory.
+        cd "$dir" || exit
+        
+        # Run `go build` to build the binary.
+        if go build -o "$(basename "$dir")" .; then
```
ao-david, line 48: I know that not all our services have typescript files and rollup etc... But it would be neat to run npm run build at the same time?

scripts/run_all_standalone.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo "Stopping all services..."
+    pkill -P $$
+    echo "All services stopped."
+    exit
+}
+
+trap cleanup SIGINT SIGTERM
+
+# Run all services.
+run_service "OceanBench" "$OCEANBENCH_PATH" "./oceanbench --standalone --tvurl $TV_URL --cronurl $CRON_URL"
```
ao-david, line 71: Is there a reason we can't just call ```go run ./cmd/oceanbench --standalone --... etc``` instead of having to build first? If for some reason we do have to build first, would it make sense for this script to call build (checking for changes before building to prevent slow rebuilds)

ao-david (changes requested):

scripts/build_all.sh
trekhopton: Added to run_all_standalone.sh.

scripts/run_all_standalone.sh
trekhopton: No reason, it's just how I usually work but it's simpler with go run so I changed it.

ao-david (approved):

Issue #360, opened 2024-11-18
oceantv: broadcast voltage checking fails and disables broadcast


trekhopton:
For some reason the broadcast hardware machine can't get the latest voltage scalar, despite the configuration seeming to be correct.

```
DEFAULT 2024-11-18T03:53:17.133962Z 2024/11/18 03:53:17 (name: Wagonga Inlet, id: 3VPfcUYIkSg) publishing event: hardwareStartRequestEvent
DEFAULT 2024-11-18T03:53:17.793622Z 2024/11/18 03:53:17 (name: Wagonga Inlet, id: 3VPfcUYIkSg) (hardware sm) handling hardware start request event
DEFAULT 2024-11-18T03:53:18.455189Z 2024/11/18 03:53:18 (name: Wagonga Inlet, id: 3VPfcUYIkSg) (hardware sm) transitioning from main.hardwareOff to main.hardwareStarting
DEFAULT 2024-11-18T03:53:18.850659Z 2024/11/18 03:53:18 (name: Wagonga Inlet, id: 3VPfcUYIkSg) could not get hardware voltage: could not get current battery voltage: could not get latest scalar: rpc error: code = FailedPrecondition desc = no matching index found. recommended index is:
DEFAULT 2024-11-18T03:53:18.850687Z 2024/11/18 03:53:18 (name: Wagonga Inlet, id: 3VPfcUYIkSg) publishing event: invalidConfigurationEvent
DEFAULT 2024-11-18T03:53:18.850692Z 2024/11/18 03:53:18 (name: Wagonga Inlet, id: 3VPfcUYIkSg) got invalid configuration event, disabling broadcast: could not get hardware voltage: could not get current battery voltage: could not get latest scalar: rpc error: code = FailedPrecondition desc = no matching index found. recommended index is:
DEFAULT 2024-11-18T03:53:18.850708Z 2024/11/18 03:53:18 (name: Wagonga Inlet, id: 3VPfcUYIkSg) broadcast context notifier is nil, setting to global notifier
DEFAULT 2024-11-18T03:53:21.826437Z 2024/11/18 03:53:21 (name: Wagonga Inlet, id: 3VPfcUYIkSg) (broadcast sm) transitioning from main.directStarting to main.directIdle
DEFAULT 2024-11-18T03:53:21.826450Z 2024/11/18 03:53:21 (name: Wagonga Inlet, id: 3VPfcUYIkSg) stopping
```

[...]

PR #404, opened 2024-12-09
tv: correct alarm voltage variable name

trekhopton:
This was done because I don't think the AlarmVoltage was being retrieved properly. There was no reference to the device.

I also added a print for save hardware state because we see an error about this all the time and it would be good to have more logging.

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -468,11 +468,12 @@ func (c *revidCameraClient) voltage(ctx *broadcastContext) (float64, error) {
 func (c *revidCameraClient) alarmVoltage(ctx *broadcastContext) (float64, error) {
 	// Get AlarmVoltage variable; if the voltage is above this we expect the controller to be on.
 	// If the voltage is below this, we expect the controller to be off.
-	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, fmt.Sprintf("%012x", ctx.cfg.ControllerMAC)+".AlarmVoltage")
+	controllerMACHex := (&model.Device{Mac: ctx.cfg.ControllerMAC}).Hex()
+	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, controllerMACHex+".AlarmVoltage")
 	if err != nil {
 		return 0, fmt.Errorf("could not get alarm voltage variable: %v", err)
 	}
-	ctx.log("got AlarmVoltage for %s: %s", fmt.Sprintf("%012x", ctx.cfg.ControllerMAC), alarmVoltageVar.Value)
+	ctx.log("got AlarmVoltage for %s: %s", controllerMACHex, alarmVoltageVar.Value)
 
 	uncalibratedAlarmVoltage, err := strconv.Atoi(alarmVoltageVar.Value)
 	if err != nil {
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -468,10 +468,11 @@ func (c *revidCameraClient) voltage(ctx *broadcastContext) (float64, error) {
 func (c *revidCameraClient) alarmVoltage(ctx *broadcastContext) (float64, error) {
 	// Get AlarmVoltage variable; if the voltage is above this we expect the controller to be on.
 	// If the voltage is below this, we expect the controller to be off.
-	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, "AlarmVoltage")
+	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, fmt.Sprintf("%012x", ctx.cfg.ControllerMAC)+".AlarmVoltage")
 	if err != nil {
 		return 0, fmt.Errorf("could not get alarm voltage variable: %v", err)
 	}
+	fmt.Printf("got AlarmVoltage for %s: %s", fmt.Sprintf("%012x", ctx.cfg.ControllerMAC), alarmVoltageVar.Value)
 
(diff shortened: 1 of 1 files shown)


cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -468,10 +468,11 @@ func (c *revidCameraClient) voltage(ctx *broadcastContext) (float64, error) {
 func (c *revidCameraClient) alarmVoltage(ctx *broadcastContext) (float64, error) {
 	// Get AlarmVoltage variable; if the voltage is above this we expect the controller to be on.
 	// If the voltage is below this, we expect the controller to be off.
-	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, "AlarmVoltage")
+	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, fmt.Sprintf("%012x", ctx.cfg.ControllerMAC)+".AlarmVoltage")
 	if err != nil {
 		return 0, fmt.Errorf("could not get alarm voltage variable: %v", err)
 	}
+	fmt.Printf("got AlarmVoltage for %s: %s", fmt.Sprintf("%012x", ctx.cfg.ControllerMAC), alarmVoltageVar.Value)
```
saxon-milton, line 475: It might be better to use the logger in the broadcastContext

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -540,6 +541,7 @@ func (c *revidCameraClient) publishEventIfStatus(event event, status bool, mac i
 }
 
 func (sm *hardwareStateMachine) saveHardwareStateToConfig() error {
+	fmt.Println("saving hardware state to config", hardwareStateToString(sm.currentState))
```
saxon-milton, line 544: The hardwareStateMachine has a logger.

saxon-milton (approved):

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -468,10 +468,11 @@ func (c *revidCameraClient) voltage(ctx *broadcastContext) (float64, error) {
 func (c *revidCameraClient) alarmVoltage(ctx *broadcastContext) (float64, error) {
 	// Get AlarmVoltage variable; if the voltage is above this we expect the controller to be on.
 	// If the voltage is below this, we expect the controller to be off.
-	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, "AlarmVoltage")
+	alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, fmt.Sprintf("%012x", ctx.cfg.ControllerMAC)+".AlarmVoltage")
```
saxon-milton, line 471: Could you just ParseInt ?

saxon-milton (changes requested):

saxon-milton:
I'm slightly confused by the title of this PR. It should just be "tv: correct alarm voltage variable name" ?

trekhopton:
Just a note, while this seems related to issue #360 it was another error that I initially saw that prompted me to fix this:

`[OceanTV] 2024/12/09 19:51:08 (name: Trek Test Stream, id: 6JAoes27RUQ) (hardware sm) transitioning from main.hardwareOff to main.hardwareStarting
[OceanTV] 2024/12/09 19:51:08 (name: Trek Test Stream, id: 6JAoes27RUQ) could not get alarm voltage: could not get alarm voltage variable: datastore: no such entity
[OceanTV] 2024/12/09 19:51:08 (name: Trek Test Stream, id: 6JAoes27RUQ) publishing event: invalidConfigurationEvent
[OceanTV] 2024/12/09 19:51:08 (name: Trek Test Stream, id: 6JAoes27RUQ) got invalid configuration event, disabling broadcast: could not get alarm voltage: could not get alarm voltage variable: datastore: no such entity`

trekhopton:
> I'm slightly confused by the title of this PR. It should just be "tv: correct alarm voltage variable name" ?

Yeah it could be that. I meant get the Alarm Voltage for the **device** as opposed to the site, which it was currently doing, which would be a global site AlarmVoltage that doesn't exist.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Yeah sorry mi have dumb moment. What about ```strconv.Itoa```

trekhopton: I could use strconv.FormatInt(value, 16) apparently for hex, but to pad with 0s what I have is the simplest way, which is what model.Device.Hex() uses but I can't use that here since I don't have access to the device.

saxon-milton: hhhmm, I almost think we should get the device some how. I just think that's a better pattern that doing this.

trekhopton: Why? Is there a reason not to trust the configured controller MAC address?

saxon-milton: No, I just mean that it's good to follow the same pattern of using the device and Hex method; it seems this is how we get variables everywhere else.

trekhopton: What about this?:

```
controllerMACHex := (&model.Device{Mac: ctx.cfg.ControllerMAC}).Hex()
alarmVoltageVar, err := model.GetVariable(context.Background(), ctx.store, ctx.cfg.SKey, controllerMACHex+".AlarmVoltage")
```

I'm trying to avoid a network request for the device when all we need is the MAC.

Btw SPrintf for the Mac is also used in OceanManager.SetupSecondary so it's not the only place that it's done this way. And it's not really that complex, it's just another representation of the same value.

saxon-milton (approved):

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: The reason I create `controllerMACHex` is because it's used twice.

Issue #383, opened 2024-11-26
AOTV: Get product from url


ao-david:
The checkout page should be able to determine what the product is from the url query parameters.

PR #413, opened 2024-12-30
AOTV: Add priceID to checkout links from plans (#383)

ao-david:
This change adds a priceID query parameter to the links from the plans page to the checkout. This will allow for the checkout to determine which product is being purchased.

These priceIDs come from stripe, and will need to be updated if the products ever get changed. We could potentially think about making these be loaded via an API call to make sure they are always up to date, but I think this works for now.

### DIFF
--- a/cmd/ausoceantv/plans.html
+++ b/cmd/ausoceantv/plans.html
@@ -9,9 +9,9 @@
   </head>
   <body class="h-screen">
     <div class="container flex h-full items-center justify-center gap-5">
-      <plan-element plan-type="Day Pass" plan-cost="4" class="h-fit w-1/3" href="/checkout?product=day-pass"></plan-element>
-      <plan-element plan-type="Monthly Subscription" plan-cost="5" class="h-fit w-1/3" href="/checkout?product=month-subscription"></plan-element>
-      <plan-element plan-type="Yearly Subscription" plan-cost="30" comp-cost="60" class="h-fit w-1/3" href="/checkout?product=year-subscription"></plan-element>
+      <plan-element plan-type="Day Pass" plan-cost="4" class="h-fit w-1/3" href="/checkout.html?priceID=price_1QZMqMDMhXzQAv2TK0EFunPO"></plan-element>
+      <plan-element plan-type="Monthly Subscription" plan-cost="5" class="h-fit w-1/3" href="/checkout.html?priceID=price_1QZMp5DMhXzQAv2Thg1erujS"></plan-element>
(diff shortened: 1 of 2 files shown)


trekhopton (approved):

cmd/ausoceantv/plans.html
```diff
@@ -9,9 +9,9 @@
   </head>
   <body class="h-screen">
     <div class="container flex h-full items-center justify-center gap-5">
-      <plan-element plan-type="Day Pass" plan-cost="4" class="h-fit w-1/3" href="/checkout?product=day-pass"></plan-element>
-      <plan-element plan-type="Monthly Subscription" plan-cost="5" class="h-fit w-1/3" href="/checkout?product=month-subscription"></plan-element>
-      <plan-element plan-type="Yearly Subscription" plan-cost="30" comp-cost="60" class="h-fit w-1/3" href="/checkout?product=year-subscription"></plan-element>
+      <plan-element plan-type="Day Pass" plan-cost="4" class="h-fit w-1/3" href="/checkout.html?priceID=price_1QZMqMDMhXzQAv2TK0EFunPO"></plan-element>
```
saxon-milton, line 12: Maybe dumb question but could we have less obscure names for the priceIDs that then map them appropriately to the strip IDs ? Then you can have these IDs change (not sure why they would) and not have it break the frontend ?

saxon-milton: I think this would be following the “Thin Client, Thick Server” (or “Server-Driven UI”) principle more, if I'm actually understanding this.

saxon-milton (changes requested):
Should we do a patch version bump ?

cmd/ausoceantv/plans.html
ao-david: I don't know that changing that would make the system any better, I think it would actually make the system more complex with no benefit. Either way, when the priceIDs change, they have to be updated somewhere, leaving them here is the least amount of code, and I would argue possibly easiest to update.

I think at some stage this setup will be changed to dynamically controlled by a request to a stripe API to get the up to date products, and their pricing.

BTW, the priceIDs shouldn't change unless we change the product (or accidentally delete, which I have already done once...)

saxon-milton: I'm not sure I really agree; having stripe price IDs hardcoded into the HTML does not sound like a good design decision IMO, and less code is certainly not always better.

This doesn't have to change now, but I think some of the other ideas on how this should work are better.

ao-david: Yeah ok, I think there may be bigger fish to fry at the moment, but will make an issue to improve later. See #419

saxon-milton (approved):

PR #415, opened 2024-12-30
AOTV: Add cart info to checkout page.

### LINKED ISSUE #383: AOTV: Get product from url

ao-david:
This change adds the current cart to the checkout page such that the user can see what they are purchasing.
image

requires #414 
closes #383

### DIFF
--- a/cmd/ausoceantv/src/checkout.js
+++ b/cmd/ausoceantv/src/checkout.js
@@ -11,10 +11,27 @@ document
 
 // Fetches a payment intent and captures the client secret
 async function initialize() {
-  const response = await fetch("/api/v1/stripe/create-payment-intent", {
-    method: "POST",
-    headers: { "Content-Type": "application/json" },
-  });
+  // Extract the 'priceID' query parameter from the URL
+  const urlParams = new URLSearchParams(window.location.search);
+  const id = urlParams.get("priceID");
+
+  if (!id) {
+    console.log("setting error message");
+    let msg = document.getElementById("msg");
+    msg.innerHTML = "<p>Choose a Plan <a href='/plans.html'>here</a></p>";
+    msg.removeAttribute("hidden");
+
+    document.getElementById("payment-form").setAttribute("hidden", true);
+    return;
+  }
+
+  const response = await fetch(
+    "/api/v1/stripe/create-payment-intent?priceID=" + id,
+    {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+    },
+  );
   const { clientSecret, dpmCheckerLink } = await response.json();
 
   const appearance = {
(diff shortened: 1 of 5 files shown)


trekhopton (approved):

cmd/ausoceantv/src/checkout.js
```diff
@@ -11,10 +11,27 @@ document
 
 // Fetches a payment intent and captures the client secret
 async function initialize() {
-  const response = await fetch("/api/v1/stripe/create-payment-intent", {
-    method: "POST",
-    headers: { "Content-Type": "application/json" },
-  });
+  // Extract the 'priceID' query parameter from the URL
```
saxon-milton, line 14: dumb question, but can we do all this on the backend ?

cmd/ausoceantv/src/checkout.js
```diff
@@ -29,6 +46,36 @@ async function initialize() {
...
+  });
+  console.debug(product);
+
+  // Show the product information in the cart.
+  let items = document.getElementById("items");
+  let name = document.getElementById("name");
+  let price = document.getElementById("price");
+  let desc = document.getElementById("desc");
+
+  // Update the values.
```
saxon-milton, line 71: DN

cmd/ausoceantv/stripe.go
```diff
@@ -274,3 +275,38 @@ func getPrice(pid string) (*stripe.Price, error) {
 	params := &stripe.PriceParams{}
 	return price.Get(pid, params)
 }
+
+func (svc *service) handleGetPrice(c *fiber.Ctx) error {
+	pid := c.Params("id")
+	if pid == "" {
+		return fmt.Errorf("must specify price_id")
```
saxon-milton, line 282: ```errors.New```

saxon-milton: Also, I think this should be "price_id empty"

cmd/ausoceantv/stripe.go
```diff
@@ -274,3 +275,38 @@ func getPrice(pid string) (*stripe.Price, error) {
...
+
+	log.Infof("price: %+v", price)
+
+	return c.JSON(price)
+}
+
+func (svc *service) handleGetProduct(c *fiber.Ctx) error {
+	pid := c.Params("id")
+	if pid == "" {
+		return fmt.Errorf("must specify product_id")
```
saxon-milton, line 299: ```errors.New```

saxon-milton (changes requested):

saxon-milton (commented):
Minor or patch bump ?

cmd/ausoceantv/src/checkout.js
ao-david: Which bit?

saxon-milton: Looking at this more, I guess don't understand why we wouldn't just get the priceID on the backend and make the required requests there (just trying to minimise javascript) ? I'm assuming the reason is because this is faster i.e. page loads and then stuff is populated on the page ?

ao-david: I'm not sure if you are aware, but we aren't using server side rendering on AusOceanTV so any dynamic content has to be loaded via js from the frontend (at least in my understanding). @trekhopton or @scott97 may be able to chime in here.

saxon-milton: hhmm maybe we'll take this offline.

PR #416, opened 2024-12-30
Model: Use namekey for subscriber

ao-david:
This changes the key of the subscriber from a 10 digit ID to a concatenation of the ID and the email.

This was done so that the entity could be queried by email by the filestore.

### DIFF
--- a/model/subscriber.go
+++ b/model/subscriber.go
@@ -35,7 +35,10 @@ const (
 	typeSubscriber = "Subscriber" // Subscriber datastore type.
 )
 
-var errDuplicateSubscriberEmails = errors.New("more than one subscriber exists for a given email")
+var (
+	errDuplicateSubscriberEmails = errors.New("more than one subscriber exists for a given email")
+	errDuplicateSubscriberIDs    = errors.New("more than one subscriber exists for a given id")
+)
 
 // Subscriber is an entity in the datastore representing a user who subscribes to AusOcean.TV.
 type Subscriber struct {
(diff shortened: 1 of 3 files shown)


model/subscriber.go
```diff
@@ -80,14 +81,14 @@ func (s *Subscriber) GetCache() datastore.Cache {
 func CreateSubscriber(ctx context.Context, store datastore.Store, s *Subscriber) error {
 	// If the subscriber has an ID, use that.
 	if s.ID != 0 {
-		key := store.IDKey(typeSubscriber, s.ID)
+		key := store.NameKey(typeSubscriber, fmt.Sprintf("%d.%s", s.ID, s.Email))
```
trekhopton, line 84: It doesn't really make sense to me to have a randomly generated int plus an email for the key. Should we just use the email? That's going to be unique to the subscriber right? @scruzin might have some ideas.

trekhopton (commented):
This was done to work with a limitation of the filestore, ie. that it can only query entities that have the query parameters in the key. Another option would be to improve the file store's query abilities.

If the datastore is going to use the email as part of the key, I don't think there's a need to keep the ID alone in the struct as an int64. We should either get rid of the ID, and just use the email, or we should change the ID field to a string and make sure it has both parts, the int and the email. Unless you can think of a reason to query only by ID (leaving out the email)?

model/subscriber.go
ao-david: @scruzin mentioned previously that he liked the idea of using the ID for any user facing links, for example a public profile page might look like ```ausocean.tv/user/1029384756```
From a technical perspective however, I don't think there is any requirement to leave it in.

model/subscriber.go
```diff
@@ -36,6 +36,7 @@ const (
 )
 
 var errDuplicateSubscriberEmails = errors.New("more than one subscriber exists for a given email")
+var errDuplicateSubscriberIDs = errors.New("more than one subscriber exists for a given id")
```
saxon-milton, line 39: Create a var block.

saxon-milton (approved):

saxon-milton (changes requested):
Patch ver bump ?

trekhopton (approved):

PR #417, opened 2024-12-31
AOTV: Update to new Gauth implementation

ao-david:
This change updates AusOceanTV to run on the new implementation of Gauth which uses the backend package.


cmd/ausoceantv/main.go
```diff
@@ -105,9 +106,19 @@ func main() {
...
 	app := fiber.New(fiber.Config{ErrorHandler: api.ErrorHandler})
 
-	// Perform one-time setup or bail.
+	// Encrypt cookies.
+	// NOTE: This must be done before any middleware which uses cookies.
 	ctx := context.Background()
-	svc.setup(ctx, app)
+	key, err := gauth.GetSecret(ctx, projectID, "sessionKey")
+	if err != nil {
+		log.Fatalf("unable to get sessionKey secret: %v", err)
```
saxon-milton, line 114: This seems more like a programmer error (?) in which case we should use panic (?). We can also recover panics (plus do fancy things like email notify), and it show a stacktrace. ```log.Fatalf``` on the other hand uses ```os.Exit(1)``` which is pretty harsh.

saxon-milton (changes requested):

saxon-milton (commented):
Patch ver bump ?

cmd/ausoceantv/main.go
ao-david: I'm not sure if there is a downside to using panic, I was using fatal because this step is really part of the setup, and the setup function says that ```Any errors are considered fatal```

ao-david:
> Patch ver bump ?

Done

PR #418, opened 2024-12-31
AOTV: Add endpoints for subscriptions

ao-david:
This change adds 2 new endpoints, ```/get/subscription``` and ```/stripe/cancel```. These will allow the front-end to get data about the subscription and cancel the subscription respectively.

### DIFF
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -26,11 +26,13 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/log"
@@ -51,7 +53,7 @@ const (
 	projectID     = "ausoceantv"
 	oauthClientID = "1005382600755-7st09cc91eqcqveviinitqo091dtcmf0.apps.googleusercontent.com"
 	oauthMaxAge   = 60 * 60 * 24 * 7 // 7 days.
-	version       = "v0.1.6"
+	version       = "v0.2.0"
 )
 
 // service defines the properties of our web service.
@@ -83,6 +85,9 @@ func registerAPIRoutes(app *fiber.App) {
 		Options("/create-payment-intent", svc.preFlightOK).
 		Post("/create-payment-intent", svc.handleCreatePaymentIntent).
 		Post("/cancel", svc.cancelSubscription)
+
+	v1.Group("/get").
+		Get("/subscription", svc.getSubscriptionHandler)
 }
 
 func main() {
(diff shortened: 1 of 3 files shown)


trekhopton (approved):

cmd/ausoceantv/main.go
```diff
@@ -51,7 +52,7 @@ const (
 	projectID     = "ausoceantv"
 	oauthClientID = "1005382600755-7st09cc91eqcqveviinitqo091dtcmf0.apps.googleusercontent.com"
 	oauthMaxAge   = 60 * 60 * 24 * 7 // 7 days.
-	version       = "v0.1.5"
+	version       = "v0.1.7"
```
saxon-milton, line 55: This change is more like a minor bump.

cmd/ausoceantv/main.go
```diff
@@ -192,3 +197,30 @@ func (svc *service) setup(ctx context.Context) {
 	svc.auth = &gauth.UserAuth{ProjectID: projectID, ClientID: oauthClientID, MaxAge: oauthMaxAge}
 	svc.auth.Init(backend.NewFiberHandler(nil))
 }
+
+func (svc *service) getSubscriptionHandler(c *fiber.Ctx) error {
+	ctx := context.Background()
+	p, err := svc.auth.GetProfile(backend.NewFiberHandler(c))
+	if err != nil {
```
saxon-milton, line 204: Do all ```err != nil``` correspond to an ```api.Unauthorized``` error ?

cmd/ausoceantv/main.go
```diff
@@ -192,3 +197,30 @@ func (svc *service) setup(ctx context.Context) {
...
+		return api.Unauthorized(err)
+	}
+
+	subscriber, err := model.GetSubscriberByEmail(ctx, svc.settingsStore, p.Email)
+	if err != nil {
+		return fmt.Errorf("error getting subscriber by email for: %s: %w", p.Email, err)
+	}
+
+	subscription, err := model.GetSubscription(ctx, svc.settingsStore, subscriber.ID, 0)
```
saxon-milton, line 213: 0 is magic number ?

cmd/ausoceantv/main.go
```diff
@@ -192,3 +197,30 @@ func (svc *service) setup(ctx context.Context) {
...
+	if err != nil {
+		return fmt.Errorf("error getting subscription for id: %d: %w", subscriber.ID, err)
+	}
+
+	// Check that the subscription is still valid.
```
saxon-milton, line 218: Maybe elaborate on this one.

cmd/ausoceantv/main.go
```diff
@@ -192,3 +197,30 @@ func (svc *service) setup(ctx context.Context) {
...
+	if subscription.Finish.Before(time.Now()) {
+		c.JSON(nil)
+		return nil
+	}
+
+	c.JSON(subscription)
```
saxon-milton, line 224: Should this be ```return c.JSON(subscription)``` ?

cmd/ausoceantv/stripe.go
```diff
@@ -129,3 +130,27 @@ func (svc *service) handleCreatePaymentIntent(c *fiber.Ctx) error {
 
 	return c.JSON(v)
 }
+
+func (svc *service) cancelSubscription(c *fiber.Ctx) error {
+	ctx := context.Background()
+	p, err := svc.auth.GetProfile(backend.NewFiberHandler(c))
+	if err != nil {
```
saxon-milton, line 137: ditto

cmd/ausoceantv/stripe.go
```diff
@@ -129,3 +130,27 @@ func (svc *service) handleCreatePaymentIntent(c *fiber.Ctx) error {
...
+	}
+
+	sub, err := model.GetSubscription(ctx, svc.settingsStore, subscriber.ID, 0)
+	if err != nil {
+		return fmt.Errorf("error getting subscription for id: %d: %w", subscriber.ID, err)
+	}
+
+	sub.Renew = false
+
+	// TODO: Update the Stripe subscription to cancel at period end.
```
saxon-milton, line 153: Should this be louder ? Should we check for period end and call ```panic("unimplemented")``` ?

saxon-milton (changes requested):

cmd/ausoceantv/stripe.go
ao-david: I agree that this should probably be louder, although I'm not sure my comment was very clear.

What I am saying here is that what I have done so far doesn't actually cancel the subscription on stripe and so the user will still get charged. Currently it only cancels it in our records. I am going to work on this immediately, but will also make it panic here?

saxon-milton: It all depends on how quickly we can get to it. If it was something very important that we weren't going to work on immediately then a ```panic("unimplemented")``` is a good idea, but if you're going to work on it pretty soon then probably DN.

PR #427, opened 2025-01-04
oceantv: split stopping state into shutdown and power off

saxon-milton:
Closes issue #31 

We would like to start doing a proper shutdown to power down the cameras as opposed to just cutting the power. To do this we have decided to have two substates in the stopping state i.e. shutdown and power off.

Another change that has resulted is that the restart state actually has substates i.e. stopping and starting.

These changes have resulted in a cascade of changes especially to testing.

### DIFF
--- a/cmd/oceanbench/t/broadcast.html
+++ b/cmd/oceanbench/t/broadcast.html
@@ -177,6 +177,10 @@
               <input type="input" name="on-actions" class="form-control w-50" value="{{.CurrentBroadcast.OnActions}}" class="actions" />
             </div>
             <div class="d-flex align-items-center gap-1 mb-1">
+              <label for="shutdown-actions" class="w-25 text-end">Shutdown Actions:</label>
+              <input type="input" name="shutdown-actions" class="form-control w-50" value="{{.CurrentBroadcast.ShutdownActions}}" class="actions">
+            </div>
+            <div class="d-flex align-items-center gap-1 mb-1">
               <label for="off-actions" class="w-25 text-end">Off Actions:</label>
               <input type="input" name="off-actions" class="form-control w-50" value="{{.CurrentBroadcast.OffActions}}" class="actions" />
             </div>
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -30,6 +30,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
@@ -104,6 +105,7 @@ type BroadcastConfig struct {
 	CameraMac                int64         // Camera hardware's MAC address.
 	ControllerMAC            int64         // Controller hardware's MAC adress (controller used to power camera).
 	OnActions                string        // A series of actions to be used for power up of camera hardware.
+	ShutdownActions          string        // A series of actions to be used for shutdown of camera hardware.
 	OffActions               string        // A series of actions to be used for power down of camera hardware.
 	RTMPVar                  string        // The variable name that holds the RTMP URL and key.
 	Active                   bool          // This is true if the broadcast is currently active i.e. waiting for data or currently streaming.
@@ -307,6 +309,21 @@ func extStart(ctx context.Context, cfg *BroadcastConfig, log func(string, ...int
 	err := setActionVars(ctx, cfg.SKey, onActions, settingsStore, log)
 	if err != nil {
 		return fmt.Errorf("could not set device variables required to start stream: %w", err)
+	}
+
+	return nil
+}
+
+var errNoShutdownActions = errors.New("no shutdown actions provided")
+
+func extShutdown(ctx context.Context, cfg *BroadcastConfig, log func(string, ...interface{})) error {
+	if cfg.ShutdownActions == "" {
+		return errNoShutdownActions
+	}
+
+	err := setActionVars(ctx, cfg.SKey, cfg.ShutdownActions, settingsStore, log)
+	if err != nil {
+		return fmt.Errorf("could not set device variables to end stream: %w", err)
 	}
 
 	return nil
--- a/cmd/oceantv/broadcast_events.go
+++ b/cmd/oceantv/broadcast_events.go
@@ -106,17 +106,19 @@ var _ = registerEvent(hardwareResetRequestEvent{})
 
 func (e hardwareResetRequestEvent) String() string { return "hardwareResetRequestEvent" }
 
-type hardwareStartFailedEvent struct{}
+type hardwareStartFailedEvent struct{ string }
 
 var _ = registerEvent(hardwareStartFailedEvent{})
 
 func (e hardwareStartFailedEvent) String() string { return "hardwareStartFailedEvent" }
+func (e hardwareStartFailedEvent) Error() string  { return e.string }
 
-type hardwareStopFailedEvent struct{}
+type hardwareStopFailedEvent struct{ string }
 
 var _ = registerEvent(hardwareStopFailedEvent{})
 
 func (e hardwareStopFailedEvent) String() string { return "hardwareStopFailedEvent" }
+func (e hardwareStopFailedEvent) Error() string  { return e.string }
 
 type hardwareStartedEvent struct{}
 
@@ -130,11 +132,12 @@ var _ = registerEvent(hardwareStoppedEvent{})
 
 func (e hardwareStoppedEvent) String() string { return "hardwareStoppedEvent" }
 
-type controllerFailureEvent struct{}
+type controllerFailureEvent struct{ string }
 
 var _ = registerEvent(controllerFailureEvent{})
 
 func (e controllerFailureEvent) String() string { return "controllerFailureEvent" }
+func (e controllerFailureEvent) Error() string  { return e.string }
 
 type slateResetRequested struct{}
 
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -35,7 +35,8 @@ func newableWithContext(new func(ctx *broadcastContext) any, args ...interface{}
 }
 
 type hardwareRestarting struct {
-	*broadcastContext `json:"-"`
+	stateWithTimeoutFields
+	substate state
 }
 
 var _ = register(hardwareRestarting{})
@@ -49,17 +50,85 @@ func (s hardwareRestarting) New(args ...interface{}) (any, error) {
 }
 
 func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
-	return &hardwareRestarting{ctx}
+	return &hardwareRestarting{newStateWithTimeoutFields(ctx), nil}
 }
 
 func (s *hardwareRestarting) enter() {
-	s.camera.stop(s.broadcastContext)
+	s.LastEntered = time.Now()
+	s.substate = newHardwareStopping(s.broadcastContext)
+	s.substate.enter()
 }
 func (s *hardwareRestarting) exit() {}
+func (s *hardwareRestarting) transition() {
+	switch s.substate.(type) {
+	case *hardwareStopping:
+		s.log("(hardwareRestarting) transitioning from substate hardwareStopping to hardwareStarting")
+		s.substate.exit()
+		s.substate = newHardwareStarting(s.broadcastContext)
+		s.substate.enter()
+	default:
+		panic("hardwareRestarting: unexpected transition")
+	}
+}
+
+func (s *hardwareRestarting) handleTimeEvent(t timeEvent) {
+	switch s.substate.(type) {
+	case *hardwareStopping:
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwareStopFailedEvent{"hardware stop timed out"})
+			return
+		}
+
+		s.substate.(*hardwareStopping).handleTimeEvent(t)
+	case *hardwareStarting:
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwareStartFailedEvent{"exceeded starting timeout during hardware restart"})
+			return
+		}
+
+		// If the camera is reporting then the start has completed.
+		if s.cameraIsReporting() {
+			s.bus.publish(hardwareStartedEvent{})
+			return
+		}
+	default:
+		// Do nothing.
+	}
+}
+
+func (s *hardwareRestarting) handleHardwareStoppedEvent(event hardwareStoppedEvent) {
+	s.log("handling hardware stopped event")
+	switch s.substate.(type) {
+	case *hardwareStopping:
+		s.transition()
+	default:
+		// For any other state ignore.
+	}
+}
+
+func (s *hardwareRestarting) handleHardwareShutdownFailedEvent(event hardwareShutdownFailedEvent) {
+	switch s.substate.(type) {
+	case *hardwareStopping:
+		s.substate.(*hardwareStopping).handleHardwareShutdownFailedEvent(event)
+	default:
+		// Ignore.
+	}
+}
+
+func (s *hardwareRestarting) cameraIsReporting() bool {
+	up, err := s.camera.isUp(s.broadcastContext, model.MacDecode(s.cfg.CameraMac))
+	if err != nil {
+		s.bus.publish(invalidConfigurationEvent{fmt.Sprintf("could not get camera reporting status: %v", err)})
+		return false
+	}
+	return up
+}
 
 type hardwareStarting struct {
+	stateWithTimeoutFields
 	*broadcastContext `json:"-"`
-	LastEntered       time.Time
 }
 
 var _ = register(hardwareStarting{})
@@ -100,7 +169,7 @@ func (s *hardwareStarting) enter() {
 		return
 	}
 
-	controllerIsOn, err := s.camera.isUp(s.broadcastContext)
+	controllerIsOn, err := s.camera.isUp(s.broadcastContext, model.MacDecode(s.cfg.ControllerMAC))
 	if err != nil {
 		msg := fmt.Sprintf("could not get controller status: %v", err)
 		s.log(msg)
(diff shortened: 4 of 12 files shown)


cmd/oceantv/broadcast.go
```diff
@@ -307,6 +309,21 @@ func extStart(ctx context.Context, cfg *BroadcastConfig, log func(string, ...int
...
 }
 
+var errNoShutdownActions = errors.New("no shutdown actions provided")
+
+func extShutdown(ctx context.Context, cfg *BroadcastConfig, log func(string, ...interface{})) error {
+	if cfg.ShutdownActions == "" {
+		return errNoShutdownActions
+	}
+
+	err := setActionVars(ctx, cfg.SKey, cfg.OffActions, settingsStore, log)
```
ao-david, line 319: should this be ```OffActions``` or ```ShutdownActions``` here?

cmd/oceantv/broadcast_events.go
```diff
@@ -106,17 +106,19 @@ var _ = registerEvent(hardwareResetRequestEvent{})
 
 func (e hardwareResetRequestEvent) String() string { return "hardwareResetRequestEvent" }
 
-type hardwareStartFailedEvent struct{}
+type hardwareStartFailedEvent struct{ string }
```
ao-david, line 109: I'm a little confused how this works? Shouldn't the field be named?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -49,17 +50,85 @@ func (s hardwareRestarting) New(args ...interface{}) (any, error) {
...
+func (s *hardwareRestarting) handleTimeEvent(t timeEvent) {
+	switch s.substate.(type) {
+	case *hardwareStopping:
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwareStopFailedEvent{"hardware stop timed out"})
+			return
+		}
+
+		s.substate.(*hardwareStopping).handleTimeEvent(t)
```
ao-david, line 83: Is this call too hardwareStopping.handleTimeEvent also going to log a timeout failure? If so, should we just call that directly?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -49,17 +50,85 @@ func (s hardwareRestarting) New(args ...interface{}) (any, error) {
...
+	case *hardwareStarting:
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwareStartFailedEvent{"failed to start hardware during restart"})
```
ao-david, line 87: Should there be a mention of a timeout in this log?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -49,17 +50,85 @@ func (s hardwareRestarting) New(args ...interface{}) (any, error) {
...
+			return
+		}
+
+		// If the camera is reporting then the start has completed.
+		if s.cameraIsReporting() {
+			s.bus.publish(hardwareStartedEvent{})
```
ao-david, line 93: I think this should be checked first to give the camera the best chance of reporting before timing out. Otherwise there is a chance that the camera has started reporting after the state has timed out but before we have cancelled the operation.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -49,17 +50,85 @@ func (s hardwareRestarting) New(args ...interface{}) (any, error) {
...
 }
 func (s *hardwareRestarting) exit() {}
+func (s *hardwareRestarting) transition() {
+	switch s.substate.(type) {
+	case *hardwareStopping:
+		s.log("(hardwareRestarting) transitioning from substate hardwareStopping to hardwareStarting")
+		s.substate.exit()
+		s.substate = newHardwareStarting(s.broadcastContext)
+		s.substate.enter()
+	default:
```
ao-david, line 69: Why is a transition from hardwareStarting considered a panic?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -192,8 +261,75 @@ func sanatisedVoltageRecoveryTimeout(ctx *broadcastContext) int {
 	return ctx.cfg.VoltageRecoveryTimeout
 }
 
+type hardwareShutdownFailedEvent struct{ string }
+
+var _ = registerEvent(hardwareShutdownFailedEvent{})
+
+func (e hardwareShutdownFailedEvent) String() string { return "hardwareShutdownFailedEvent" }
```
ao-david, line 268: Should this be Name() or String()

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -192,8 +261,75 @@ func sanatisedVoltageRecoveryTimeout(ctx *broadcastContext) int {
...
+func (e hardwareShutdownFailedEvent) Error() string  { return e.string }
+
+type hardwareShutdownEvent struct{}
+
+var _ = registerEvent(hardwareShutdownEvent{})
+
+func (e hardwareShutdownEvent) String() string { return "hardwareShutdownEvent" }
```
ao-david, line 275: Ditto

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -206,12 +342,79 @@ func (s hardwareStopping) New(args ...interface{}) (any, error) {
...
+func (s *hardwareStopping) handleTimeEvent(t timeEvent) {
+	switch s.substate.(type) {
+	case *hardwareShuttingDown:
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwareShutdownFailedEvent{"hardware shutdown timed out"})
+			return
+		}
+
+		if !s.cameraIsReporting() {
```
ao-david, line 377: Again, I think this should be checked before the timeout

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -206,12 +342,79 @@ func (s hardwareStopping) New(args ...interface{}) (any, error) {
...
+		}
+
+	case *hardwarePoweringOff:
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwarePowerOffFailedEvent{"hardware power off timed out"})
+			return
+		}
+
+		if !s.cameraIsReporting() {
```
ao-david, line 390: check before the timeout

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -206,12 +342,79 @@ func (s hardwareStopping) New(args ...interface{}) (any, error) {
...
+	case *hardwareShuttingDown:
+		s.logAndNotify(broadcastHardware, "shutdown failed during hardware stop, skipping to power off: %v", event.Error())
+		s.transition()
+	default:
+		// Ignore.
+	}
+}
+
+func (s *hardwareStopping) cameraIsReporting() bool {
+	up, err := s.camera.isUp(s.broadcastContext, model.MacDecode(s.cfg.CameraMac))
```
ao-david, line 410: This seems to be fairly common code, would it be possible to handle the error logging inside the s.camera.isUp function?

cmd/oceanbench/t/broadcast.html
```diff
@@ -75,7 +75,27 @@ <h2 class="pt-5">Stream Settings</h2>
...
+            <div class="d-flex align-items-center gap-1 mb-1">
+              <div class="w-25"></div>
+              <div class="form-check form-switch d-flex align-items-center gap-1 p-0">
+                <!-- Hidden input for the 'off' state -->
+                <!-- This is what the value of the hardware-state will be when the checkbox is not checked.-->
+                <input type="hidden" name="hardware-state" value="hardwareOff">
+            
+                <!-- Checkbox for the 'failure' state -->
+                <!-- This is what the value of the hardware-state will be when the checkbox is checked.-->
+                <input
```
ao-david, line 90: This is not normally how we format HTML. I think that we generally keep all args on one line where possible (even if it makes the line too long)

ao-david (changes requested):
Would be good to have a screenshot of the UI change.

cmd/oceanbench/t/broadcast.html
saxon-milton: Do we have a html formatter that we use ?

cmd/oceantv/broadcast.go
saxon-milton: Ha yeah should be ShutdownActions

cmd/oceantv/broadcast_events.go
saxon-milton: No it's just embedded

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: because this code shouldn't transition to ```hardwareStarting``` unless there's a programmer error.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: You mean on the ```hardwareStopping``` timeout ? These are two different timeouts and could have there own timeout values. Also, I feel like that might constitute an abstraction leak ?

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Yeah that's a fair call

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: I don't really think there's a right answer here. Equally I could argue that your suggestion would be a bad idea for strict timeouts. It's well possible we go over the timeout (but haven't checked yet) then almost a minute later the camera is reporting (haven't checked yet), then when we check the camera first it seems like it's working before the timeout when actually it started reporting well after.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Events are currently still using ```String```

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: ditto

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: not sure

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: What error logging ?

trekhopton:
I think it would help a lot to see a state machine diagram for this hardware state machine. I have a somewhat completed one here:

https://miro.com/welcomeonboard/WXpIK3I2RDQrK3IvUktFeUN0YnRwMDM0NE1BTmlBMytuRy9LbWRSVFZIVXpYNWF2c3hvWXdQNWJ4V3hPc2FqdFlZY0JRNVk4eWpSdkZ6a1B4amlwL09QSUJGQlJ0YXU2UnZ2aGV1L05vSEdEbWEyTWhTTVh2MVBGS2VJWk9NZkkhZQ==?share_link_id=518036862511

Do you think you could make one up? Or would you like to share the one that I've already started? 
It would help me see how this works. My concern is that the substate idea seems to be complicating things. I like the idea of a hardware restarter which is a seperate state machine, rather than the substates (which as you said _is_ technically another state machine.)

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -206,12 +342,79 @@ func (s hardwareStopping) New(args ...interface{}) (any, error) {
...
+		s.log("(hardwareStopping) transitioning from substate hardwareShuttingDown to hardwarePoweringOff")
+		s.substate.exit()
+		s.substate = newHardwarePoweringOff(s.broadcastContext)
+		s.substate.enter()
+	default:
+		panic("hardwareStopping: unexpected transition")
+	}
+}
+
+func (s *hardwareStopping) handleTimeEvent(t timeEvent) {
```
trekhopton, line 368: This seems really un-intuitive, having states have their own handlers for events, essentially being listeners. I fear that it's just going to add confusion to people trying to follow this code and the states. Unless you can show me a state diagram where this makes more sense than having another separate state machine eg. a restarter, I think it's too complicated.

trekhopton (changes requested):

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: I disagree, it's a common pattern to have states of a state machine have event handlers/listeners.

hardwareStopping has sub-states and _is_ therefore a state machine, so it's not that different to having a "restarter". I think the abstraction of having hardwareStopping have sub-states shuttingDown and poweringOff is a nicer abstraction than having a "restarter". Also, that doesn't really make sense with the stopping state right ? Maybe, with restarting but not stopping.

trekhopton: Okay, I see. I'm just trying to prevent unnecessary complication but I can see the merit.

saxon-milton:
> I think it would help a lot to see a state machine diagram for this hardware state machine. I have a somewhat completed one here:
> 
> https://miro.com/welcomeonboard/WXpIK3I2RDQrK3IvUktFeUN0YnRwMDM0NE1BTmlBMytuRy9LbWRSVFZIVXpYNWF2c3hvWXdQNWJ4V3hPc2FqdFlZY0JRNVk4eWpSdkZ6a1B4amlwL09QSUJGQlJ0YXU2UnZ2aGV1L05vSEdEbWEyTWhTTVh2MVBGS2VJWk9NZkkhZQ==?share_link_id=518036862511
> 
> Do you think you could make one up? Or would you like to share the one that I've already started? It would help me see how this works. My concern is that the substate idea seems to be complicating things. I like the idea of a hardware restarter which is a seperate state machine, rather than the substates (which as you said _is_ technically another state machine.)

I don't understand how a restarter is not just a sub-state of hardwareStopping and hardwareRestarting though ? Having the shutdown process as a separate state machine seems weird to me, even un-intuitive because there's no good reason why it can't be the responsibility of the stopping and restarting states to deal with that. 

Remember, the hardware state machine is only 2 layers deep, that's hardly stretching the limits. I'd argue that it's easier to understand what's happening when you enter the stopping or restarting states, because the states and transitions are clearly visible right there. You don't have to go to another place in the code to figure out how that works.

I'm not sure how much value drawing up the state machine really is. If it helps the hardwareStopping and hardwareRestarting state machines are linear.

trekhopton:
> > I think it would help a lot to see a state machine diagram for this hardware state machine. I have a somewhat completed one here:
> > https://miro.com/welcomeonboard/WXpIK3I2RDQrK3IvUktFeUN0YnRwMDM0NE1BTmlBMytuRy9LbWRSVFZIVXpYNWF2c3hvWXdQNWJ4V3hPc2FqdFlZY0JRNVk4eWpSdkZ6a1B4amlwL09QSUJGQlJ0YXU2UnZ2aGV1L05vSEdEbWEyTWhTTVh2MVBGS2VJWk9NZkkhZQ==?share_link_id=518036862511
> > Do you think you could make one up? Or would you like to share the one that I've already started? It would help me see how this works. My concern is that the substate idea seems to be complicating things. I like the idea of a hardware restarter which is a seperate state machine, rather than the substates (which as you said _is_ technically another state machine.)
> 
> I don't understand how a restarter is not just a sub-state of hardwareStopping and hardwareRestarting though ? Having the shutdown process as a separate state machine seems weird to me, even un-intuitive because there's no good reason why it can't be the responsibility of the stopping and restarting states to deal with that.
> 
> Remember, the hardware state machine is 3 layers deep, whilst I probably wouldn't go any deeper than that, this isn't too bad. I'd argue that it's easier to understand what's happening when you enter the stopping or restarting states, because the states and transitions are clearly visible right there. You don't have to go to another place in the code to figure out how that works.
> 
> I'm not sure how much value drawing up the state machine really is. If it helps the hardwareStopping and hardwareRestarting state machines are linear.

Yeah, I've changed my mind, I can see why you've done it this way. 
State diagrams just really help me visualise and compare approaches, but I get it this time. No need right now.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -192,8 +261,75 @@ func sanatisedVoltageRecoveryTimeout(ctx *broadcastContext) int {
 	return ctx.cfg.VoltageRecoveryTimeout
 }
 
+type hardwareShutdownFailedEvent struct{ string }
```
trekhopton, line 264: Why have hardwareShutdownFailedEvent and hardwareShutdownEvent here when all the other hardware events are defined in broadcast_events.go?

saxon-milton: Yeah, I was thinking that maybe I should define events (if published on from a specific place) where they are most relevant, but not 100% on this yet.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: The camera reporting is more important than the strict timeout since that's what we're waiting for, but yeah we can always adjust the timeouts, no biggy.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -49,17 +50,85 @@ func (s hardwareRestarting) New(args ...interface{}) (any, error) {
...
+		withTimeout := s.substate.(stateWithTimeout)
+		if withTimeout.timedOut(t.Time) {
+			s.bus.publish(hardwareStartFailedEvent{"failed to start hardware during restart"})
+			return
+		}
+
+		// If the camera is reporting then the start has completed.
+		if s.cameraIsReporting() {
+			s.bus.publish(hardwareStartedEvent{})
+			return
```
trekhopton, line 94: Should we do something like set the substate to nil, since it's started, it's no longer hardwareStarting. I could imagine it being confusing if it's left in that state. I know ultimately we will transition out of this higherlevel hardwareRestarting state, but sometimes thinks get stopped or stuck etc and I think it's best to update states as soon as we can to make them true as much as possible.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Your reply could be more helpful Mr. Tharon. David knows what a panic is meant for. Transitioning from hardwareStarting shouldn't happen because at that point the higher level hardwareRestarting state is finished and will be transitioned out of. There's no more substates we need to transition to. (Unless we chose to transition to hardwareStarted, this wouldn't do anything but could be more readable and current.)

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Or perhaps transition to hardwareStarted (even though it won't do anything, it's a more true state.) This sort of thing comes in handy if it's ever shown in the UI for example.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Hm, yeah okay. Not too fussed.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: As opposed to states.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: We can always adjust timeouts.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: We can always adjust timeouts.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -575,8 +793,8 @@ func (c *revidCameraClient) alarmVoltage(ctx *broadcastContext) (float64, error)
 	return alarmVoltage, nil
 }
 
-func (c *revidCameraClient) isUp(ctx *broadcastContext) (bool, error) {
-	controllerIsOn, err := model.DeviceIsUp(context.Background(), ctx.store, model.MacDecode(ctx.cfg.ControllerMAC))
+func (c *revidCameraClient) isUp(ctx *broadcastContext, mac string) (bool, error) {
+	controllerIsOn, err := model.DeviceIsUp(context.Background(), ctx.store, mac)
```
trekhopton, line 797: controllerIsOn -> deviceIsOn

and the error message below, controller -> device

Since this is used for cameras too now. ln 122 in cameraIsReporting()

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Logging twice isn't so bad. As Saxon said, they're different timeouts.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: It'd be good if the states reflected the true states immediately but I don't think we can practically do this without making the software horrible. The limitation is that we only tick the state machine every minute (corresponding to the broadcast check cron frequency; a side effect of trying to do a state machine on a stateless service i.e. GAE) therefore almost everything is inherently delayed, and anything shown in the UI will probably be out of date to some degree. If we didn't have this limitation everything would happen immediately (ticking frequency could be much higher).

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: This is probably exactly the reason why i used ```controllerIsOn``` to make it clear that we're checking the controller status; ```deviceIsOn``` is ambiguous.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Why can't we transition multiple times within one tick? I guess I wasn't sure if the ticking was just to handle events.

cmd/oceanbench/t/broadcast.html
ao-david: We don't have a formatter set up for oceanbench at the moment, we are using prettier for ausoceantv.
The main guidlines we follow are that we use 2 space indenting, and keep all tags on a single line.

It can be left for now, but we should setup a formatter at some point.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Actually, I've rethought and don't think it should.

ao-david (approved):
Still needs a few small comments addressed that you have acknowledged, but otherwise LGTM.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: But you're checking the camera MAC in the code mentioned above. Is that a mistake?

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Actually that's not quite right, let me clarify, the transitions actually should happen virtually immediately (within one tick) unless obviously you trigger the hardware to do something and need to wait for the next tick, or trigger YT to transition (in the YT case, because we use a go routine, the context is cancelled and events are stored until the next tick). 

In the particular instance we're looking at here, the hardware state would transition immediately from the publishing of that event (literally before the return), so I'm not sure how much value there is in doing what you proposed.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Yeah, sorry, you're right, I'll change this.

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Okay yeah I see.

trekhopton:
Looking good once the existing comments are addressed. Keen to get it merged.

trekhopton:
We've decided to use prettier for bench html to try to keep things consistent. See PR #433. Feel free to merge and then I'll merge your changes and format it in my PR.

trekhopton (approved):

PR #430, opened 2025-01-13
bench: increase resolution of graph while keeping grouping

trekhopton:
This was done so that there is a better balance between performance and resultion of data. These values seem to scal the graph nicely at different zoom levels.

I also removed the progress bar once loading is done since it would be duplicated if the user loaded another graph.

image


saxon-milton:
I feel like the duplicate progress bars is a regression, but i guess this fixes it so whatevs.

cmd/oceanbench/s/graph.js
```diff
@@ -165,6 +168,21 @@ function graph() {
     var dateAxis = chart.xAxes.push(new am4charts.DateAxis());
     dateAxis.renderer.minGridDistance = 60;
     dateAxis.groupData = true;
+    dateAxis.groupCount = 500;
+    dateAxis.groupIntervals.setAll([
```
saxon-milton, line 172: Maybe comment this, it's kind of weird.

saxon-milton (approved):

cmd/oceanbench/s/graph.js
```diff
@@ -165,6 +168,21 @@ function graph() {
     var dateAxis = chart.xAxes.push(new am4charts.DateAxis());
     dateAxis.renderer.minGridDistance = 60;
     dateAxis.groupData = true;
```
saxon-milton, line 170: I think we should be able to turn data grouping and the scaling below on/off.

saxon-milton (changes requested):

cmd/oceanbench/s/graph.js
trekhopton: Done.

cmd/oceanbench/s/graph.js
trekhopton: Done.

cmd/oceanbench/s/graph.js
```diff
@@ -182,6 +220,16 @@ function graph() {
...
   });
+  console.log("graphing done");
+}
+
+function toggleGrouping() {
+  if (chart){
+    chart.invalidateRawData();
+    graph();
+  } else {
+    console.log("no chart to toggle grouping yet");
```
ao-david, line 231: this could maybe be a console.warn or console.error

cmd/oceanbench/s/graph.js
```diff
@@ -157,14 +168,41 @@ function graph() {
...
+      { timeUnit: "millisecond", count: 100 },
+      { timeUnit: "second", count: 1 },
+      { timeUnit: "second", count: 10 },
+      { timeUnit: "minute", count: 1 },
+      { timeUnit: "minute", count: 2 },
+      { timeUnit: "minute", count: 5 },
+      { timeUnit: "minute", count: 10 },
+      { timeUnit: "minute", count: 20 },
+      { timeUnit: "hour", count: 1 },
+      { timeUnit: "hour", count: 3 }
```
ao-david, line 204: Are these just the options of how the data will group based on the zoom?

ao-david (approved):
Can you attach a screenshot of the UI, and also bump the Version number

Otherwise LGTM

saxon-milton (approved):

cmd/oceanbench/s/graph.js
trekhopton: Yes.

PR #431, opened 2025-01-18
bench: add device throughput graphing

saxon-milton:
This is useful to see throughput characteristics over time.

We're doing this by adding a quasi-pin i.e. it's not a real sensor pin, but we can set the pin to "throughput" on the UI for the device, and then have special handling of this case to use the throughputs endpoint which gets historical data to provide for graphing. The process is very similar to the data endpoint handling however, the datastore ops are different to get the data.

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -238,6 +238,7 @@ func main() {
 	http.HandleFunc("/admin/broadcast", adminHandler)
 	http.HandleFunc("/admin/utils", adminHandler)
 	http.HandleFunc("/data/", dataHandler)
+	http.HandleFunc("/throughputs", throughputsHandler)
 	http.HandleFunc("/", indexHandler)
 
 	if standalone {
--- a/cmd/oceanbench/monitor.go
+++ b/cmd/oceanbench/monitor.go
@@ -149,30 +149,66 @@ func monitorHandler(w http.ResponseWriter, r *http.Request) {
 	writeTemplate(w, r, "monitor.html", &data, "")
 }
 
+// scalarCount returns the number of scalars received for the first pin of a device
+// for the period of time defined by start and end. This can be used to determine
+// the throughput of a device.
+func scalarCount(ctx context.Context, device *model.Device, start, end int64) (count int, err error) {
+	for _, pin := range strings.Split(device.Inputs, ",") {
+		if count != 0 || (pin[0] != 'A' && pin[0] != 'D' && pin[0] != 'X') {
+			continue
+		}
+		sid := model.ToSID(model.MacDecode(device.Mac), pin)
+		keys, err := model.GetScalarKeys(ctx, mediaStore, sid, []int64{start, end})
+		if err != nil {
+			return 0, fmt.Errorf("could not get scalar keys: %v", err)
+		}
+		count = len(keys)
+		break
+	}
+	return count, nil
+}
+
 // throughput determines the number of scalars received within the count
 // period for a device, as well as the maximum number of scalars expected.
-// The first A, D or X pin that provides count data is used.
-func throughput(ctx context.Context, device model.Device) (count, maxCount int, err error) {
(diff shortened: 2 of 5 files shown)


cmd/oceanbench/main.go
```diff
@@ -117,7 +117,7 @@ var (
 	setupMutex    sync.Mutex
 	templates     *template.Template
 	setTemplates  *template.Template
-	dataHost      = "https://bench.cloudblue.org"
+	dataHost      = "https://dev-dot-oceanbench.ts.r.appspot.com"
```
ao-david, line 120: This shouldn't be changed in the PR

cmd/oceanbench/monitor.go
```diff
@@ -149,32 +149,60 @@ func monitorHandler(w http.ResponseWriter, r *http.Request) {
 	writeTemplate(w, r, "monitor.html", &data, "")
 }
 
+func scalarCount(ctx context.Context, device model.Device, start, end int64) (count int, err error) {
```
ao-david, line 152: I assume that this is just factored out so that it can be used elsewhere and hasn't been changed? I can't see any changes?

ao-david: I know that it isn't required by golang idiom to comment a unexported function, but it can still be quite useful so that the input and output expectations are known to a developer that is trying to debug code, or use the function. In this case, throughput says that the first pin is used to calculate count, but this doesn't get described for this function.

cmd/oceanbench/monitor.go
```diff
@@ -149,32 +149,60 @@ func monitorHandler(w http.ResponseWriter, r *http.Request) {
...
+	for _, pin := range strings.Split(device.Inputs, ",") {
+		if count != 0 || (pin[0] != 'A' && pin[0] != 'D' && pin[0] != 'X') {
+			continue
```
ao-david, line 155: I'm curious if there is ever a time where count is != 0 that we wouldn't want to return an error, or at very least break?
from my perspective, it either seems that a programmer error has occurred, or that we are going to loop over every pin unnecessarily?

cmd/oceanbench/monitor.go
```diff
@@ -149,32 +149,60 @@ func monitorHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	start := time.Now().Add(-countPeriod).Unix()
+	count, err = scalarCount(ctx, device, start, -1)
+	if err != nil {
+		return 0, 0, fmt.Errorf("could not get scalar count: %v", err)
 	}
 	return count, maxCount, nil
 }
 
+func throughputsFor(ctx context.Context, device model.Device, from, to time.Time) ([]float64, error) {
```
ao-david, line 187: again, a comment for this would be nice. Especially to explain what the format of the return is, since it differs from the function above

cmd/oceanbench/monitor.go
```diff
@@ -149,32 +149,60 @@ func monitorHandler(w http.ResponseWriter, r *http.Request) {
...
+	if device.Inputs == "" || device.MonitorPeriod == 0 {
+		return nil, nil
```
ao-david, line 189: Again, I understand that this is copied, but it feels like this could return an error, letting the user know that the device won't be returning any information?

cmd/oceanbench/monitor.go
ao-david: Not sure how go handles it under the hood, but is it worth changing this to pass by reference to avoid having to copy the data to make it slightly faster?

ao-david (approved):
Just a few questions

trekhopton (approved):
Need to fix the dataHost.

cmd/oceanbench/monitor.go
saxon-milton: Yeah I kinda agree, but not super interested in changing how this works in this PR.

ao-david: Yeah seems reasonable, I was kinda just curious honestly

PR #432, opened 2025-01-19
bench: display broadcast states on UI

saxon-milton:
Closes issue #178


ao-david (approved):

trekhopton:
Can you please include a screenshot of the UI for UI changes?

cmd/oceanbench/t/broadcast.html
```diff
@@ -125,6 +125,18 @@ <h2 class="pt-5">Stream Settings</h2>
...
+                <label class="w-25 text-end">Broadcast State:</label>
+                <input class="w-50 align-items-center form-control" id="broadcast-state" size="1" value="{{.CurrentBroadcast.BroadcastState}}" readonly>
+              </div>
+              <div class="d-flex align-items-center gap-1 mb-1">
+                <label class="w-25 text-end">Hardware State:</label>
+                <input class="w-50 align-items-center form-control" id="hardware-state" size="1" value="{{.CurrentBroadcast.HardwareState}}" readonly>
+              </div>
+              <div class="d-flex align-items-center gap-1 mb-1">
+                <label class="w-25 text-end">Hardware State Data:</label>
+                <input class="w-50 align-items-center form-control" id="hardware-state-data" size="1" value="{{.CurrentBroadcast.PrettyHardwareStateData}}" readonly>
```
trekhopton, line 138: What's the difference between displaying the hardware state and the hardware state data?

trekhopton (approved):

trekhopton:
Nice work.
Approved because this is a good start, but I think the real issue behind the inclusion of this issue in the AusOceanTV MVP is that the states are often wrong or don't make sense. This needs to be cleaned up and made reliable before we can have proper feedback from the UI. 
This UI element is a good first step in making the misalignment clearly visible. I'll be making issues as I see them.

trekhopton (changes requested):
Actually, I see no reason to have hardwareState data. By my testing it's most often empty and I'm not if it contains anything useful?

saxon-milton:
> Nice work. Approved because this is a good start, but I think the real issue behind the inclusion of this issue in the AusOceanTV MVP is that the states are often wrong or don't make sense. This needs to be cleaned up and made reliable before we can have proper feedback from the UI. This UI element is a good first step in making the misalignment clearly visible. I'll be making issues as I see them.

It'd be nice to see some a particular examples/evidence of "the states are often wrong or don't make sense" and what exactly "needs to be cleaned up and made reliable before we can have proper feedback from the UI.". 

You've often stated in general that the software is unreliable etc, but you haven't often given precise examples of what exactly you're referring to, which is frustrating.

I think this should give you "proper" feedback from the UI in regards to what state the broadcast and hardware are in, unless you can find an exception where it doesn't and define it.

cmd/oceanbench/t/broadcast.html
saxon-milton: The hardware state data may contain useful information such as any substates.

saxon-milton:
> Can you please include a screenshot of the UI for UI changes?

Can do.

trekhopton:
Not if it's an easy fix but the UI fields are blank upon saving a broadcast, only after some time do they appear with values upon refresh.
It would be nice to have a keepalive connection or a realtime connection like websocket to have auto-updating state.

trekhopton:
> Actually, I see no reason to have hardwareState data. By my testing it's most often empty and I'm not certain if it contains anything useful?

I take it back, hardwareStateData is useful since it contains LastEntered, timeout etc.

trekhopton (approved):

trekhopton:
In fact it would actually be very helpful to have the broadcastState data too. So we can see the timeout and lastEntered for hardwareStarting state.

PR #433, opened 2025-01-21
bench: use prettier to format our html

trekhopton:
For a long time our html format hasn't been consistent or checked. 
Running prettier with npm run format should help us keep our html formatting consistent. 
This change is the initial output from prettier.

I tested that bench compiles and the templates parse and the pages all look right.

### DIFF
--- a/cmd/oceanbench/.prettierrc
+++ b/cmd/oceanbench/.prettierrc
@@ -1,0 +1,7 @@
+{
+    "printWidth": 2000,
+    "tabWidth": 2,
+    "useTabs": false,
+    "singleQuote": false,
+    "htmlWhitespaceSensitivity": "ignore"
+}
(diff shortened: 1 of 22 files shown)


cmd/oceanbench/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+    "printWidth": 2000,
```
ao-david, line 2: While we're at it, if you are going to update this config, do you want to make the ausoceanTV one match exactly, just for simplicity?

ao-david (approved):

cmd/oceanbench/.prettierrc
trekhopton: I'll do that in another PR, this one is big enough.

PR #437, opened 2025-01-27
oceantv: add and use composite store

saxon-milton:
Closes issue #360 

It's awkward to need to provide a different datastore based on the entity kind, and is also error prone. We're trying to improve this situation by creating a "facade" datastore, CompositeStore, which selects the appropriate datastore instance based on the kind.

### DIFF
--- a/cmd/oceantv/main.go
+++ b/cmd/oceantv/main.go
@@ -46,22 +46,21 @@ import (
 
 const (
 	projectID          = "oceantv"
-	version            = "v0.5.0"
+	version            = "v0.6.0"
 	projectURL         = "https://oceantv.appspot.com"
 	cronServiceAccount = "oceancron@appspot.gserviceaccount.com"
 	locationID         = "Australia/Adelaide" // TODO: Use site location.
 )
 
 var (
-	setupMutex    sync.Mutex
-	settingsStore datastore.Store
-	mediaStore    datastore.Store
-	debug         bool
-	standalone    bool
-	notifier      notify.Notifier
-	ofsvc         openfish.OpenfishService
-	cronSecret    []byte
-	storePath     string
+	setupMutex sync.Mutex
+	store      *CompositeStore
+	debug      bool
+	standalone bool
+	notifier   notify.Notifier
+	ofsvc      openfish.OpenfishService
+	cronSecret []byte
+	storePath  string
 )
 
 func main() {
@@ -150,7 +149,7 @@ func errNoGlobalNotifierHandler(secrets map[string]string) utils.RecoveryHandler
 			notifier, err = notify.NewMailjetNotifier(
 				notify.WithSecrets(secrets),
 				notify.WithRecipientLookup(tvRecipients),
-				notify.WithStore(notify.NewStore(settingsStore)),
+				notify.WithStore(notify.NewStore(store)),
 			)
 			if err != nil {
 				log.Printf("could not remediate missing global notifier: %v", err)
@@ -180,11 +179,14 @@ func setup(ctx context.Context) {
 	setupMutex.Lock()
 	defer setupMutex.Unlock()
 
-	if settingsStore != nil {
+	if store != nil {
 		return
 	}
 
-	var err error
+	var (
+		err                       error
+		settingsStore, mediaStore datastore.Store
+	)
 	if standalone {
 		log.Printf("Running in standalone mode")
 		settingsStore, err = datastore.NewStore(ctx, "file", "vidgrind", storePath)
@@ -203,6 +205,8 @@ func setup(ctx context.Context) {
 	}
 	model.RegisterEntities()
 
+	store = ausOceanCompositeStore(settingsStore, mediaStore)
+
 	cronSecret, err = gauth.GetHexSecret(ctx, projectID, "cronSecret")
 	if err != nil || cronSecret == nil {
 		log.Printf("could not get cronSecret: %v", err)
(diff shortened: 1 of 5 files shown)


cmd/oceantv/main.go
```diff
@@ -64,6 +65,29 @@ var (
...
 )
 
+// ausoceanCompositeStoreFactory returns a composite store that delegates to the
+// appropriate store based on the kind of the entity. This tries to fix the
+// awkwardness of selecting the right store based on the kind of the entity
+// you're dealing with.
+func ausoceanCompositeStoreFactory() *CompositeStore {
+	return NewCompositeStore(
+		map[string]datastore.Store{
+			"Scalar":     mediaStore,
```
ao-david, line 75: Do we not have constants for these kinds?

cmd/oceantv/main.go
```diff
@@ -56,6 +56,7 @@ var (
 	setupMutex    sync.Mutex
 	settingsStore datastore.Store
```
ao-david, line 57: Should the settingsStore and mediaStore be moved somewhere else or at least commented to say to not use them?

cmd/oceantv/store.go
```diff
@@ -0,0 +1,157 @@
...
+// on the kind.
+func (s *CompositeStore) Get(ctx context.Context, key *Key, dst datastore.Entity) error {
+	return s.stores[key.Kind].Get(ctx, key, dst)
+}
+
+// GetAll implements the Store.GetAll by calling GetAll on the appropriate store.
+// We find the appropriate store through trial and error given that the query
+// does not contain the kind. We look at possible stores and try to find the matching
+// one.
+func (s *CompositeStore) GetAll(ctx context.Context, query datastore.Query, dst interface{}) ([]*Key, error) {
```
ao-david, line 108: How often do we use the GetAll call like this? This concerns me slightly if we use it regularly as we could end up making almost double the amount of db calls. If its an uncommon call its probably fine. Although if this is what we will commonly be using, than perhaps we should try and move functions away from using this call.

ao-david (commented):
Is the fact that we have a settings store and a media store just because we haven't properly gotten around to migrating them into one place. I'm not sure that we need to do this if we can just migrate the data.

saxon-milton:
> Is the fact that we have a settings store and a media store just because we haven't properly gotten around to migrating them into one place. I'm not sure that we need to do this if we can just migrate the data.

I think migrating is more complicated than it sounds. This is a simple interim solution.

cmd/oceantv/main.go
saxon-milton: Yeah that's a fair idea, maybe I'll "un-global" them. I don't think this is necessary anymore.

cmd/oceantv/main.go
saxon-milton: Unfortunately they are unexported. Once that changes (if that changes) we can use them.

cmd/oceantv/store.go
saxon-milton: Yeah this seemed like the only (and poor) solution, but I think I've got an idea that would require reflect; still not amazing, but perhaps better than excessive calls to the DB.

cmd/oceantv/store.go
```diff
@@ -0,0 +1,157 @@
...
+	// Compose list of possible stores (don't repeat stores).
+	storeMap := make(map[datastore.Store]struct{})
+	for _, store := range s.stores {
+		storeMap[store] = struct{}{}
+	}
+
+	// Try each store GetAll call until we find a nil error.
```
trekhopton, line 115: Could you instead use reflect on dst to get the expected type/kind? Something like this:

```
dstType := reflect.TypeOf(dst)
if dstType.Kind() != reflect.Ptr || dstType.Elem().Kind() != reflect.Slice {
	return nil, fmt.Errorf("dst must be a pointer to a slice, got %T", dst)
}
elemType := dstType.Elem().Elem() // The type of the elements in the slice.
```

That might be faster than doing datastore requests.

cmd/oceantv/main.go
ao-david: I think its worth exporting them now as part of this change if they exist somewhere already, I can't see a reason why they should be 'private'

cmd/oceantv/store.go
trekhopton: Oh yeah, just saw your reply to David.

cmd/oceantv/main.go
saxon-milton: It's a big change though.

ao-david (approved):

cmd/oceantv/store.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// ausOceanCompositeStore returns a composite store that delegates to the
+// appropriate store based on the kind of the entity. This tries to fix the
+// awkwardness of selecting the right store based on the kind of the entity
+// you're dealing with.
+func ausOceanCompositeStore(settingsStore, mediaStore datastore.Store) *CompositeStore {
+	getKindFromQuery := func(query datastore.Query) string {
+		getKindField := func(v reflect.Value) string {
+			const kindField = "kind"
```
trekhopton, line 19: Weird, I didn't think you could reflect a field that wasn't exported, hence why I suggested reflecting the dst type instead. 
This solution does break our datastore abstraction layer, but probably better than doing requests to get the kind.

trekhopton (approved):
Make sure you test the new getKindFromQuery functionality, otherwise, looks good.

saxon-milton:
> Make sure you test the new getKindFromQuery functionality, otherwise, looks good.

Yeah, I did some manual testing in go playground.

PR #438, opened 2025-01-28
Bench: Use sensor names when available

ao-david:
This change uses the sensor names for a pin when they exist on the search page.


trekhopton (approved):
Remove the todo on ln: 160.

PR #439, opened 2025-01-29
Bench: Add unique sandbox endpoint (#309)

### LINKED ISSUE #309: Bench: give sandbox configuration page unique endpoint
Currently the sandbox devices page has a separate page to display configuration of a device. It is proposed to move this to a separate endpoint.

ao-david:
This change adds a new endpoint /sandbox for the sandbox site which users are automatically redirected to.

This also fixes a few small bugs with the config process, such as loosing the device key upon configuration.

closes #309

### DIFF
--- a/cmd/oceanbench/sandbox.go
+++ b/cmd/oceanbench/sandbox.go
@@ -35,6 +35,60 @@ import (
 	"github.com/ausocean/cloud/system"
 )
 
+type sandboxData struct {
+	Devices []model.Device
+	Mac     string
+	Device  model.Device
+	commonData
+}
+
+func sandboxHandler(w http.ResponseWriter, r *http.Request) {
+	profile, err := getProfile(w, r)
+	if err != nil {
+		if err != gauth.TokenNotFound {
+			log.Printf("authentication error: %v", err)
+		}
+		http.Redirect(w, r, "/", http.StatusUnauthorized)
+		return
+	}
+	skey, _ := profileData(profile)
+
+	data := sandboxData{
+		commonData: commonData{
+			Pages:   pages("home"),
+			Profile: profile,
+		},
+	}
+
+	if skey != model.SandboxSkey {
+		writeTemplate(w, r, "sandbox.html", &data, "Must be on Sandbox Site.")
+		return
+	}
+
+	ctx := context.Background()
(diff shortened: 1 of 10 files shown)


cmd/oceanbench/sandbox.go
```diff
@@ -0,0 +1,224 @@
...
+		},
+	}
+
+	if skey != model.SandboxSkey {
+		writeTemplate(w, r, "sandbox.html", &data, "Must be on Sandbox Site.")
+		return
+	}
+
+	ctx := context.Background()
+	data.Devices, err = model.GetDevicesBySite(ctx, settingsStore, skey)
```
saxon-milton, line 47: no error handling ?

cmd/oceanbench/sandbox.go
```diff
@@ -0,0 +1,224 @@
...
+
+	ma := r.FormValue("ma")
+	if !model.IsMacAddress(ma) && ma != "" {
```
saxon-milton, line 50: shouldn't this be an or (`||`) condition ?

cmd/oceanbench/sandbox.go
```diff
@@ -0,0 +1,224 @@
...
+//
+// Form Fields:
+//
+//	dn = Name of the new device
+//	ma = MAC address
+//	dt = device type
+//	wi = comma seperated WiFi name and password (optional)
+//	ll = comma seperated latitude and longitude (optional)
+//	sk = target site key for the new device
+func configDevicesHandler(w http.ResponseWriter, r *http.Request) {
```
saxon-milton, line 76: so this is just a copy from the other file ? May have been better as two changes.

saxon-milton (approved):

saxon-milton:
maybe a minor version bump ?

cmd/oceanbench/sandbox.go
ao-david: Yeah, might do that actually.

cmd/oceanbench/sandbox.go
ao-david: No, this is so the page will still render with no device selected yet. This lets it continue through if the mac is blank, otherwise it renders an error message.

PR #444, opened 2025-01-30
bench: add live redirect options for mute, autoplay, related 

trekhopton:
This was done to support a more customised experience on AusOceanTV with embedded live streams.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -603,38 +603,30 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
 
 	redirectURL := v.Value
 
-	// Append query parameters to the URL.
 	u, err := url.Parse(redirectURL)
 	if err != nil {
 		http.Error(w, "Invalid livestream URL", http.StatusInternalServerError)
 		return
 	}
-
-	// Ensure query parameters for embedding and additional YouTube options.
 	query := u.Query()
 
-	// Always add rel=0
-	query.Set("rel", "0")
-
-	// Provide embed link if requested.
+	// Append embed, mute and autoplay if present in the request query.
 	if _, ok := r.URL.Query()["embed"]; ok {
 		u.Path = strings.Replace(u.Path, "watch?v=", "embed/", 1)
+		// Always add rel=0 for embedded videos.
+		query.Set("rel", "0")
 	}
-
-	// Append mute and autoplay if present in the request query.
 	if _, ok := r.URL.Query()["mute"]; ok {
 		query.Set("mute", "1")
 	}
-
 	if _, ok := r.URL.Query()["autoplay"]; ok {
 		query.Set("autoplay", "1")
 	}
 
 	u.RawQuery = query.Encode()
-
 	redirectURL = u.String()
 
-	log.Printf("Redirecting to livestream link: %s", redirectURL)
+	log.Printf("redirecting to livestream link: %s", redirectURL)
 	http.Redirect(w, r, redirectURL, http.StatusFound)
 }
 
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -603,31 +603,54 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
 
 	redirectURL := v.Value
 
-	u, err := url.Parse(redirectURL)
+	// Transform the YouTube URL based on options in query parameters.
+	redirectURL, err = transformYouTubeURL(redirectURL, r)
 	if err != nil {
-		http.Error(w, "Invalid livestream URL", http.StatusInternalServerError)
+		log.Printf("error transforming YouTube URL: %v", err)
+		http.Error(w, "invalid livestream URL", http.StatusInternalServerError)
 		return
 	}
-	query := u.Query()
-
-	// Append embed, mute and autoplay if present in the request query.
-	if _, ok := r.URL.Query()["embed"]; ok {
-		u.Path = strings.Replace(u.Path, "watch?v=", "embed/", 1)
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/broadcast.go
```diff
@@ -602,15 +603,56 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
 	redirectURL := v.Value
 
-	// Provide embed link if requested.
-	if _, ok := r.URL.Query()["embed"]; ok {
-		redirectURL = strings.ReplaceAll(redirectURL, "watch?v=", "embed/")
+	// Transform the YouTube URL based on options in query parameters.
+	redirectURL, err = transformYouTubeURL(redirectURL, r)
+	if err != nil {
+		log.Printf("error transforming YouTube URL: %v", err)
+		http.Error(w, "invalid livestream URL", http.StatusInternalServerError)
```
ao-david, line 610: you could use writeHttpError here,

cmd/oceanbench/broadcast.go
```diff
@@ -602,15 +603,56 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	log.Printf("redirecting to livestream link, link: %s", redirectURL)
+	log.Printf("redirecting to livestream link: %s", redirectURL)
 	http.Redirect(w, r, redirectURL, http.StatusFound)
 }
 
+func transformYouTubeURL(rawURL string, r *http.Request) (string, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return "", err
```
ao-david, line 621: Could wrap this with what it failed doing.

cmd/oceanbench/broadcast.go
```diff
@@ -602,15 +603,56 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	query := u.Query()
+	videoID := query.Get("v")
+	if videoID == "" {
+		return "", fmt.Errorf("invalid YouTube watch URL, missing video ID")
```
ao-david, line 627: errors.New()

cmd/oceanbench/broadcast.go
```diff
@@ -602,15 +603,56 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "invalid livestream URL", http.StatusInternalServerError)
+		return
 	}
 
-	log.Printf("redirecting to livestream link, link: %s", redirectURL)
+	log.Printf("redirecting to livestream link: %s", redirectURL)
 	http.Redirect(w, r, redirectURL, http.StatusFound)
 }
 
+func transformYouTubeURL(rawURL string, r *http.Request) (string, error) {
```
ao-david, line 618: After testing with this slightly modified code, it seems that the URL doesn't come back the same if no values are passed? ie not embed, or mute or autoplay.
```
package main

import (
	"fmt"
	"log"
	"net/url"
)

func main() {
	isEmbed := false
	isMute := false
	isAutoPlay := false

	rawUrl := "https://www.youtube.com/watch?v=_F6Nf26a1_w"

	u, err := url.Parse(rawUrl)
	if err != nil {
	}

	query := u.Query()
	videoID := query.Get("v")
	if videoID == "" {
		log.Fatal(err)
	}

	// Update path if embed is requested.
	embed := false
	if isEmbed {
		embed = true
		u.Path = fmt.Sprintf("/embed/%s", videoID)
		u.RawQuery = "" // Reset query parameters.
	}
	newQuery := url.Values{}

	// Always set rel=0 for embedded videos.
	if embed {
		newQuery.Set("rel", "0")
	}

	// Conditionally set mute and autoplay if requested.
	if isMute {
		newQuery.Set("mute", "1")
	}
	if isAutoPlay {
		newQuery.Set("autoplay", "1")
	}

	u.RawQuery = newQuery.Encode()
	log.Println("final string:", u.String())
}
```

cmd/oceanbench/broadcast.go
```diff
@@ -602,15 +603,56 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
+		return "", fmt.Errorf("invalid YouTube watch URL, missing video ID")
+	}
+
+	// Update path if embed is requested.
+	embed := false
+	if _, ok := r.URL.Query()["embed"]; ok {
+		embed = true
+		u.Path = fmt.Sprintf("/embed/%s", videoID)
+		u.RawQuery = "" // Reset query parameters.
+	}
```
ao-david, line 636: probably need an else here to keep the ```watch?v=```

ao-david (changes requested):

cmd/oceanbench/broadcast.go
trekhopton: Fixed with an if statement like you suggested below.

cmd/oceanbench/broadcast.go
```diff
@@ -602,15 +603,56 @@ func liveHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	// Update path if embed is requested, otherwise keep the video ID in the query.
+	newQuery := url.Values{}
+	if _, ok := r.URL.Query()["embed"]; ok {
+		u.Path = fmt.Sprintf("/embed/%s", videoID)
+		u.RawQuery = "" // Reset query parameters.
+		// Always set rel=0 for embedded videos.
+		newQuery.Set("rel", "0")
+	} else {
+		newQuery.Set("v", videoID)
```
ao-david, line 641: Not that I think it really matters, but I'm not sure that the mute and autoplay params make any difference to a watch request, and so you can probably just return here.

Although that may be an unintuitive flow for any future changes to the transform with other params.

ao-david (approved):
Make a patch bump

PR #447, opened 2025-01-31
update gcloud ignored

trekhopton:
Added node modules in new ausoceantv/webapp folder to the gcloud ignore file as well as the youtube-api-credentials.json file in the oceantv folder which gets copied there from time to time.

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -26,6 +26,7 @@ cmd/oceantv/oceantv
 cmd/oceantv/*~
 cmd/oceantv/#*
 cmd/oceantv/store/*
+cmd/oceantv/youtube-api-credentials.json
 # oceancron
 cmd/oceancron/.git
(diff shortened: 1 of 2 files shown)


.gcloudignore
```diff
@@ -52,4 +53,5 @@ cmd/vidforward/#*
 # ausoceantv
 cmd/ausoceantv/.git
 cmd/ausoceantv/.gitignore
-cmd/ausoceantv/node_modules/
```
ao-david, line 55: Is it worth just leaving this here, since git won't remove the files automatically, it would be reasonably easy for someone to accidentally upload node_modules

ao-david (approved):

.gcloudignore
trekhopton: I did leave it there, for some reason the diff is shown weirdly.

PR #450, opened 2025-02-02
Set privacy on broadcast completion

saxon-milton:
This will allow us to broadcast live with unlisted privacy, but then on completion, set to public. This means we can restrict live viewing to the AusOceanTV platform, but still have past broadcasts on the YouTube channel.

### DIFF
--- a/cmd/oceanbench/t/broadcast.html
+++ b/cmd/oceanbench/t/broadcast.html
@@ -86,11 +86,22 @@
               </div>
             </div>
             <div class="d-flex align-items-center gap-1 mb-1">
-              <div class="w-25"></div>
+              <label for="account" class="w-25 text-end">Live Privacy:</label>
               <div class="d-flex align-items-center gap-3 flex-row">
                 {{range .Settings.Privacy}}
                 <div>
-                  <input type="radio" name="privacy" value="{{.}}" id="{{.}}-radio" {{if eq . $.CurrentBroadcast.Privacy}}checked{{end}} />
+                  <input type="radio" name="live-privacy" value="{{.}}" id="{{.}}-radio" {{if eq . $.CurrentBroadcast.LivePrivacy}}checked{{end}} />
+                  <label class="text-capitalize" for="{{.}}-radio">{{.}}</label>
+                </div>
+                {{end}}
+              </div>
+            </div>
+            <div class="d-flex align-items-center gap-1 mb-1">
+              <label for="account" class="w-25 text-end">Post Live Privacy:</label>
+              <div class="d-flex align-items-center gap-3 flex-row">
+                {{range .Settings.Privacy}}
+                <div>
+                  <input type="radio" name="post-live-privacy" value="{{.}}" id="{{.}}-radio" {{if eq . $.CurrentBroadcast.PostLivePrivacy}}checked{{end}} />
                   <label class="text-capitalize" for="{{.}}-radio">{{.}}</label>
                 </div>
                 {{end}}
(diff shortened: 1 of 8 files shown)


cmd/oceanbench/t/broadcast.html
```diff
@@ -90,7 +90,18 @@ <h2 class="pt-5">Stream Settings</h2>
               <div class="d-flex align-items-center gap-3 flex-row">
                 {{range .Settings.Privacy}}
                 <div>
-                  <input type="radio" name="privacy" value="{{.}}" id="{{.}}-radio" {{if eq . $.CurrentBroadcast.Privacy}}checked{{end}} />
+                  <input type="radio" name="live-privacy" value="{{.}}" id="{{.}}-radio" {{if eq . $.CurrentBroadcast.LivePrivacy}}checked{{end}} />
```
ao-david, line 93: These should have a label so that a user know which is live and which is post-live
image

cmd/oceantv/broadcast_manager.go
```diff
@@ -356,10 +356,11 @@ func (m *OceanBroadcastManager) SetupSecondary(ctx Ctx, cfg *Cfg, store Store) e
 		// configuration, except for a few of the fields.
 		_cfg.Name = secondaryName
 		_cfg.StreamName = secondaryName
-		_cfg.Privacy = "unlisted" // We don't want the secondary broadcast to be easily discovered by youtube watchers.
-		_cfg.OnActions = ""       // We don't need it to have any control of the camera hardware.
-		_cfg.OffActions = ""      // Ditto.
-		_cfg.SendMsg = true       // It would be handy to have sensors stored in the store broadcasts too.
+		_cfg.LivePrivacy = "unlisted" // We don't want the secondary broadcast to be easily discovered by youtube watchers.
+		_cfg.PostLivePrivacy = "public"
```
ao-david, line 360: I'm not familiar with this part of the code, is this the config for all secondary streams or just the default? I think that if it is the actual config that the default value shouldn't be public. Especially while we are still testing secondary streams.

cmd/oceantv/broadcast_service.go
```diff
@@ -298,3 +299,27 @@ func (s *YouTubeBroadcastService) RTMPKey(ctx context.Context, streamName string
 func (s *YouTubeBroadcastService) PostChatMessage(cID, msg string) error {
 	return broadcast.PostChatMessage(cID, s.tokenURI, msg)
 }
+
+// SetBroadcastPrivacy sets the broadcast privacy of the broadcast with
+// identification ID to the provided privacy using the YouTube API.
+// The privacy can be one of "public", "unlisted", or "private".
```
ao-david, line 305: It might be good to make it clear when this can be called. ie. can it be done whilst the broadcast is live, or does it need to be done before or after etc...

ao-david (approved):
Nice, LTGM

trekhopton (approved):
Nice.

cmd/oceanbench/t/broadcast.html
saxon-milton: whoops, I just copied the first one, which also didn't have a label ?

cmd/oceantv/broadcast_manager.go
saxon-milton: hhmm, yeah you're probably right. This gets set up programmatically so maybe for now I'll set to unlisted.

cmd/oceantv/broadcast_service.go
saxon-milton: yeah sure

PR #451, opened 2025-02-04
Bench: Add search shortcut to monitor page.

ao-david:
This change allows a user to easily click through to search a given sensor, saving a number of form submissions to achieve the same result otherwise.

image


trekhopton (approved):
That's great! Assuming you've tested it, LGTM!

PR #454, opened 2025-02-05
AOTV: Handle stripe webhooks (#421)

### LINKED ISSUE #421: AOTV: Use stripe billing periods
Currently we are calculating subscription periods separately to stripe. Since stripe handles the billing, we should make sure that our periods align with the stripe period.

ao-david:
This change creates a new endpoint that stripe will send events to.
The events that we currently handle are payment_intent.succeeded, and invoice_payment.succeeded which will tell us when payments are successful for one off and subscriptions respectively.
This will help us stay up to date with the stripe database.

closes #421

### DIFF
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -85,6 +85,7 @@ func registerAPIRoutes(app *fiber.App) {
 
 	if !svc.lite {
 		v1.Group("/stripe").
+			Post("/webhook", svc.handleStripeWebhook).
 			Options("/create-payment-intent", svc.preFlightOK).
 			Post("/create-payment-intent", svc.handleCreatePaymentIntent).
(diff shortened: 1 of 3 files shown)


model/subscription.go
```diff
@@ -98,24 +100,70 @@ func GetSubscription(ctx context.Context, store datastore.Store, sid, fid int64)
...
 }
 
+// WithStartEnd sets the start and finish time of the subscription to the passed unix timestamps. This should be
+// used whenever a subscription is linked with an external service which sets the start and end times.
+func WithStartEnd(start, end int64) subscriptionOption {
+	return func(s *Subscription) error {
+		s.Start = time.Unix(start, 0)
+		s.Finish = time.Unix(end, 0)
+
+		// Approximate durations
```
trekhopton, line 110: Fullstop.

PR #461, opened 2025-02-06
ausoceantv: make header a lit element

trekhopton:
This was done so we can reuse it across pages.

I ran prettier which seems to have changed other ts files.

Set VITE_LITE as true for now.


cmd/ausoceantv/webapp/src/web-components/app-header.ts
```diff
@@ -0,0 +1,44 @@
+import { html } from "lit";
+import { customElement } from "lit/decorators.js";
+import { TailwindElement } from "../shared/tailwind.element.ts";
+
+@customElement("app-header")
+export class AppHeader extends TailwindElement() {
+  connectedCallback() {
+    super.connectedCallback();
+    if (!document.querySelector('link[href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.0.0-beta3/css/all.min.css"]')) {
```
ao-david, line 9: Why do we need to do this?

ao-david (approved):

cmd/ausoceantv/webapp/src/web-components/app-header.ts
trekhopton: I'm not sure, for some reason it doesn't seem to work without it.

PR #463, opened 2025-02-07
Bench: Use different endpoint for /config

ao-david:
This change moves the configuration endpoint to a sandbox/ endpoint as well as moving the handler function to sandbox.go.

This is just a straight copy of the existing function, plus changing the endpoint.


cmd/oceanbench/sandbox.go
```diff
@@ -0,0 +1,187 @@
+/*
+NAME
+  Ocean Bench sandbox handlers.
```
trekhopton, line 3: Just remove this NAME section. It was originally for the file name but we don't need that, and we don't need a description either in most cases.

trekhopton (approved):

PR #465, opened 2025-02-07
ausoceantv: light theme

trekhopton:
image

image

this was done after comments about modernising the styling. I think this is a step in the right direction.


ao-david:
Could we get a screenshot please :blush:

ao-david (approved):

PR #466, opened 2025-02-07
ausoceantv: create banner for limited free offer

trekhopton:
This adds the banner and a description about why we're doing this and a button to the support page of our website.

image

image


cmd/ausoceantv/webapp/src/web-components/free-banner.ts
ao-david, line 1: It would be good if this would collapse as well if the user clicks anywhere but on the banner

ao-david (approved):

cmd/ausoceantv/webapp/src/web-components/free-banner.ts
trekhopton: I'll leave that out for now but noted.

PR #467, opened 2025-02-07
ausoceantv: make watch page have light theme and loading animation

trekhopton:
image

This was done to match the other pages and get rid of the "90s" gradient. While the YouTube video hasn't loaded the user can now see a loading animation.


cmd/ausoceantv/webapp/watch.html
```diff
@@ -8,22 +8,52 @@
...
             </div>
+
+            <!-- Video Description -->
+            <p class="mt-4 text-center text-sm text-gray-300">Rapid Bay live via AusOceanTV.</p>
           </div>
         </div>
       </div>
     </auth-wrapper>
+
+    <script>
```
ao-david, line 49: Does having this at the bottom guarantee that it only fires after the DOM has loaded? If not, might be good to use a onload event on the body.

cmd/ausoceantv/webapp/watch.html
```diff
@@ -8,22 +8,52 @@
     <link rel="stylesheet" href="./src/index.css" />
     <script type="module" src="src/web-components/authenticator.ts"></script>
     <script type="module" src="src/web-components/app-header.ts"></script>
+    <style>
+      /* Loading Animation Styling */
+      @keyframes spin {
```
ao-david, line 13: I'm not sure that this is still actually being used? it looks like you're using the tailwind class instead?

ao-david (approved):
Just delete the unused CSS, but otherwise LGTM.

PR #468, opened 2025-02-10
Model: Update subscriber and subscription functions

ao-david:
This change updates some of the handling and indexing for subscriber and subscription functions.

### DIFF
--- a/model/subscriber.go
+++ b/model/subscriber.go
@@ -90,10 +90,23 @@ func CreateSubscriber(ctx context.Context, store datastore.Store, s *Subscriber)
 	s.Created = time.Now()
 
 	// Otherwise generate and use a unique ID.
+	q := store.NewQuery(typeSubscriber, true, "ID", "Email")
 	for {
 		s.ID = utils.GenerateInt64ID()
+		err := q.FilterField("ID", "=", s.ID)
+		if err != nil {
+			return err
+		}
+		keys, err := store.GetAll(ctx, q, nil)
+		if err != nil {
+			return err
+		}
+		if len(keys) != 0 {
+			continue
+		}
+
 		key := store.NameKey(typeSubscriber, fmt.Sprintf("%d.%s", s.ID, s.Email))
-		err := store.Create(ctx, key, s)
+		err = store.Create(ctx, key, s)
 		if err == nil {
 			return nil
 		} else if err != datastore.ErrEntityExists {
--- a/model/subscription.go
+++ b/model/subscription.go
@@ -25,6 +25,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/ausocean/openfish/datastore"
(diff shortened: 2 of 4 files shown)


model/subscriber.go
```diff
@@ -90,10 +90,23 @@ func CreateSubscriber(ctx context.Context, store datastore.Store, s *Subscriber)
...
 
 	// Otherwise generate and use a unique ID.
+	q := store.NewQuery(typeSubscriber, true, "ID", "Email")
 	for {
 		s.ID = utils.GenerateInt64ID()
+		err := q.FilterField("ID", "=", s.ID)
+		if err != nil {
+			return err
+		}
+		keys, err := store.GetAll(ctx, q, nil)
```
trekhopton, line 100: We shouldn't need to do a `GetAll` every time we create a subscriber, instead generate the ID, then do `store.Create`.
Most times this will work (no collision), but in the rare case of a collision (`ErrEntityExists` from `Create`) then we handle it (`continue`).

model/subscription.go
```diff
@@ -91,31 +100,103 @@ func GetSubscription(ctx context.Context, store datastore.Store, sid, fid int64)
...
 	if len(subscriptions) > 1 {
 		return nil, fmt.Errorf("for SubscriberID: %d, and FeedID: %d, failed with error: %w", sid, fid, errDuplicateSubscriptions)
 	}
 
 	return &subscriptions[0], nil
 }
 
+// WithStartEnd sets the start and finish time of the subscription to the passed unix timestamps. This should be
+// used whenever a subscription is linked with an external service which sets the start and end times.
+func WithStartEnd(start, end int64) subscriptionOption {
```
trekhopton, line 116: Wasn't this and the below function in another PR I looked at?

trekhopton (changes requested):

model/subscriber.go
ao-david: The issue is that the entities use nameKeys and so multiple entities with the same ID can exist, but with different other values.

for example, consider there is an existing account with email ```one@test.org``` and ID ```123``` the key will be ```123.one@test.org```

if another subscriber is generated with the ID ```123`` and email ```two@test.org```, the key will be ```123.two@test.org``` which is different to the first key even though the ID is different.

unless I'm missing something, I think this is necessary.

trekhopton:
Just the order of checking for collisions that needs changing.

model/subscription.go
ao-david: Yes, that was #454, but there were a few issues I discovered after making that PR, I have split the PR into a few smaller portions now to hopefully be a little easier too review. But does mean that there was some overlap sorry :pray:

model/subscriber.go
trekhopton: Oh good point, why did we decide to put the email in the key? If the ID is unique then that is sufficient. I actually think we should remove the emails from the keys, unless I'm missing something.

trekhopton: That might have been an oversight by me, I was thinking we should do it like the cloudblue user but that need to have the email because it's the site+email to be unique, but this subscriber only needs the ID. The filestore issue is another thing but that should be what causes us to add things to the keys.

trekhopton: I'll approve this PR but I think we should now consider getting rid of the email from the subscription key.. better to do it now than with lots of users.

model/subscription.go
trekhopton: No worries.

trekhopton (approved):

PR #470, opened 2025-02-10
ausoceantv: user survey

trekhopton:
Adds a survey page that will show for users that are over a day old. After completion they get redirected to the home page.

Handlers were added to check for survey completion and handle survey submission.

This is not the final form of the survey. The location will be made more precise ie. City / Suburb (To get post code) so we can have more targeted demographic info in SA.

image

### DIFF
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -86,7 +86,7 @@ func registerAPIRoutes(app *fiber.App) {
 
 	v1.Group("/survey").
 		Get("/check", svc.checkSurveyHandler).
-		Post("/submit", svc.handleSurveyFormSubmission)
+		Post("/", svc.handleSurveyFormSubmission)
 
 	if !svc.lite {
 		v1.Group("/stripe").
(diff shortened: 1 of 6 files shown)


cmd/ausoceantv/main.go
```diff
@@ -246,3 +251,84 @@ func (svc *service) getSubscriptionHandler(c *fiber.Ctx) error {
...
+	if errors.Is(err, gauth.SessionNotFound) || errors.Is(err, gauth.TokenNotFound) {
+		return fiber.NewError(fiber.StatusUnauthorized, "user not authenticated")
+	} else if err != nil {
+		return fmt.Errorf("unable to get profile: %w", err)
+	}
+
+	// Fetch the subscriber by email from the datastore.
+	subscriber, err := model.GetSubscriberByEmail(ctx, svc.store, p.Email)
+	if err != nil {
+		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{"error": "subscriber not found"})
```
ao-david, line 308: Not that I think that it should be done here, but I think we should have a think about how we are handling and returning errors. I think every time we make a new change we implement a new way of error handling. 

Maybe we could use some more middleware or something?

cmd/ausoceantv/main.go
```diff
@@ -246,3 +251,84 @@ func (svc *service) getSubscriptionHandler(c *fiber.Ctx) error {
...
+	}
+
+	// Encode demographic info as JSON and store it in Subscriber.
+	demographicJSON, err := json.Marshal(demographicInfo)
+	if err != nil {
+		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{"error": "failed to encode demographic info"})
+	}
+	subscriber.DemographicInfo = string(demographicJSON)
+
+	// Save updated subscriber to Datastore.
```
ao-david, line 328: either stick to capital or lowercase for datastore in comments

cmd/ausoceantv/webapp/watch.html
```diff
@@ -16,12 +16,12 @@
 
         <!-- Video Section -->
         <div class="flex flex-1 items-center justify-center p-6">
-          <div class="bg-black bg-opacity-80 shadow-lg rounded-2xl w-full max-w-[1920px] p-4 md:p-24">
```
ao-david, line 19: Are these changes relevant to this PR?

ao-david: Are we also implementing the survery onto the watch page? Otherwise a user could just navigate directly to /watch.html right?

cmd/ausoceantv/webapp/survey.html
```diff
@@ -0,0 +1,93 @@
...
+              <!-- Interest Dropdown -->
+              <div class="text-left">
+                <label for="user-category" class="block text-lg font-semibold text-gray-900">Why are you interested in underwater livestreams?</label>
+                <select id="user-category" name="user-category" class="mt-2 w-full rounded-lg border-gray-300 p-3 text-gray-800">
+                  <option value="">Select your interest...</option>
+                  <option value="scuba-diving">Scuba Diving</option>
+                  <option value="recreational-fishing">Recreational Fishing</option>
+                  <option value="commercial-fishing">Commercial Fishing</option>
+                  <option value="snorkelling">Snorkelling</option>
+                  <option value="snorkelling">Spear Fishing</option>
```
ao-david, line 72: inconsistent value to labelling

cmd/ausoceantv/webapp/src/survey.ts
```diff
@@ -0,0 +1,40 @@
...
+      method: "POST",
+      headers: { "Content-Type": "application/x-www-form-urlencoded" },
+      body: new URLSearchParams({ city, "user-category": interest }).toString(),
+    });
+
+    if (!response.ok) {
+      const error = await response.json();
+      alert(`Error: ${error.error}`);
+    } else {
+        alert("Survey successfully submitted! Redirecting to home...");
```
ao-david, line 19: Do we actually want this alert here?

cmd/ausoceantv/main.go
```diff
@@ -83,6 +84,10 @@ func registerAPIRoutes(app *fiber.App) {
 
 	v1.Get("version", svc.versionHandler)
 
+	v1.Group("/survey").
```
ao-david, line 87: Possibly worth considering what this endpoint could grow into, easier to change now than in the future.
ie. are we going to have multiple surveys that we''ll want to check by ID? In which case maybe we could do a /check/:surveyID and /submit/:surveyID

ao-david (approved):
Just a few comments.

cmd/ausoceantv/main.go
trekhopton: Yeah good point haha

cmd/ausoceantv/main.go
trekhopton: I think survey/ is a good first group, it denotes anything following is survey related. If we want to do more surveys with IDs we could do survey/check/123 etc. 

I might however remove the submit part of the path. We could just have POST /survey

cmd/ausoceantv/webapp/src/survey.ts
trekhopton: Nope. Meant to get rid of it.

cmd/ausoceantv/main.go
ao-david: Yeah, I don't think there's anything inherently wrong with it, but I do think that it is worth making sure that we've at least thought about how it could be extended.

cmd/ausoceantv/webapp/watch.html
trekhopton: No, this is just a side effect from running npm run format. Obviously it wasn't run when I merged these changes.

Yes, we should also redirect on the watch page.

trekhopton: I think I'll do the watch page redirection in another PR.

Issue #420, opened 2025-01-02
AOTV: Cancel Stripe subscription


ao-david:
Currently we are only marking the AusOcean subscription entity as non-renewing, but we also need to update the stripe subscription to make sure that the customer doesn't get charged for a cancelled subscription.

PR #471, opened 2025-02-10
AOTV: Update stripe subscription to cancel (#420)

ao-david:
This change update the subscription on the stripe server to make sure that we are in sync when cancelling to avoid users getting charged when it seemed like they had cancelled the subscription.

This change also adds a few checks and conditional rendering in the profile-details element to make it easier to see the state of the subscription.

### DIFF
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -238,6 +238,9 @@ func (svc *service) getSubscriptionHandler(c *fiber.Ctx) error {
 	}
 
 	subscription, err := model.GetSubscription(ctx, svc.store, subscriber.ID, model.NoFeedID)
+	if errors.Is(err, datastore.ErrNoSuchEntity) {
+		return c.JSON(nil)
+	}
 	if err != nil {
 		return fmt.Errorf("error getting subscription for id: %d: %w", subscriber.ID, err)
 	}
--- a/cmd/ausoceantv/stripe.go
+++ b/cmd/ausoceantv/stripe.go
@@ -379,13 +379,16 @@ func (svc *service) cancelSubscription(c *fiber.Ctx) error {
 		return fmt.Errorf("error getting subscription for id: %d: %w", subscriber.ID, err)
 	}
 
+	// Since the day pass is not a renewing subscription, we shouldn't need to cancel it with stripe.
+	if sub.Class != model.SubscriptionDay {
+		subParams := &stripe.SubscriptionParams{CancelAtPeriodEnd: stripe.Bool(true)}
+		_, err = subscription.Update(sub.StripeSubscriptionID, subParams)
+		if err != nil {
+			return fmt.Errorf("failed to cancel subscription: %w", err)
+		}
+	}
+
 	sub.Renew = false
-
-	// TODO: Update the Stripe subscription to cancel at period end.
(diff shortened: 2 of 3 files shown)


cmd/ausoceantv/main.go
```diff
@@ -233,6 +233,9 @@ func (svc *service) getSubscriptionHandler(c *fiber.Ctx) error {
 	}
 
 	subscription, err := model.GetSubscription(ctx, svc.store, subscriber.ID, model.NoFeedID)
+	if errors.Is(err, datastore.ErrNoSuchEntity) {
+		return c.JSON(nil)
+	}
 	if err != nil {
 		return fmt.Errorf("error getting subscription for id: %d: %w", subscriber.ID, err)
```
trekhopton, line 240: I know this is not this change but our handlers should return valid JSON for errors I think. That was the client can handle them in the browser much nicer.

cmd/ausoceantv/stripe.go
```diff
@@ -379,13 +379,16 @@ func (svc *service) cancelSubscription(c *fiber.Ctx) error {
...
-	// TODO: Update the Stripe subscription to cancel at period end.
-	log.Panic("cannot cancel by stripe, unimplemented")
-	c.WriteString("failed to cancel subscription, please try again, or contact tv@ausocean.org")
-	c.Status(fiber.StatusInternalServerError)
+	// Since the day pass is not a renewing subscription, we shouldn't need to cancel it with stripe.
+	if sub.Class != model.SubscriptionDay {
+		subParams := &stripe.SubscriptionParams{CancelAtPeriodEnd: stripe.Bool(true)}
+		_, err = subscription.Update(sub.StripeSubscriptionID, subParams)
+		if err != nil {
+			return fmt.Errorf("failed to cancel subscription: %w", err)
```
trekhopton, line 387: I think it would be best to log with fiber's log package, then return the error in JSON for the client:

```
log.Errorf("failed to cancel subscription: %w", err)
return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
    "error": "Internal Server Error",
})
```

cmd/ausoceantv/webapp/src/web-components/profile-details.ts
```diff
@@ -84,25 +102,26 @@ export class ProfileDetails extends TailwindElement() {
             <td class="p-2 pb-0">${this.subscription.Class}</td>
           </tr>
           <tr>
-            <td class="p-2 pb-0"><strong>Next Billing Date:</strong></td>
+            <td class="p-2 pb-0"><strong>${this.subscription.Renew ? "Next Billing Date:" : "Cancelling On:"}</strong></td>
```
trekhopton, line 105: I think "Subscription Ends" makes more sense than "Cancelling On" since it's the user that did the cancelling by pressing the cancel button.

trekhopton (changes requested):
Screenshot would be good.

cmd/ausoceantv/main.go
ao-david: Take a look at #474

cmd/ausoceantv/stripe.go
ao-david: I might also handle this with a separate change and tackle them all at once.

ao-david:
> Screenshot would be good.

image

PR #473, opened 2025-02-11
otv: fix composite store missing kinds

saxon-milton:
This change adds some missing kinds to the ausocean composite store factory. We're also wrapping the store get process so that we can have better panic messaging if the mapping doesn't exist.


ao-david (approved):
LGTM

PR #475, opened 2025-02-12
ausoceantv: use google maps API for location field autocomplete in survey

trekhopton:
This was done so that we can quickly and reliably get the post code and city/suburb of the user, otherwise this might deter them from completing the survey.

image

image


ao-david:
There seem to be a lot of commits in this PR are they all for this PR?

trekhopton:
> There seem to be a lot of commits in this PR are they all for this PR?

No, this seems to happen if I initially branch off a previous branch. I should probably rebase?

ao-david:
> > There seem to be a lot of commits in this PR are they all for this PR?
> 
> No, this seems to happen if I initially branch off a previous branch. I should probably 

If the branch has been merged, I probably wouldn't worry about it, since you already have merge commits in the history, so I suspect it would get very messy (and the actual changes in the PR will stay the same). If it hasn't been merged, make the PR target that branch

trekhopton:
> > > There seem to be a lot of commits in this PR are they all for this PR?
> > 
> > 
> > No, this seems to happen if I initially branch off a previous branch. I should probably
> 
> If the branch has been merged, I probably wouldn't worry about it, since you already have merge commits in the history, so I suspect it would get very messy (and the actual changes in the PR will stay the same). If it hasn't been merged, make the PR target that branch

Yeah it has been merged.

cmd/ausoceantv/webapp/src/survey.ts
```diff
@@ -3,13 +3,14 @@ async function handleFormSubmit(event: Event): Promise<void> {
...
-  const interest = (document.querySelector("#user-category") as HTMLSelectElement).value;
+  const postcode = (document.querySelector("#postcode") as HTMLSelectElement).value;
+  const userCategory = (document.querySelector("#user-category") as HTMLSelectElement).value;
 
   try {
     const response = await fetch("/api/v1/survey", {
       method: "POST",
       headers: { "Content-Type": "application/x-www-form-urlencoded" },
-      body: new URLSearchParams({ city, "user-category": interest }).toString(),
+      body: new URLSearchParams({ city: city, postcode: postcode, "user-category": userCategory }).toString(),
```
ao-david, line 13: Should the labels in this struct be in quotes?
```ts
      body: new URLSearchParams({ "city": city, "postcode": postcode, "user-category": userCategory }).toString(),
```

ao-david: Otherwise I think you might end up with:
```json
{"adelaide":"adelaide", "5001":"5001", "user-category":"snorkeller"}
```

cmd/ausoceantv/webapp/src/survey.ts
```diff
@@ -3,13 +3,14 @@ async function handleFormSubmit(event: Event): Promise<void> {
   event.preventDefault();
 
   const city = (document.querySelector("#city") as HTMLSelectElement).value;
-  const interest = (document.querySelector("#user-category") as HTMLSelectElement).value;
+  const postcode = (document.querySelector("#postcode") as HTMLSelectElement).value;
```
ao-david, line 6: postcode is actually an input element not a select element

ao-david (approved):
Just a small issue.

I also wonder if maybe we could add a little message that says why we are collecting the data?

ie. this will only be used to improve your experience or something?

trekhopton:
> Just a small issue.
> 
> I also wonder if maybe we could add a little message that says why we are collecting the data?
> 
> ie. this will only be used to improve your experience or something?

Could do but I suppose we may want to use it in other ways like display it on their profile or something in the future.

PR #476, opened 2025-02-12
AOTV: Add logAndReturnError function (#474)

### LINKED ISSUE #474: AOTV: Create helper log/return error function
At the moment, we are returning and logging errors in a variety of different ways in different places, it would be nice if we had some sort of helper function which could log errors, and return usefully formatted errors, reducing code duplication and variance.

ao-david:
This change adds a simple helper function which can log an error, as well as return an error using properly formatted JSON.

Callers can set a different value to be returned to the frontend, as well as specifying the response code through functional options.

This change may introduce more logging than previously, we may have to replace some calls to just return errors, or even have another functional option to set the log level.

requires #471 
closes #474

### DIFF
--- a/cmd/ausoceantv/stripe.go
+++ b/cmd/ausoceantv/stripe.go
@@ -94,8 +94,7 @@ func (svc *service) handleStripeWebhook(c *fiber.Ctx) error {
 	event := stripe.Event{}
 
 	if err := json.Unmarshal(c.Body(), &event); err != nil {
-		log.Errorf("Failed to parse webhook body json: %w", err.Error())
-		return err
+		return logAndReturnError(c, fmt.Sprintf("failed to parse webhook body json: %v", err.Error()))
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/ausoceantv/stripe.go
```diff
@@ -94,42 +94,38 @@ func (svc *service) handleStripeWebhook(c *fiber.Ctx) error {
 	event := stripe.Event{}
 
 	if err := json.Unmarshal(c.Body(), &event); err != nil {
-		log.Errorf("Failed to parse webhook body json: %w", err.Error())
-		return err
+		return logAndReturnError(c, fmt.Sprintf("Failed to parse webhook body json: %v", err.Error()))
```
trekhopton, line 97: This and the following two error messages start with capitals. Should be lowercase for consistency.

trekhopton (approved):
Legend.

PR #480, opened 2025-02-13
AOTV: Create customer with subscriber (#479)

### LINKED ISSUE #479: AOTV: Create stripe customer when creating subscriber entity
This may allow us to access some handy features through stripe during the free launch and move to the paid launch when users will need to start subscriptions.

ao-david:
This change creates a stripe customer when a new user signs in, this should give us more functionality with access to the customers via stripe.

closes #479

### DIFF
--- a/cmd/ausoceantv/auth.go
+++ b/cmd/ausoceantv/auth.go
@@ -72,6 +72,10 @@ func (svc *service) callbackHandler(c *fiber.Ctx) error {
 		if err != nil {
 			return logAndReturnError(c, fmt.Sprintf("unable to create susbcriber %v: %v", subscriber, err))
 		}
+		_, err = svc.getCustomer(subscriber)
+		if err != nil {
+			return logAndReturnError(c, fmt.Sprintf("unable to create stripe customer: %v", err))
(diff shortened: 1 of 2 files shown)


cmd/ausoceantv/auth.go
```diff
@@ -72,6 +72,10 @@ func (svc *service) callbackHandler(c *fiber.Ctx) error {
 		if err != nil {
 			return logAndReturnError(c, fmt.Sprintf("unable to create susbcriber %v: %v", subscriber, err))
 		}
+		_, err = svc.getCustomer(subscriber)
+		if err != nil {
+			return logAndReturnError(c, fmt.Sprintf("unable to create stripe customer: %v", err))
```
trekhopton, line 77: I think this is an instance where we don't want to send this error message to the user. The above error too. We refer to them as subscribers but that should be internal I think. It might be weird for a user to see "unable to create stripe customer" when they haven't seen anything about stripe. Internal Server Error would perhaps be better, or "We encountered an error processing your request."

trekhopton (approved):

PR #483, opened 2025-02-13
ausoceantv: use geocode instead of postcode and city

trekhopton:
This was done so that we're always capturing the location data despite the possibility of not including the post code in the input.

After this PR we will create an Entity that stores the geocode instead of the JSON being stored in the subscriber. This is just a step in the right direction.

### DIFF
--- a/cmd/ausoceantv/webapp/src/survey.ts
+++ b/cmd/ausoceantv/webapp/src/survey.ts
@@ -2,22 +2,24 @@ async function handleFormSubmit(event: Event): Promise<void> {
   console.log("handling form submission...");
   event.preventDefault();
 
-  const city = (document.querySelector("#city") as HTMLInputElement).value;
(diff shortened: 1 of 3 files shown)


cmd/ausoceantv/webapp/src/survey.ts
```diff
@@ -40,11 +42,10 @@ document.addEventListener("DOMContentLoaded", initFormHandler);
 
 function initAutocomplete(): void {
   const input = document.getElementById("location") as HTMLInputElement;
-  const cityInput = document.getElementById("city") as HTMLInputElement;
-  const postcodeInput = document.getElementById("postcode") as HTMLInputElement;
+  const geocodeInput = document.getElementById("geocode") as HTMLInputElement;
```
ao-david, line 45: We could make these global so that we only have to get them here, and not also in the form submission. Might make it ever so slightly faster.

ao-david (approved):
Nice.

Issue #482, opened 2025-02-13
AOTV: Prevent user from creating multiple subscriptions


ao-david:
Currently, we only support one kind of subscription (to all feeds - ie FeedID = 0) whilst this is the case, we should prevent users from creating multiple subscriptions to avoid people from accidentally signing up multiple times to the same thing which could lead them to dislike the service.

PR #484, opened 2025-02-13
AOTV: Prevent multiple subscriptions (#482)

ao-david:
This change checks if a user has a current subscription before returning a payment intent which would allow them to create a new subscription.

This change also handles the errors better in the frontend by redirecting the user depending on the error.

### DIFF
--- a/cmd/ausoceantv/webapp/src/checkout.js
+++ b/cmd/ausoceantv/webapp/src/checkout.js
@@ -16,23 +16,32 @@ async function initialize() {
   const id = urlParams.get("priceID");
 
   if (!id) {
-    console.log("setting error message");
-    let msg = document.getElementById("msg");
-    msg.innerHTML = "<p>Choose a Plan <a href='/plans.html'>here</a></p>";
-    msg.removeAttribute("hidden");
-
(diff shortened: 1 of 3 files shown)


cmd/ausoceantv/webapp/src/checkout.js
```diff
@@ -119,6 +138,20 @@ function showMessage(messageText) {
...
 
+// redirectWithMessage shows a message to the user, and redirects them via a button
+// to a new location.
+function redirectWithMessage(url, messageText, buttonText) {
+  let info = document.getElementById("cart-info");
+  info.innerHTML = `
+    <p class="text-center mb-1">${messageText}</p>
+    <a href="${url}" class="m-auto w-fit">
+      <button class="rounded-lg bg-[#0c69ad] px-6 py-1 text-lg font-semibold text-white shadow-md transition-all hover:shadow-lg hover:brightness-110">
+        ${buttonText ? buttonText : "Okay"}
```
trekhopton, line 149: I think going with "OK" is more common.

trekhopton (approved):

PR #486, opened 2025-02-14
AOTV: Add more content to landing page

ao-david:
This change adds a little bit of information to the landing page to hopefully help it to rank a little higher in the search results.

This change also adds a footer.

image
image

PR #487, opened 2025-02-14
AOTV: Turn footer into lit element

ao-david:
This change moves the footer into a lit element to make it easily reusable across pages.

requires #486


cmd/ausoceantv/webapp/src/web-components/aotv-footer.ts
```diff
@@ -0,0 +1,33 @@
+import { html } from "lit";
+import { customElement } from "lit/decorators.js";
+import { TailwindElement } from "../shared/tailwind.element.ts";
+
+@customElement("custom-footer")
+export class customFooter extends TailwindElement() {
+  render() {
+    return html`
+      <footer class="z-10 w-full bg-[#0c69ad] p-5 text-white">
```
trekhopton, line 9: I think this z-10 is still causing issues on the watch page with the text going over the video. Granted I think there's a issue with the video going over the footer but I think it would be better if we could apply the z classes to the <custom-element> tag in the html file.. not sure if that's how it works with lit but I think it should..

trekhopton (approved):

cmd/ausoceantv/webapp/src/web-components/aotv-footer.ts
trekhopton: Happy for you to merge and we can address this later though. Good start.

ao-david: Again, where are you seeing this? I can't replicate this at all? The issue that I am seeing now, is that the footer isn't always at the bottom of the screen, and can float sometimes.

I've tested chrome, firefox and on my phone and I don't get any issues with text over video

ao-david: Sorry, that sounds rude, I'm just curious how to replicate sorry :pray:

trekhopton: Hm strange, I can't actually replicate it with this branch, only when I merged it into my snapshot branch. Perhaps I lost something important in the merge.

PR #490, opened 2025-02-14
ausoceantv: make SubscriberRegion entity, methods and use it in the survey

trekhopton:
This was done so that the location (region) data is query-able in the datastore.

### DIFF
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -331,10 +331,10 @@ func (s *service) handleSurveyFormSubmission(c *fiber.Ctx) error {
 		return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"message": "request body cannot be empty"})
 	}
 
-	// Directly store the JSON in DemographicInfo.
+	// Encode demographic info as JSON and store it in Subscriber.
 	subscriber.DemographicInfo = string(body)
 
-	// Save updated subscriber data in Datastore.
+	// Save updated subscriber to datastore.
 	if err := model.UpdateSubscriber(ctx, s.store, subscriber); err != nil {
 		log.Error("failed to update subscriber:", err)
 		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{"message": "failed to update subscriber data"})
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -320,6 +320,20 @@ func (s *service) handleSurveyFormSubmission(c *fiber.Ctx) error {
 		return logAndReturnError(c, "empty request body")
 	}
 
+	// Parse the JSON body directly into SubscriberRegion.
+	subscriberRegion := &model.SubscriberRegion{}
+	if err := json.Unmarshal(body, subscriberRegion); err != nil {
+		return logAndReturnError(c, "failed to parse region data")
+	}
+
+	// Set the SubscriberID field.
+	subscriberRegion.SubscriberID = subscriber.ID
+
+	// Create the SubscriberRegion entity using the new helper function.
+	if err := model.CreateSubscriberRegion(ctx, s.store, subscriberRegion); err != nil {
+		return logAndReturnError(c, "failed to save subscriber region")
+	}
+
 	// Encode demographic info as JSON and store it in Subscriber.
 	subscriber.DemographicInfo = string(body)
 
--- a/cmd/ausoceantv/main.go
+++ b/cmd/ausoceantv/main.go
@@ -323,23 +323,31 @@ func (s *service) handleSurveyFormSubmission(c *fiber.Ctx) error {
 	// Parse the JSON body directly into SubscriberRegion.
 	subscriberRegion := &model.SubscriberRegion{}
 	if err := json.Unmarshal(body, subscriberRegion); err != nil {
-		return logAndReturnError(c, "failed to parse region data")
+		return logAndReturnError(c, fmt.Sprintf("failed to parse region data: %v", err))
 	}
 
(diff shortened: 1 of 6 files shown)


model/region.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+// CreateSubscriberRegion creates a new SubscriberRegion in the datastore.
+func CreateSubscriberRegion(ctx context.Context, store datastore.Store, r *SubscriberRegion) error {
+	if r == nil {
+		return errors.New("SubscriberRegion is nil")
+	}
+
+	// If the SubscriberRegion has an ID, use that.
```
ao-david, line 75: This comment makes it sound like an optional check. It could be more like:
```
check that SubscriberRegion has an ID
```

model/region.go
```diff
@@ -0,0 +1,129 @@
...
+	// Since datastore uses kind and key to reference a particular entity, using the subscriberID as the SubscriberRegion key is fine.
+	key := store.IDKey(TypeSubscriberRegion, r.SubscriberID)
+	err := store.Create(ctx, key, r)
+	if err != nil {
+		return fmt.Errorf("error creating SubscriberRegion: %w", err)
+	}
+	return nil
+}
+
+// UpdateSubscriberRegion updates the SubscriberRegion matching sr.SubscriberID.
```
ao-david, line 88: sr.SubscriberID -> r.SubscriberID

model/region.go
```diff
@@ -0,0 +1,129 @@
...
+func UpdateSubscriberRegion(ctx context.Context, store datastore.Store, r *SubscriberRegion) error {
```
ao-david, line 89: It feels like the UpdateSubscriberRegion and PutSubscriberRegion almost seem to be the wrong way round. Update is putting without checking that it exists.

model/region.go
```diff
@@ -0,0 +1,129 @@
...
+	key := store.IDKey(TypeSubscriberRegion, r.SubscriberID)
+	_, err := store.Put(ctx, key, r)
+	if err != nil {
+		return fmt.Errorf("failed to put SubscriberRegion: %w", err)
+	}
+
+	return nil
+}
+
+// GetSubscriberRegion gets the SubscriberRegion with the given SubscriberID.
```
ao-david, line 110: shouldn't subsciberID be an int64 here?

model/region.go
```diff
@@ -0,0 +1,129 @@
...
+func GetSubscriberRegion(ctx context.Context, store datastore.Store, subscriberID string) (*SubscriberRegion, error) {
+	q := store.NewQuery(TypeSubscriberRegion, false, "SubscriberID")
+	q.FilterField("SubscriberID", "=", subscriberID)
+	var regions []SubscriberRegion
+	_, err := store.GetAll(ctx, q, &regions)
```
ao-david, line 115: I don't think that we should need to getAll if you have the whole ID?

model/region.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get SubscriberRegion: %w", err)
+	}
+
+	if len(regions) == 0 {
+		return nil, datastore.ErrNoSuchEntity
+	}
+
+	if len(regions) > 1 {
+		return nil, fmt.Errorf("duplicate SubscriberRegion entries found for SubscriberID: %s", subscriberID)
```
ao-david, line 125: This should be made into a proper error so that it can be compared using errors.Is

cmd/ausoceantv/main.go
```diff
@@ -320,12 +320,54 @@ func (s *service) handleSurveyFormSubmission(c *fiber.Ctx) error {
...
+	var surveyData map[string]interface{}
+	if err := json.Unmarshal(body, &surveyData); err != nil {
+		return logAndReturnError(c, fmt.Sprintf("failed to parse survey data: %v", err))
+	}
+
+	demographicInfo := make(map[string]string)
+	if userCategory, ok := surveyData["user-category"].(string); ok {
+		demographicInfo["user-category"] = userCategory
+	} else {
+		demographicInfo["user-category"] = "unknown"
```
ao-david, line 358: If we are requiring this survey to be completed, why can't we just leave this blank to check in the future? I thought we were checking if the value was empty to see if they had completed the survey or not yet.

cmd/ausoceantv/main.go
```diff
@@ -320,12 +320,54 @@ func (s *service) handleSurveyFormSubmission(c *fiber.Ctx) error {
...
+	}
+	demographicJSON, err := json.Marshal(demographicInfo)
```
ao-david, line 360: Do we still need to encode into JSON?

ao-david (changes requested):
I think the entity methods need a little tweak

model/region.go
trekhopton: Yeah, I think I'll just remove UpdateSubscriberRegion for now, I mean to implement Put.

I was basing it off of the UpdateSubscriber method which just looks like it uses Put.

model/region.go
trekhopton: Yup.

model/region.go
trekhopton: True.

model/region.go
trekhopton: No longer an issue since I'm now just using Get.

trekhopton:
> I think the entity methods need a little tweak

Comments addressed.

model/region.go
ao-david: yeah my datastore methods ended up being terrible aha, don't ever base anything on them...

ao-david (approved):
LGTM

PR #491, opened 2025-02-14
ausoceantv: make watch page background an image

trekhopton:
This was done to immerse the viewer while not being too distracting.

image

image


cmd/ausoceantv/webapp/watch.html
```diff
@@ -10,11 +10,13 @@
     <script type="module" src="src/web-components/app-header.ts"></script>
     <script type="module" src="src/survey-check.ts"></script>
   </head>
-  <body class="h-screen bg-neutral-50 text-white">
+  <body class="h-screen bg-gradient-to-b from-sky-600 via-blue-900 to-black text-white">
     <auth-wrapper>
       <div id="page-column" class="flex h-full flex-col">
         <app-header></app-header>
 
+        <img src="src/assets/web-bg-1.jpg" alt="Background" class="fixed inset-0 z-[-1] h-full w-full object-cover opacity-30" />
```
ao-david, line 18: Should it be h-full w-full or h-screen w-screen?

ao-david (approved):
I'm not sure I'm a fan, but otherwise I think this is okay.

cmd/ausoceantv/webapp/watch.html
trekhopton: I think it should be what it is, w-screen is fixed to the screen size whereas w and h-full are 100% of the container that they're in I believe.

PR #492, opened 2025-02-14
ausoceantv: fix logo corners

trekhopton:
The concentric rectangles corners made the tv part of the logo look like it wasn't a consistent line. This has been fixed in the new image.


ao-david (approved):
Thank you!

PR #493, opened 2025-02-15
otv: update battery voltage pin

saxon-milton:
Make battery voltage pin configurable and default to the ESP32 A4 pin when it's not defined in the broadcast config.


ao-david (approved):

ao-david:
It's probably worth noting when we deploy this we will need to update all the current broadcasts as otherwise they will fail with voltage issues, since they currently use ESP8266s?

saxon-milton:
> It's probably worth noting when we deploy this we will need to update all the current broadcasts as otherwise they will fail with voltage issues, since they currently use ESP8266s?

Indeed

PR #498, opened 2025-02-18
otv: add custom marshalling for nested hardware states

saxon-milton:
Because the nested states use substates of interface state, we need to add custom marshaling/unmarshaling to deal with this.

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -36,7 +36,7 @@ func newableWithContext(new func(ctx *broadcastContext) any, args ...interface{}
 
 type hardwareRestarting struct {
 	stateWithTimeoutFields
-	substate state
+	Substate state
 }
 
 var _ = register(hardwareRestarting{})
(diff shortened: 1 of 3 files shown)


cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+	Type string          `json:"type"`
+	Data json.RawMessage `json:"data"`
+}
+
+func (s *hardwareRestarting) MarshalJSON() ([]byte, error) {
+	substateType := ""
+	substateData := []byte("null")
+
+	if s.Substate != nil {
+		substateType = s.Substate.(registry.Named).Name()
```
ao-david, line 67: Should we ok this assertion?
```golang
namedSS, ok := s.Substate.(registry.Named)
if !ok {...}
substateType = namedSS.Name()
```

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+type hardwareRestartingStateWrapper struct {
+	Type string          `json:"type"`
+	Data json.RawMessage `json:"data"`
+}
+
+func (s *hardwareRestarting) MarshalJSON() ([]byte, error) {
+	substateType := ""
+	substateData := []byte("null")
+
+	if s.Substate != nil {
```
ao-david, line 66: should the alternate be an error? ie
```golang
s.Substate == nil
```

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+		substateType = s.Substate.(registry.Named).Name()
+		data, err := json.Marshal(s.Substate)
+		if err != nil {
+			return nil, fmt.Errorf("could not marshal substate: %w", err)
+		}
+		substateData = data
```
ao-david, line 72: why can't you marshall directly into substateData?
If it is because otherwise it overrides the existing "null" string, is that even valid JSON anyway?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+		StateWithTimeoutFields: s.stateWithTimeoutFields,
+		Substate: hardwareRestartingStateWrapper{
+			Type: substateType,
+			Data: substateData,
+		},
+	}
+
+	data, err := json.Marshal(alias)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal alias: %w", err)
```
ao-david, line 88: This error message could perhaps be a little more user friendly.
ie, could not marshall state with substate using type alias:

this way it's a little more obvious from looking at logs what the alias is referring to?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+		return errors.New("hardwareRestarting broadcastContext is nil")
+	}
+
+	alias := struct {
+		StateWithTimeoutFields stateWithTimeoutFields         `json:",inline"`
+		Substate               hardwareRestartingStateWrapper `json:"substate"`
+	}{StateWithTimeoutFields: stateWithTimeoutFields{broadcastContext: s.broadcastContext}}
+
+	if err := json.Unmarshal(data, &alias); err != nil {
+		return err
```
ao-david, line 105: wrapping this could be good.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+	}
+
+	s.stateWithTimeoutFields = alias.StateWithTimeoutFields
+
+	// Deserialize substate
+	if alias.Substate.Type != "" {
```
ao-david, line 111: again, should we consider the opposite an error?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+		Substate               hardwareRestartingStateWrapper `json:"substate"`
+	}{StateWithTimeoutFields: stateWithTimeoutFields{broadcastContext: s.broadcastContext}}
+
+	if err := json.Unmarshal(data, &alias); err != nil {
+		return err
+	}
+
+	s.stateWithTimeoutFields = alias.StateWithTimeoutFields
+
+	// Deserialize substate
```
ao-david, line 110: fullstop

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
...
+	if alias.Substate.Type != "" {
+		substate, err := registry.Get(alias.Substate.Type, s.broadcastContext)
+		if err != nil {
+			return fmt.Errorf("could not get substate from registry: %w", err)
```
ao-david, line 114: Given that we generally check the error first anyway I don't know that it probably matters, but worth noting that the returned data will not be nil if we are just returning the error. Not sure what the typical way of handling that would be?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -348,30 +424,107 @@ func (s hardwareStopping) New(args ...interface{}) (any, error) {
...
 	return &hardwareStopping{newStateWithTimeoutFields(ctx), nil}
 }
+
+// For Marshaling/Unmarshaling
+type hardwareStoppingStateWrapper struct {
+	Type string          `json:"type"`
+	Data json.RawMessage `json:"data"`
+}
+
+func (s *hardwareStopping) MarshalJSON() ([]byte, error) {
```
ao-david, line 434: I imagine ditto to all previous comments

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -348,30 +424,107 @@ func (s hardwareStopping) New(args ...interface{}) (any, error) {
 func newHardwareStopping(ctx *broadcastContext) *hardwareStopping {
 	return &hardwareStopping{newStateWithTimeoutFields(ctx), nil}
 }
+
+// For Marshaling/Unmarshaling
```
ao-david, line 428: fullstop

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -53,36 +53,112 @@ func newHardwareRestarting(ctx *broadcastContext) *hardwareRestarting {
 	return &hardwareRestarting{newStateWithTimeoutFields(ctx), nil}
 }
 
+// For Marshaling/Unmarshaling
```
ao-david, line 56: Full stop.

cmd/oceantv/broadcast_hardware_machine_test.go
```diff
@@ -662,3 +664,40 @@ func TestHardwareStopAndRestart(t *testing.T) {
...
+	}
+
+	data, err := json.Marshal(original)
+	assert.NoError(t, err, "Failed to marshal JSON")
+
+	unmarshaled := newHardwareRestarting(minimalMockBroadcastContext(t))
+	err = json.Unmarshal(data, unmarshaled)
+	assert.NoError(t, err, "Failed to unmarshal JSON")
+
+	// Validate main struct fields
```
ao-david, line 688: fullstops on comments

ao-david (approved):
LGTM, just a few comments

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: No need IMO because it'll just panic, but I don't mind a panic because this should be implemented otherwise programmer error.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: I don't think we need to, because it gets set when nil anyways.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: It's so that the custom Marshal method on the substate in s.Substate is called.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: The error message on the upper level should point you in the right direction TBH. As long as the error message is differentiable from other error messages in this scope, it's fine.

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Yeah agreed

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: ditto

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: I think in most cases if unmarshaling fails, and you don't have specific errors to handle you'd probably just ignore whatever has loaded. It'd be dependent on whether the method specifies otherwise though.

PR #499, opened 2025-02-18
oceantv: extra logs for hardwareStopping sub-state machine

trekhopton:
Logs were added to the broadcast machine to assist with debugging.

Note: I removed unrelated timezone changes from this PR.


ao-david:
I'm not sure I know what the issue that this is fixing is

trekhopton:
> I'm not sure I know what the issue that this is fixing is? Is there some way to replicate the bug or an issue written up for it?

Hmm, now I'm not even sure what I was fixing. 

It seems the timezone was being retrieved from the client timezone. I just thought it was supposed to be from the site timezone, which wasn't happening so I changed it. Althought in retrospect, I think it makes more sense to use the client timezone. That way we don't have to keep updating it... maybe.

But there was some sort of issue that made it seem like the start stop timing was off, that prompted me to look at this but now I'm not sure.

trekhopton:
I removed the timezone changes that were probably unnecessary. Do we think these logs will help? I think the more the better until we can understand what's happening consistently.

ao-david (approved):

ao-david:
> I removed the timezone changes that were probably unnecessary. Do we think these logs will help? I think the more the better until we can understand what's happening consistently.

I am a little worried that we may end up with too many logs that hide what we are looking for, but I think eventually when we have some amount of logging/state updates on the broadcast page, that this becomes less of a concern

PR #506, opened 2025-02-20
ausoceantv: login link under sign up

trekhopton:
This was done so that existing users don't feel like they have to sign up again.

image

### DIFF
--- a/cmd/ausoceantv/webapp/index.html
+++ b/cmd/ausoceantv/webapp/index.html
@@ -44,6 +44,12 @@
           <a href="/api/v1/auth/login?redirect=/home.html">
             <button class="rounded-lg bg-[#0c69ad] px-6 py-3 text-lg font-semibold text-white shadow-md transition-all hover:shadow-lg hover:brightness-110">Sign Up</button>
           </a>
+          <p class="mt-4 text-lg text-gray-800">
+            Already signed up? 
+            <a href="/api/v1/auth/login?redirect=/home.html" class="text-[#0c69ad] hover:underline">
+              Log in
+            </a>
+          </p>
         </div>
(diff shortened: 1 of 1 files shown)


scruzin (approved):
This is okay as a temporary workaround. Ideally, though, we should detect if the user is already signed up, so we don't present them with an unnecessary choice.

trekhopton:
> This is okay as a temporary workaround. Ideally, though, we should detect if the user is already signed up, so we don't present them with an unnecessary choice.

Yeah it was done this was to be quick. I think we should be able to redirect them straight to the home page if they have a valid Auth token, but otherwise I don't think we can detect that they have an account with them inputting some information.
I was thinking we could have the standard "sign in with google" button since "sign in" works in both situations, but I didn't quite get there with it last night.

ao-david:
I also think that we can remove the approvalForce option during signup. This will still prompt users to approve on their first signup, but any time after that they will just need to select their account, which further reduces friction.

We can also still store a cookie with a users email to use to hint google, even if they don't have a valid session.

trekhopton:
> I also think that we can remove the approvalForce option during signup. This will still prompt users to approve on their first signup, but any time after that they will just need to select their account, which further reduces friction.
> 
> We can also still store a cookie with a users email to use to hint google, even if they don't have a valid session.

Yes, absolutely, can you make the former a high priority issue and make an issue for the latter too?

ao-david:
I've though a little more about the second one with the cookies, and I'm not sure whether that's a good idea. If a user has logged out on a shared computer, I'm not sure we necessarily want to keep storing the data accessibly to other users?

I have made an issue for the first one though

PR #508, opened 2025-02-21
AOTV: Fix parsing and error handling for user survey

ao-david:
This change checks for empty responses in user survey submissions to avoid json parsing errors, as well as handling error responses better so that users can see why the error occured.


ao-david:
Currently the time restriction for users to complete the survey has been extended until 1 week to allow time for this bug fix

trekhopton (approved):
Great, thank you

Issue #315, opened 2024-09-25
Saving broadcast config during broadcast seems to clear live state


saxon-milton:
It was observed that if during the directLive state the end time is modified on the broadcast UI, the broadcast state somehow gets cleared to directIdle. This meant that it was no longer checking if the broadcast should be ended and as a result streamed over night.

scruzin:
@saxon-milton, is this still an issue?

saxon-milton:
I think this is still an issue. I believe this is a bug in the UI.

saxon-milton:
> I think this is still an issue. I believe this is a bug in the UI.

Bug in the UI, but I think the fact that the UI interfaces directly with broadcast configs is an issue. There should probably be an intermediate layer.

saxon-milton:
Closed by #509

PR #509, opened 2025-02-21
OTV: Use better update function on initial broadcast save (#315)

ao-david:
This change ensures that only fields that are actually set by the broadcasting UI are actually saved to the config. This should prevent state data from being overwritten.

### DIFF
--- a/cmd/oceantv/main.go
+++ b/cmd/oceantv/main.go
@@ -46,7 +46,7 @@ import (
 
 const (
 	projectID          = "oceantv"
-	version            = "v0.7.10"
+	version            = "v0.8.1"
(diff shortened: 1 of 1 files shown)


saxon-milton (approved):
Good one

saxon-milton:
> > Good one
> 
> I'm a little hesitant to merge this before we have any way of resetting the state, because I worry that we may end up in a bad state that we can't get out of?
> 
> Is this a reasonable concern?

The way that I have dealt with this in the past is by modifying the datastore entity directly in the google cloud console. This is fine for the occasional issue, but if we expected this to happen more often, we could consider adding a "reset" button.

saxon-milton:
Closes #315

Issue #510, opened 2025-02-27
AOTV: Review oauth2 parameters


ao-david:
Currently we are using approval force, which requires the 2-step profile, then permissions signup flow. We should be able to not use this and therefore users will only need to do this on their first use of the service.

PR #513, opened 2025-02-27
Gauth: remove ApprovalForce option (#510)

ao-david:
This change removes the approval force option which forces the user to consent to all scopes every time they log in. This reduces the friction for a user to login.


trekhopton (approved):
Tested it locally, works. Nice.

PR #516, opened 2025-03-04
Gauth: Add support for google login (onetap/signin with google)

ao-david:
This change adds support to the gauth library for the newer method of google signins. Either through the onetap method, or the signin with google button.

### DIFF
--- a/gauth/userauth.go
+++ b/gauth/userauth.go
@@ -218,7 +218,7 @@ func (ua *UserAuth) LoginHandler(h backend.Handler) error {
 	}
 
 	// NB: Offline access is required to obtain a refresh token.
-	url := ua.cfg.AuthCodeURL(sessID, oauth2.ApprovalForce, oauth2.AccessTypeOffline)
+	url := ua.cfg.AuthCodeURL(sessID, oauth2.AccessTypeOffline)
 	return h.Redirect(url, http.StatusFound)
 }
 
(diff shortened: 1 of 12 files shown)


gauth/userauth.go
```diff
@@ -380,39 +442,64 @@ func (ua *UserAuth) GetProfile(h backend.Handler) (*Profile, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("fetch profile error: %w", err)
-	}
 
-	profile.Data = data // Restore optional data.
-	err = sess.Set(oauthTokenSessionKey, newTok)
-	if err != nil {
-		return nil, fmt.Errorf("unable to set token session key: %w", err)
+	// Using Oauth2 Tokens.
+	if *tok != (oauth2.Token{}) {
```
saxon-milton, line 463: Comment says "Using Oauth2 Tokens" but we're checking here if we're not ?

saxon-milton: Also, do you need parentheses?

ao-david: There may be a better way to do this, but this is checking if the token is non-nil, ie we are actually using it. 
Is there a more idiomatic way of doing this.

PS. My IDE complains if I don't have parentheses on the open bracket of the if statement:
```syntax: expected ';', found '{'```

trekhopton (approved):

PR #517, opened 2025-03-04
Bench better broadcast testing

ao-david:
This change updates the cron page to make it much easier to set a different endpoint for an rpc cron.

To make the cron stay, the broadcast save method needs to be changed so that it checks before just blindly putting a cron.

image

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -73,7 +73,7 @@ import (
 )
 
 const (
-	version     = "v0.30.2"
+	version     = "v0.30.3"
 	localSite   = "localhost"
 	localDevice = "localdevice"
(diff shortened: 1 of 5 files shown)


cmd/oceanbench/main.go
```diff
@@ -73,7 +73,7 @@ import (
 )
 
 const (
-	version     = "v0.30.2"
+	version     = "v0.30.3"
```
saxon-milton, line 76: More like a minor version bump, because it's a newish feature

saxon-milton (approved):
Approved, besides version bump correction. Would also recommend getting second review from someone with more lit expertise.

trekhopton (approved):

PR #519, opened 2025-03-13
Model subfeed methods

ao-david:
Add methods and testing for subfeed methods.

### DIFF
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -1471,11 +1471,9 @@ func TestSubFeed(t *testing.T) {
 		t.Fatalf("could not get store: %v", err)
 	}
 
-	// Clear any existing subfeeds.
-	store.Delete(ctx, store.NameKey(typeSubFeed, fmt.Sprintf("%d.%d", testSubFeedFeedID, testSubFeedID)))
-	store.Delete(ctx, store.NameKey(typeSubFeed, fmt.Sprintf("%d.%d", testSubFeedFeedID, testSubFeedID+1)))
-
 	startTime := time.Now().UTC().Truncate(0)
+
+	// Add an arbitrary amount of time to differentiate start and finish.
 	finishTime := startTime.Add(1 * time.Hour)
 
 	subfeed := &SubFeed{
@@ -1544,6 +1542,14 @@ func TestSubFeed(t *testing.T) {
 	if subfeed4 != nil {
 		t.Errorf("expected nil, got %v", subfeed4)
 	}
+
+	// Cleanup.
+	t.Cleanup(func() {
+		err := os.RemoveAll("vidgrind")
+		if err != nil {
+			panic(err)
+		}
+	})
 }
 
 // Benchmarks follow.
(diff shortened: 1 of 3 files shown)


model/model_test.go
```diff
@@ -1458,6 +1458,94 @@ func testFeed(t *testing.T, kind string) {
...
+	)
+
+	ctx := context.Background()
+	store, err := datastore.NewStore(ctx, "file", "vidgrind", "")
+	if err != nil {
+		t.Fatalf("could not get store: %v", err)
+	}
+
+	// Clear any existing subfeeds.
+	store.Delete(ctx, store.NameKey(typeSubFeed, fmt.Sprintf("%d.%d", testSubFeedFeedID, testSubFeedID)))
```
saxon-milton, line 1475: I think you should consider using t.Cleanup for any sort of DB type testing, then you shouldn't need to do this. I also think it might be better to just do a check for any existing stuff, and panic if it's there; we don't expect anything in the file DB for the start of a test right ? Doing this hides any test cleanup issues.

model/model_test.go
```diff
@@ -1458,6 +1458,94 @@ func testFeed(t *testing.T, kind string) {
...
+	store.Delete(ctx, store.NameKey(typeSubFeed, fmt.Sprintf("%d.%d", testSubFeedFeedID, testSubFeedID+1)))
+
+	startTime := time.Now().UTC().Truncate(0)
+	finishTime := startTime.Add(1 * time.Hour)
```
saxon-milton, line 1479: Does 1 hour have any significance, or just arbitrary ?

saxon-milton (approved):

model/model_test.go
ao-david: This is just an arbitrary time so that they are different (not that it really matters)

model/model_test.go
ao-david: Would you recommend that the cleanup function deletes the directory directly instead of the entities

model/subfeed.go
```diff
@@ -0,0 +1,126 @@
+/*
+AUTHORS
+  David Sutton <davidsutton@ausocean.org>
+
+LICENSE
+  Copyright (C) 2025 the Australian Ocean Lab (AusOcean).
+
+  This is free software: you can redistribute it and/or modify it
```
trekhopton, line 8: Use the AusOceanTV comment header, see AusOceanTV's main.go.

trekhopton: Actually, this is in model, nevermind.

trekhopton (approved):

model/model_test.go
saxon-milton: Maybe comment that it's arbitrary

model/model_test.go
saxon-milton: Yes actually. I guess it should get rid of anything that wasn't there before the testing.

ao-david:
> version bump

We don't have a version number for model, the only version number that we have for it is the global cloud repo version tag, which I will update once this has been merged.

PR #520, opened 2025-03-18
Model: Add helper functions to feed

ao-david:
This change adds a couple of helper functions to the feed entity for accessing and editing the source and source type.

This will be useful when updating the source of a Feed on AOTV.

### DIFF
--- a/model/feed.go
+++ b/model/feed.go
@@ -24,6 +24,8 @@ package model
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


model/feed.go
```diff
@@ -114,8 +116,38 @@ func UpdateFeed(ctx context.Context, store datastore.Store, feed *Feed) (*Feed,
 	return updated, err
 }
 
+func SetFeedSource(ctx context.Context, store datastore.Store, fid int64, source string) error {
```
trekhopton, line 119: Document it.

trekhopton (approved):

PR #522, opened 2025-03-23
bench: only add broadcast cron when doesn't exist

saxon-milton:
This change makes sure we only create the broadcastCheck cron if it doesn't exist already. This ensures that if we've modified the broadcast check cron to hit the testing endpoint we don't overwrite that whenever we save the broadcast config.

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -390,16 +390,23 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 		msg = "broadcast saved successfully"
 
 		// Ensure that the CheckBroadcast cron exists.
-		c := &model.Cron{Skey: cfg.SKey, ID: "Broadcast Check", TOD: "* * * * *", Action: "rpc", Var: tvURL + "/checkbroadcasts", Enabled: true}
-		err = model.PutCron(context.Background(), settingsStore, c)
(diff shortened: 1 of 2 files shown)


trekhopton (approved):

cmd/oceanbench/broadcast.go
```diff
@@ -390,16 +390,23 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
...
-			reportError(w, r, req, "warning: failed to verify checkbroadcasts cron: %v", err)
-			return
-		}
+		const broadcastCheckCronID = "Broadcast Check"
+		_, err = model.GetCron(ctx, settingsStore, cfg.SKey, broadcastCheckCronID)
+		if errors.Is(err, datastore.ErrNoSuchEntity) {
+			c := &model.Cron{Skey: cfg.SKey, ID: broadcastCheckCronID, TOD: "* * * * *", Action: "rpc", Var: tvURL + "/checkbroadcasts", Enabled: true}
+			err = model.PutCron(context.Background(), settingsStore, c)
+			if err != nil {
+				reportError(w, r, req, "warning: failed to verify checkbroadcasts cron: %v", err)
```
ao-david, line 399: You could update this error since it is now failing to put the cron that is the actual error

ao-david (approved):

PR #523, opened 2025-03-26
cmd/ausoceantv: if there is no refresh token, prompt consent

trekhopton:
There's no way of getting another refresh token if it's gone for some reason (user clears cookies), so we need to check and prompt for consent if it's missing to get another refresh token. Unless we decide we want to store the refresh tokens in our DB.

### DIFF
--- a/gauth/userauth.go
+++ b/gauth/userauth.go
@@ -219,8 +219,22 @@ func (ua *UserAuth) LoginHandler(h backend.Handler) error {
 		return fmt.Errorf("could not save session %s: %w", sessID, err)
 	}
 
+	// Check for refresh token in the user's main session.
+	mainSession, _ := h.LoadSession(ua.SessionID)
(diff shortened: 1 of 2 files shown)


gauth/userauth.go
```diff
@@ -219,8 +219,22 @@ func (ua *UserAuth) LoginHandler(h backend.Handler) error {
 		return fmt.Errorf("could not save session %s: %w", sessID, err)
 	}
 
+	// Check for refresh token in the user's main session.
+	mainSession, _ := h.LoadSession(ua.SessionID)
```
ao-david, line 223: Is there a good reason we aren't capturing this error?

ao-david (approved):

gauth/userauth.go
trekhopton: No good reason, changed now.

PR #524, opened 2025-03-26
cmd/ausoceantv: add site notification for maintenance

trekhopton:
This was done as a quick fix to notify users of the site maintenance.

image


ao-david (approved):
I'm interested to see what this will become going forward, looks good!

PR #530, opened 2025-04-01
OTV: Don't notify with no shutdown actions

### LINKED ISSUE #529: OTV: Silence no shutdown actions error
At the moment if we have shutdown actions registered for a camera without syncreboot we will get a notification every time it tries to shutdown with a timeout error. If we remove the shutdown actions (since they aren't working) we will get a notification reporting that there are no shutdown actions supplied.

This is just adding extra notifications which mean that any notifications that are important are more likely to be missed.

ao-david:
This change stops a notification if the reason that the shutdown failed was due to there being no shutdown actions provided. This was done so that if cameras are not configured correctly to shutdown that we won't get notified every time that it turns off if we don't set any shutdown actions.

closes #529

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -562,7 +562,12 @@ func (s *hardwareStopping) handleTimeEvent(t timeEvent) {
 func (s *hardwareStopping) handleHardwareShutdownFailedEvent(event hardwareShutdownFailedEvent) {
 	switch s.Substate.(type) {
 	case *hardwareShuttingDown:
-		s.logAndNotify(broadcastHardware, "shutdown failed during hardware stop, skipping to power off: %v", event.Error())
+		// Don't notify if there were no shutdown actions.
+		if errors.Is(event, errNoShutdownActions) {
+			s.log("no shutdowns to perform, transitioning to powering off")
+		} else {
+			s.logAndNotify(broadcastHardware, "shutdown failed during hardware stop, skipping to power off: %v", event.Error())
+		}
 		s.transition()
 	default:
 		// Ignore.
--- a/cmd/oceantv/main.go
+++ b/cmd/oceantv/main.go
@@ -46,7 +46,7 @@ import (
 
 const (
 	projectID          = "oceantv"
-	version            = "v0.8.5"
+	version            = "v0.8.6"
 	projectURL         = "https://oceantv.appspot.com"
 	cronServiceAccount = "oceancron@appspot.gserviceaccount.com"
 	locationID         = "Australia/Adelaide" // TODO: Use site location.


trekhopton (approved):

saxon-milton (changes requested):
Like I said, I think notification on configuration issues is absolutely fine. We need to address these sorts of problems quick, especially for things that have potential to damage devices.

trekhopton:
> Like I said, I think notification on configuration issues is absolutely fine. We need to address these sorts of problems quick, especially for things that have potential to damage devices.

I think a warning upon configuration is more fitting, if a user doesn't want shutdown actions and they just want to power off the device, then they should be able to leave the field empty purposefully and not have to type "nil" or something.
We have mentioned before how we think email notifications should be reserved for high priority things, otherwise they are just spammy.
I think this change is okay.
It would be good if we had log levels.. this should be a warning IMO.

saxon-milton:
> > Like I said, I think notification on configuration issues is absolutely fine. We need to address these sorts of problems quick, especially for things that have potential to damage devices.
> 
> I think a warning upon configuration is more fitting, if a user doesn't want shutdown actions and they just want to power off the device, then they should be able to leave the field empty purposefully and not have to type "nil" or something. We have mentioned before how we think email notifications should be reserved for high priority things, otherwise they are just spammy. I think this change is okay. It would be good if we had log levels.. this should be a warning IMO.

I've put some comments on the issue too if you're interested.

From the OTV standpoint I would prefer that the configuration contains `None` or something along those lines so that it's indicated that it's at least been thought about. With an empty field it's not clear that it's been thought about as far as OTV is concerned. The front end can certainly just use a warning if you'd like, but I'd request that it sets the field to `None` when nothing is entered. It doesn't necessarily need to show this, but I think it'd be nice if this was the case behind the scenes.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -562,7 +562,15 @@ func (s *hardwareStopping) handleTimeEvent(t timeEvent) {
 func (s *hardwareStopping) handleHardwareShutdownFailedEvent(event hardwareShutdownFailedEvent) {
 	switch s.Substate.(type) {
 	case *hardwareShuttingDown:
-		s.logAndNotify(broadcastHardware, "shutdown failed during hardware stop, skipping to power off: %v", event.Error())
+		// We want to get notified for failures, warned for unconfigured broadcasts,
+		// and do nothing if the actions were purposely set to skip.
+		if errors.Is(event, warnSkipShutdown) {
+			// Do nothing.
+		} else if errors.Is(event, errNoShutdownActions) {
```
saxon-milton, line 569: this is a notify right ?

saxon-milton (changes requested):

cmd/oceantv/broadcast_hardware_machine.go
trekhopton: Yeah, I think we said this would be a notify and the previous case would be a warning log (instead of nothing).

ao-david: done

trekhopton (approved):

PR #531, opened 2025-04-02
otv: check there's no YouTube broadcast if not enabled

### LINKED ISSUE #43: broadcast.go: the checkBroadcastsHandler should ensure disabled broadcasts are stopped
Currently in performChecksInternalThroughStateMachine if the broadcast is disabled, no checks are performed.

We have observed a case where broadcasts were started, then a new VidGrind version was pushed which disabled the broadcasts. Somehow the state machine and broadcasts were out of sync, the broadcasts were running but disabled. This shouldn’t happen and we should check to make sure that disabled broadcasts aren’t running.

In this case, all rigs ran their broadcasts overnight because they weren’t stopped.

saxon-milton:
Occasionally we were getting into a situation where a broadcast would be disabled but the YouTube broadcast object would still be broadcasting. This ensures this doesn't happen by checking that there are no live broadcast YouTube objects associated.

### DIFF
--- a/cmd/oceantv/system.go
+++ b/cmd/oceantv/system.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/ausocean/cloud/cmd/oceantv/broadcast"
 	"github.com/ausocean/cloud/notify"
 	"github.com/ausocean/cloud/utils"
 )
@@ -138,9 +139,9 @@ func newBroadcastSystem(ctx context.Context, store Store, cfg *BroadcastConfig, 
 // cancellation the last time we ticked, and then publish a time event
 // to advanced the state machines again.
 func (bs *broadcastSystem) tick() error {
-	// Don't do anything if not enabled.
+	// If not enabled make sure we're in the idle state and have no broadcasts still in progress.
 	if !bs.ctx.cfg.Enabled {
-		// Also make sure it's in the idle state when not enabled, so we're not starting, transitioning or active.
+		// Make sure it's in the idle state when not enabled, so we're not starting, transitioning or active.
 		try(
 			bs.ctx.man.Save(nil, func(_cfg *BroadcastConfig) {
 				_cfg.AttemptingToStart = false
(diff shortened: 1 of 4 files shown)


saxon-milton:
Resolves #43

cmd/oceantv/system.go
```diff
@@ -150,7 +150,25 @@ func (bs *broadcastSystem) tick() error {
...
 		)
-		bs.log("not enabled, not doing anything")
+
+		// If there's a broadcast ID, set to complete if live and then clear it.
+		if bs.ctx.cfg.ID != "" {
+			status, err := bs.ctx.svc.BroadcastStatus(context.Background(), bs.ctx.cfg.ID)
+			if err != nil {
+				bs.log("could not get broadcast status: %v", err)
+			} else {
+				if status == "live" {
```
ao-david, line 160: Is status "live" not a constant somewhere?

cmd/oceantv/system.go
```diff
@@ -150,7 +150,25 @@ func (bs *broadcastSystem) tick() error {
 			"could not update config with callback",
 			log.Printf,
 		)
-		bs.log("not enabled, not doing anything")
+
+		// If there's a broadcast ID, set to complete if live and then clear it.
+		if bs.ctx.cfg.ID != "" {
+			status, err := bs.ctx.svc.BroadcastStatus(context.Background(), bs.ctx.cfg.ID)
+			if err != nil {
+				bs.log("could not get broadcast status: %v", err)
```
ao-david, line 158: Do we want to be continuing here if we get an error? ie do we want to clear the ID from the config if this fails?

cmd/oceantv/system.go
```diff
@@ -150,7 +150,25 @@ func (bs *broadcastSystem) tick() error {
...
+					err = bs.ctx.svc.CompleteBroadcast(context.Background(), bs.ctx.cfg.ID)
+					if err != nil {
+						bs.ctx.logAndNotify(broadcastService, "could not complete broadcast, please check this manually: %v", err)
+					}
+				}
+			}
+			try(bs.ctx.man.Save(nil, func(_cfg *BroadcastConfig) { _cfg.ID = "" }), "could not clear broadcast ID", bs.log)
+		}
+
+		bs.log("broadcast not enabled, not doing anything")
```
ao-david, line 170: Sometimes we are doing something though, should this log be amended?

ao-david (approved):
Few minor comments, but LGTM.

also bump version

cmd/oceantv/system.go
saxon-milton: I guess we'll see under what conditions we get an error. I don't think getting an error here necessarily means the ID is invalid.

cmd/oceantv/system.go
saxon-milton: nup, but I can add it.

cmd/oceantv/system.go
saxon-milton: We just return nil here though, so not doing much of anything ?

PR #532, opened 2025-04-04
otv: enable state/event based webhooks

### LINKED ISSUE #526: OTV: State Machine webhook system
It would be helpful for the broadcast manager to have a webhook system for each of the states so that we can add functions when different states are reached.

For example, we could add a webhook for the directLive state, that triggers an endpoint such as /tvwebhook on bench with a payload like:
```json
{
  "id": "Rapid Bay Live by AusOcean",
  "state": "directLive",
}
```
This could then be used to trigger a variety of registered actions, such as registering the stream with AOTV.

saxon-milton:
This change provides the option to add event or state based webhooks (or any other sort of hook for that matter). This is is motivated by the desire for services such as AusOceanTV and OpenFish to have an awareness of broadcast events and states.

We have done this by using options to an oceantv service, which then propagates these hooks to the state machine.

closes #526

### DIFF
--- a/cmd/oceantv/broadcast.go
+++ b/cmd/oceantv/broadcast.go
@@ -168,9 +168,44 @@ func (c *BroadcastConfig) parseStartEnd() error {
 	return nil
 }
 
+type oceanTVService struct {
+	eventHooks []eventHook
+	stateHooks []stateHook
+}
+
+type oceanTVOption func(*oceanTVService) error
+
+type eventHook func(event, *Cfg)
+type stateHook func(state, *Cfg)
+
+func withEventHooks(hooks ...eventHook) oceanTVOption {
+	return func(s *oceanTVService) error {
+		s.eventHooks = hooks
+		return nil
+	}
+}
+
+func withStateHooks(hooks ...stateHook) oceanTVOption {
+	return func(s *oceanTVService) error {
+		s.stateHooks = hooks
+		return nil
+	}
+}
+
+func newOceanTVService(opts ...oceanTVOption) (*oceanTVService, error) {
+	otv := &oceanTVService{}
+	for i, opt := range opts {
+		err := opt(otv)
+		if err != nil {
+			return nil, fmt.Errorf("could not apply option (%d) to oceanTV service: %w", i, err)
+		}
+	}
+	return otv, nil
+}
+
 // checkBroadcastsHandler checks the broadcasts for a single site.
 // It is designed to be invoked via OceanCron rpc requests, not cron.yaml.
-func checkBroadcastsHandler(w http.ResponseWriter, r *http.Request) {
+func (s *oceanTVService) checkBroadcastsHandler(w http.ResponseWriter, r *http.Request) {
 	logRequest(r)
 
 	ctx := r.Context()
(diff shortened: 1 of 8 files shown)


saxon-milton:
Changing to draft while I fix the test issues....

saxon-milton:
OK good for review now...

cmd/oceantv/main.go
```diff
@@ -111,15 +111,112 @@ func main() {
...
+				}
+
+				data := struct {
+					ID    string `json:"id"`
+					State string `json:"state"`
+				}{
+					ID:    cfg.Name,
+					State: stateToString(s),
+				}
+				const ausoceanTVWebHookDest = "https://www.ausocean.org/ausocean-tv/tvwebhook"
```
ao-david, line 168: This doesn't seem like the right endpoint, was this just a placeholder? I think it should go in the const declarations at the top of the file.

At the moment, I don't know exactly what the endpoint would be, but I think it will be something like ```https://ausocean.tv/api/v1/tvwebhook

ao-david (approved):
Nice one!

cmd/oceantv/broadcast.go
```diff
@@ -168,9 +168,44 @@ func (c *BroadcastConfig) parseStartEnd() error {
 	return nil
 }
 
+type oceanTVService struct {
+	eventHooks []eventHook
+	stateHooks []stateHook
+}
+
+type oceanTvOption func(*oceanTVService) error
```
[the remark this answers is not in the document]
saxon-milton: Yer should be; I think it was a typo

cmd/oceantv/broadcast.go
```diff
@@ -168,9 +168,44 @@ func (c *BroadcastConfig) parseStartEnd() error {
...
+		return nil
+	}
+}
+
+func newOceanTVService(opts ...oceanTvOption) (*oceanTVService, error) {
+	otv := &oceanTVService{}
+	for _, opt := range opts {
+		err := opt(otv)
+		if err != nil {
+			return nil, fmt.Errorf("could not apply option to oceanTV service: %w", err)
```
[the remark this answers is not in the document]
saxon-milton: Yeah that seems like a fair idea

cmd/oceantv/main.go
saxon-milton: Just a place holder. 

I'm not sure it should be defined as a global const because this is the only thing that uses it and it's nice to keep everything needed by this hook right here so that it's not noise anywhere else.

cmd/oceantv/main.go
```diff
@@ -111,15 +111,112 @@ func main() {
...
+	if err != nil {
+		return fmt.Errorf("failed to send request: %v", err)
+	}
+	defer func(b io.ReadCloser) {
+		if err := b.Close(); err != nil {
+			log.Printf("failed to close webhook response body: %v", err)
+		}
+	}(resp.Body)
+
+	if resp.StatusCode == http.StatusOK {
```
scott97, line 213: Could we add http.StatusCreated and http.StatusAccepted as acceptable status codes too? Openfish uses `http 201 status created` for register streams?

cmd/oceantv/main.go
```diff
@@ -111,15 +111,112 @@ func main() {
...
+				if err != nil {
+					log.Printf("could not setup openfish service: %v", err)
+					return
+				}
+
+				err = ofsvc.RegisterStream(cfg.SID, cs, cfg.Start, cfg.End)
+				if err != nil {
+					log.Printf("could not register stream with OpenFish: %v", err)
+					return
+				}
```
scott97, line 143: In a future PR can we replace ofsvc.RegisterStream with sendWebhook? sendWebhook does everything we need I believe

scott97 (approved):
Looks great :+1: Left a couple of comments but they are not blockers to merging imo

cmd/oceantv/main.go
saxon-milton: No worries, can do!

cmd/oceantv/main.go
saxon-milton: Cool!

PR #545, opened 2025-04-21
otv: improve error event handling

saxon-milton:
The intention here is to improve the function of events that represent erroneous conditions and make error information more available and clearer. We're also trying to use more consistent patterns around error conditions, and do more central notification on these error events.

Events representing erroneous conditions are errors themselves. They also hold errors instead of basic strings. This means we can chain error events/errors and maintain a history of the errors that led us to where we are. As such we have unwrapping functionality to find the initial error event type. We've also added a Kind() method to the interface that gives us the corresponding notify.Kind so that we can perform a notification of the correct kind.

Finally we've fixed some fundemental issues around storing events for processing; we weren't properly storing the error events internal error information, and would therefore loose this on the next SM tick.

### DIFF
--- a/cmd/oceantv/broadcast_events.go
+++ b/cmd/oceantv/broadcast_events.go
@@ -1,15 +1,26 @@
 package main
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"context"
 
 	"github.com/ausocean/cloud/cmd/oceantv/registry"
+	"github.com/ausocean/cloud/notify"
 )
 
 type event interface{ fmt.Stringer }
+
+type errorEvent interface {
+	event
+	error
+	registry.Newable
+	Kind() notify.Kind
+}
 
 // registerEvent registers an event in the registry.
 // It returns a struct{} to allow us to register the event in a var declaration in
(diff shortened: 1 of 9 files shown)


cmd/oceantv/broadcast_events.go
```diff
@@ -52,18 +63,63 @@ var _ = registerEvent(startedEvent{})
...
 
-type startFailedEvent struct{}
+type startFailedEvent struct{ error }
 
 var _ = registerEvent(startFailedEvent{})
 
 func (e startFailedEvent) String() string { return "startFailedEvent" }
+func (e startFailedEvent) Error() string {
+	if e.error == nil {
+		return "(" + e.String() + ") <nil>"
```
trekhopton, line 73: Not sure I'm a fan of the <nil> in the error message. It's supposed to be helpful to a human right? This makes it look like a template wasn't filled out or something.

trekhopton (approved):

cmd/oceantv/broadcast_events.go
saxon-milton: This is common in the go idiom (at least in the std library) when you're formatting something and the parameters you're providing for formatting are nil; just following this pattern.

Issue #542, opened 2025-04-17
bench: TestGetScalars fails on main when run locally


trekhopton:
```
--- FAIL: TestGetScalars (1.23s)
    data_test.go:96: TestGetScalars
    data_test.go:112: Expected 86400 scalars, got 569
    data_test.go:119: #0: expected 512, got 355.000000
    data_test.go:119: #1: expected 513, got 354.000000
    data_test.go:119: #2: expected 514, got 353.000000
...
data_test.go:119: #565: expected 672, got 507.000000
    data_test.go:119: #566: expected 671, got 508.000000
    data_test.go:119: #567: expected 670, got 509.000000
    data_test.go:119: #568: expected 669, got 510.000000
panic: runtime error: index out of range [569] with length 569 [recovered]
        panic: runtime error: index out of range [569] with length 569

goroutine 59 [running]:
testing.tRunner.func1.2({0xd5cbc0, 0xc00003fb90})
        /home/trek/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.linux-amd64/src/testing/testing.go:1632 +0x230
testing.tRunner.func1()
        /home/trek/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.linux-amd64/src/testing/testing.go:1635 +0x35e
panic({0xd5cbc0?, 0xc00003fb90?})
        /home/trek/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.linux-amd64/src/runtime/panic.go:785 +0x132
github.com/ausocean/cloud/cmd/oceanbench.TestGetScalars(0xc0004d61a0)
        /home/trek/go/src/github.com/ausocean/cloud/cmd/oceanbench/data_test.go:118 +0x3ed
testing.tRunner(0xc0004d61a0, 0xe24f78)
        /home/trek/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.linux-amd64/src/testing/testing.go:1690 +0xf4
created by testing.(*T).Run in goroutine 1
        /home/trek/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.linux-amd64/src/testing/testing.go:1743 +0x390
FAIL    github.com/ausocean/cloud/cmd/oceanbench        116.735s
```

Sometimes it takes a crazy amount of time to test too. 473s.

trekhopton:
I re-tested on the STARLINK network and it passed.

```
go test ./... --count=1
[...]

PR #546, opened 2025-04-23
bench: fix TestGetScalars panic

trekhopton:
This was done so that the test wouldn't panic when it fails. It's still unclear why it failed and is now passing.

Fixes the panic in issue #542


scruzin (approved):
It would be good to understand the root cause of the failure, though.

trekhopton:
> It would be good to understand the root cause of the failure, though.

Yeah, that would be good. I'm actually running into it again so might do more more investigation.

PR #549, opened 2025-04-28
otv & bench: add notification suppression rules

saxon-milton:
This allows us to add notification rules through the broadcast page. This means we can suppress notifications of a particular kind or containing a particular string. The rules are in a JSON format: { "SuppressKinds" : ["kind1","kind2", ... ], "SuppressContaining": ["some str"] }

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -158,6 +158,7 @@ type BroadcastConfig struct {
 	VoltageRecoveryTimeout   int           // Max allowable hours for voltage recovery before failure.
 	RegisterOpenFish         bool          // True if the video should be registered with openfish for annotation.
 	OpenFishCaptureSource    string        // The capture source to register the stream to.
+	NotifySuppressRules      string        // Suppression rules for notifications.
 }
 
 func (b *BroadcastConfig) PrettyHardwareStateData() string {
@@ -230,6 +231,7 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 			RegisterOpenFish:      r.FormValue("register-openfish") == "register-openfish",
 			OpenFishCaptureSource: r.FormValue("openfish-capturesource"),
 			BatteryVoltagePin:     r.FormValue("battery-voltage-pin"),
+			NotifySuppressRules:   r.FormValue("notify-suppress-rules"),
 		},
 		Action:             r.FormValue("action"),
 		ListingSecondaries: r.FormValue("list-secondaries") == "listing-secondaries",
--- a/cmd/oceantv/system.go
+++ b/cmd/oceantv/system.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/ausocean/cloud/cmd/oceantv/broadcast"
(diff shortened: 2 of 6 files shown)


cmd/oceanbench/broadcast.go
```diff
@@ -158,6 +158,7 @@ type BroadcastConfig struct {
 	VoltageRecoveryTimeout   int           // Max allowable hours for voltage recovery before failure.
 	RegisterOpenFish         bool          // True if the video should be registered with openfish for annotation.
 	OpenFishCaptureSource    string        // The capture source to register the stream to.
+	NotifySuppressRules      string        // Suppression rules for notifications.
```
ao-david, line 161: I think that you will need to update the save endpoint so that this doesn't get overwritten?

cmd/oceantv/system.go
```diff
@@ -140,10 +142,49 @@ func newBroadcastSystem(ctx context.Context, store Store, cfg *BroadcastConfig,
...
+		//  "SuppressKinds": ["broadcast-kind1" , "broadcast-kind2"],
+		// 	"SuppressContaining": ["shutdown failed", "failed to start"]
+		// }
+		suppressionRules := &struct {
+			SuppressKinds      []string
+			SuppressContaining []string
+		}{}
+
+		// Completely empty string indicates no suppression rules.
+		if cfg.NotifySuppressRules != "" {
```
ao-david, line 165: I think that this could be inverted:
```golang
if cfg.NotifySuppressRules == "" {
  broadcastContext.logAndNotify(errEvent.Kind(), "error event: %s", errEvent.Error())
  return nil
}
err := json.Unmarshal([]byte(cfg.NotifySuppressRules), suppressionRules)
if err != nil {
  broadcastContext.logAndNotify(errEvent.Kind(), "could not unmarshal notification suppression rules: %v", err)
  return nil
}
```
Although it does mean repeating logAndNotify() so I dunno

cmd/oceantv/system.go
```diff
@@ -140,10 +142,49 @@ func newBroadcastSystem(ctx context.Context, store Store, cfg *BroadcastConfig,
...
+		if _, ok := event.(errorEvent); !ok {
+			return nil
 		}
+
+		errEvent := event.(errorEvent)
+
+		// Unmarshal the notification suppression rules from the broadcast configuration.
+		// It is of format:
+		// {
+		//  "SuppressKinds": ["broadcast-kind1" , "broadcast-kind2"],
```
ao-david, line 156: Are the kinds notify kinds or broadcast kinds? It makes it seem like there are different kinds of broadcast which you can suppress, but as I understand it, it is notify kinds? To me it would make more sense if this hint was ```"notify-kind1"```

ao-david (approved):
LGTM. Are you planning on adding a more user-friendly UI for this, or do you want to write an issue for this, and get someone else to tackle?

Also, I wonder whether it would be possible/useful to have site wide supression rules (although I think that this should be handled by bench rather than oceantv)

cmd/oceanbench/broadcast.go
saxon-milton: True; i think we need a more consolidated way to deal with this.

cmd/oceantv/system.go
saxon-milton: Probably not worth it IMO, it's only a small block and small amount of nesting.

cmd/oceantv/system.go
saxon-milton: Not sure it matters; it's just an example comment. I'm just trying to be consistent with the broadcast kinds.

saxon-milton:
> LGTM. Are you planning on adding a more user-friendly UI for this, or do you want to write an issue for this, and get someone else to tackle?
> 
> Also, I wonder whether it would be possible/useful to have site wide supression rules (although I think that this should be handled by bench rather than oceantv)

Could add a more user friendly UI for this, but also not sure how much it adds given that it's just an advanced admin feature that will hopefully be updated only rarely. Personally wouldn't even add the noise of an issue given it's low priority and triviality, but I could be convinced otherwise.

ao-david:
> > LGTM. Are you planning on adding a more user-friendly UI for this, or do you want to write an issue for this, and get someone else to tackle?
> > Also, I wonder whether it would be possible/useful to have site wide supression rules (although I think that this should be handled by bench rather than oceantv)
> 
> Could add a more user friendly UI for this, but also not sure how much it adds given that it's just an advanced admin feature that will hopefully be updated only rarely. Personally wouldn't even add the noise of an issue given it's low priority and triviality, but I could be convinced otherwise.

I think that we should see how we end up using it and if it becomes necessary I guess

PR #551, opened 2025-04-29
OTV: Handle hardwarePowerOffFailedEvents

### LINKED ISSUE #550: OTV: Handle HardwarePowerOffFailedEvent
Currently, the hardware state machine doesn't handle HardwarePowerOffFailedEvents, this can cause looping of timeouts and publishing failed events, without achieving anything. I think that we should probably keep trying to turn off the camera, as we don't want to waster power, but these failures shouldn't happen silently. 

I am proposing that it notifies after a certain number of failures, and also puts the broadcast into failure mode

ao-david:
This change handles the event that a camera fails to turn off (after a timeout)
by notifying the ops email, and transitioning into hardware failure.

This change also adds unit tests and helpers for this use case.

There is possibly an argument for not transitioning to hardwareFailure (through publishing a hardwareStopFailedEvent), and this can be done, but we need to add a new state type which tracks whether a notification has been sent, otherwise without transitioning, we would likely receive a notification on every timeout period. However, given that hardwareFailure does not change the variables at all, if the camera was setup to power off correctly, transitioning to hardwareFailure should not prevent this anyway.

This notification should probably be treated as high-priority, given that it could likely mean that a camera is sucking power overnight, and so the shutdown behaviour of the camera after the notification shouldn't matter as much.

closes #550

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -359,6 +359,12 @@ func (e hardwareShutdownFailedEvent) New(args ...any) (any, error) {
 	}
 	return hardwareShutdownFailedEvent{err}, nil
 }
+func (e hardwareShutdownFailedEvent) Is(target error) bool {
+	if targetEvent, ok := target.(hardwareShutdownFailedEvent); ok {
+		return errors.Is(e.error, targetEvent.error)
+	}
+	return errors.Is(e.error, target)
+}
 
 // Kind implements the errorEvent interface.
 func (e hardwareShutdownFailedEvent) Kind() notify.Kind {
@@ -616,7 +622,7 @@ func (s *hardwareStopping) handleHardwareShutdownFailedEvent(event hardwareShutd
 		// We want to get notified for failures and misconfigured configs, and log
 		// when shutdown is skipped.
 		if errors.Is(event, warnSkipShutdown) {
-			s.log("skipping shutdown: %v:", event.Error)
+			s.log("skipping shutdown: %v", event.Error())
 		} else if errors.Is(event, errNoShutdownActions) {
 			s.logAndNotify(broadcastHardware, "shutdown skipped: %v", event.Error())
 		} else {
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -634,6 +634,16 @@ func (s *hardwareStopping) handleHardwareShutdownFailedEvent(event hardwareShutd
 	}
 }
 
+func (s *hardwareStopping) handleHardwarePowerOffFailedEvent(event hardwarePowerOffFailedEvent) {
+	switch s.Substate.(type) {
+	case *hardwarePoweringOff:
+		s.logAndNotify(broadcastHardware, "power off failed during hardware stop: %v", event.Error())
+		s.bus.publish(hardwareStopFailedEvent{})
+	default:
+		// Ignore.
+	}
+}
+
 func (s *hardwareStopping) cameraIsReporting() bool {
 	up, err := s.camera.isUp(s.broadcastContext, model.MacDecode(s.cfg.CameraMac))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


trekhopton (approved):

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -628,6 +634,16 @@ func (s *hardwareStopping) handleHardwareShutdownFailedEvent(event hardwareShutd
 	}
 }
 
+func (s *hardwareStopping) handleHardwarePowerOffFailedEvent(event hardwarePowerOffFailedEvent) {
+	switch s.Substate.(type) {
+	case *hardwarePoweringOff:
+		s.logAndNotify(broadcastHardware, "power off failed during hardware stop: %v", event.Error())
```
saxon-milton, line 640: This would be doubling up on notification; we now have a an error event notifier subscribed to the event bus i.e. a centralised place where error event notifications happen, so we don't have to have notifications calls scattered throughout the code.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -359,6 +359,12 @@ func (e hardwareShutdownFailedEvent) New(args ...any) (any, error) {
 	}
 	return hardwareShutdownFailedEvent{err}, nil
 }
+func (e hardwareShutdownFailedEvent) Is(target error) bool {
```
saxon-milton, line 362: So this works ? Do we have to do any unwrapping for e.error below, or does this happen automatically ?

saxon-milton (changes requested):

saxon-milton:
> This change handles the event that a camera fails to turn off (after a timeout) by notifying the ops email, and transitioning into hardware failure.
> 
> This change also adds unit tests and helpers for this use case.
> 
> There is possibly an argument for not transitioning to hardwareFailure (through publishing a hardwareStopFailedEvent), and this can be done, but we need to add a new state type which tracks whether a notification has been sent, otherwise without transitioning, we would likely receive a notification on every timeout period. However, given that hardwareFailure does not change the variables at all, if the camera was setup to power off correctly, transitioning to hardwareFailure should not prevent this anyway.
> 
> This notification should probably be treated as high-priority, given that it could likely mean that a camera is sucking power overnight, and so the shutdown behaviour of the camera after the notification shouldn't matter as much.
> 
> closes #550

Re the argument against hardwareFailure; i think we need another implementation of the hardwareManager that's for an "always on" camera. So then we can modify how we check for successful shutdown and power down. Maybe the interface can have a isOff interface or something, or is shutdownComplete, or isPowerOffComplete and then they can just return true. Maybe needs some thinking.

cmd/oceantv/broadcast_hardware_machine.go
ao-david: Does this mean that if we are publishing an event we should only log, not log and notify? If we aren't publishing an event do we still need to log and notify if we are handling an event (ie :621 where we handle the hardwareShutdownFailedEvent)?

cmd/oceantv/broadcast_hardware_machine.go
ao-david: It works in the use case that we currently are using it for (ie the error is passed directly in an event), but I will test further to confirm whether it works generally

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Neither, the error event handler will do both log and notify. Probably OK to keep handling logging.

ao-david: I still don't quite understand what the requirement of the event handler is?

I think there are two situations, 
1. an event handler receives an event and doesn't fire a new event. 
2. an event handler receives an event and fires a new event.

In both 1 and 2, what should the handler do? log AND notify, or just log, or do nothing?

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: Need a rebase ? this is already on master ?

cmd/oceantv/broadcast_hardware_machine.go
saxon-milton: There's not really requirements per se. And those aren't the only two things an event handler can do. But in the spirit of the question, in case 1, if you're not emitting another event and not doing anything else, then it probably doesn't need to exist. In case 2, you probably don't need to do any logging or notification. ( this applies to error events, which have the centralised logging and notification, you might choose to handle other events differently if it's useful to know something about these )

ao-david:
@saxon-milton can you confirm that this should be closed in preference of #568

PR #552, opened 2025-04-29
gauth: pass on refresh token when getting new access token

trekhopton:
This was done so that we don't lose the refresh token. I think this has been happening.


trekhopton:
This is unnecessary, the refresh token gets passed on without this.

PR #553, opened 2025-04-29
model: don't delete the test device mac=1

trekhopton:
The test device was being deleted causing issues for our test code that uses this device for testing purposes. 
Also made an error fatal to prevent panics in bench.
Updated utils for security.


ao-david (approved):
I think it would be worth writing up an issue regarding refactoring tests at some point. I think that we have been talking about it for a while, and are starting to run into some issues with it, and it would be good to start capturing the specific issues so that when we finally get around to refactoring (if we ever do) we know what things we would like to change.

PR #554, opened 2025-04-29
Model: Add last seen field to subscriber

ao-david:
This change adds a last seen field to the subscriber which tracks the last time that the profile made a request (not just signed in)

### DIFF
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -1350,8 +1350,9 @@ func testSubscriber(t *testing.T, kind string) {
 	sub2Key := fmt.Sprintf("%d.%s", sub2ID, sub2Email)
 
 	// Create test subscriber objects.
-	s1a := &Subscriber{sub1ID, "", sub1Email, "first", "last", nil, "", "", time.Now().Round(time.Second).UTC()}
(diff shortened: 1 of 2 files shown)


model/subscriber.go
```diff
@@ -88,6 +89,7 @@ func CreateSubscriber(ctx context.Context, store datastore.Store, s *Subscriber)
 	}
 
 	s.Created = time.Now()
+	s.LastSeen = time.Now()
```
trekhopton, line 92: So you're only updating it in CreateSubscriber. Is the per request update coming in a future PR?

model/model_test.go
```diff
@@ -1429,6 +1430,21 @@ func testSubscriber(t *testing.T, kind string) {
 	if !found2 {
 		t.Errorf("GetAllSubscribers did not return the second subscriber: %+v", s2)
 	}
+
+	// Update the last seen time for the first subscriber
```
trekhopton, line 1434: Fullstops.

trekhopton:
We should consider updating the subscriber last seen in the lite version in cloud too, if it's easy enough.

trekhopton (approved):

ao-david:
As per @scruzin's request, we will close this PR in preference to a subscriber activity log as a separate entity type.

PR #555, opened 2025-04-30
otv: fix event marshalling/unmarshalling and error matching

saxon-milton:
Events are complex in that they can be of errorEvent type which means that they can chain other errorEvents or normal errors.

This change fixes error matching using Is and marshalling/unmarshalling so that it happen recursively and reconstruct the events exactly as they were. We've also added testing to verify things are working as expected.

### DIFF
--- a/cmd/oceantv/broadcast_events_test.go
+++ b/cmd/oceantv/broadcast_events_test.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"errors"
+	"fmt"
 	"reflect"
 	"sync"
 	"testing"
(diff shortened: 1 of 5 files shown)


cmd/oceantv/broadcast_events_test.go
```diff
@@ -152,3 +154,152 @@ func TestStringToEvent(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			data := marshalEvent(tc.input)
+			unmarshalled := unmarshalEvent(data)
+
+			if _, ok := tc.input.(error); ok && tc.target != nil {
+				// Check that Error() returns the same string
```
trekhopton, line 286: Fullstops.

trekhopton (approved):

PR #560, opened 2025-05-05
otv: handle no voltage value when ESP off

saxon-milton:
When we have a low voltage case, the ESP goes into alarm mode, but then there will be no voltage sensor value. So if we get a no such entity error, we assume the voltage is the alarm voltage.

### DIFF
--- a/cmd/oceantv/broadcast_hardware_machine.go
+++ b/cmd/oceantv/broadcast_hardware_machine.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ausocean/cloud/cmd/oceantv/registry"
 	"github.com/ausocean/cloud/model"
 	"github.com/ausocean/cloud/notify"
+	"github.com/ausocean/openfish/datastore"
 )
 
 func register(state registry.Named) struct{} {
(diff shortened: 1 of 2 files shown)


cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -1008,9 +1009,19 @@ func (c *revidCameraClient) voltage(ctx *broadcastContext) (float64, error) {
 
 	// Get current battery voltage.
 	voltage, err := model.GetSensorValue(context.Background(), ctx.store, sensor)
-	if err != nil {
+	switch {
+	case errors.Is(err, datastore.ErrNoSuchEntity):
```
ao-david, line 1013: Is there any validation prior to this point that ensures that the sensor actually exists? 

Whilst I think that most cases of NoSuchEntity are likely non-reporting errors, I worry that this could still hide some configuration errors?

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/ausocean/cloud/cmd/oceantv/registry"
 	"github.com/ausocean/cloud/model"
 	"github.com/ausocean/cloud/notify"
+	"google.golang.org/appengine/datastore"
```
ao-david, line 14: I think that we should be using the openfish/datastore implementation of ErrNoSuchEntity. Otherwise this is a bad abstraction

ao-david (changes requested):

cmd/oceantv/broadcast_hardware_machine.go
ao-david: gh just cut it off aha

ao-david (approved):
LGTM

PR #563, opened 2025-05-14
Poll requests now update variables supplied by the client

scruzin:
And var requests no longer do so.
Implements Issue #562.


trekhopton (approved):

trekhopton:
Actually, not sure if this will work. See email.

PR #569, opened 2025-05-17
otv: improve directLiveUnhealthy fix code

saxon-milton:
* In the case that there's a fix failure, we're now publishing a fixFailureEvent.
* fixFailureEvent now implements the errorEvent interface.
* We've fix the attempt count increment logic.
* Correcting possition of LastResetAttempt update.


trekhopton:
A more high level reason to why this is needed would be useful, or an issue link. 

Is it mainly clearer error reporting and increasing attempts, motivated by silent failures of fix attempts?

trekhopton (approved):

saxon-milton:
> A more high level reason to why this is needed would be useful, or an issue link.
> 
> Is it mainly clearer error reporting and increasing attempts, motivated by silent failures of fix attempts?

Yeah OK, I'll improve the commit description.

PR #575, opened 2025-05-27
System: Define the default value for camera variables (#195)

ao-david:
This change defines a set of default values to be used for camera variables.

These variables were taken from the most recently configured camera (Port Noarlunga Reef Cam)

### DIFF
--- a/system/camera/defaults.go
+++ b/system/camera/defaults.go
@@ -1,0 +1,41 @@
+/*
+AUTHORS
+  David Sutton <david@ausocean.org>
+
+LICENSE
+  Copyright (C) 2025 the Australian Ocean Lab (AusOcean).
+
+  This is free software: you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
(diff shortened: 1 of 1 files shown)


scruzin (changes requested):
There are other devices, such as speakers that will also have defaults, and some of the names will be identical, such as DefaultMode.

Therefore, I think there should be different name spaces for each type of device.
For example, the camera defaults would live under system/camera, the speaker defaults, under system/speaker, etc.

system/defaults.go
```diff
@@ -0,0 +1,40 @@
...
+	DefaultHDR              = "auto"
+	DefaultHeight           = 1080
+	DefaultInput            = "rpicam"
+	DefaultOutput           = "RTMP"
+	DefaultRotation         = 180
+	DefaultSaturation       = 1
+	DefaultSharpness        = 1
+	DefaultWidth            = 1920
+	DefaultLogging          = "Debug"
+	DefaultMode             = ""
```
scruzin, line 39: For clarity, while the empty string is valid, `DefaultMode` should be **Normal** not an empty string.

trekhopton (approved):

PR #579, opened 2025-05-29
Add controller error state handling

saxon-milton:
We can check the controller error variable first to easily get any error state. If the controller does not yet support this, we still check other pieces of information.


cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -234,6 +234,35 @@ func (s *hardwareStarting) enter() {
...
+	}
+
+	switch {
+	case errors.Is(hwErr, LowVoltageAlarm):
+		s.log("controller voltage is low, waiting for recovery before starting")
+		s.bus.publish(lowVoltageEvent{})
+		return
+	case errors.Is(hwErr, None):
+		// Continue other checks, this is good.
+	case hwErr != nil:
```
scruzin, line 256: This just happens when the the device error is something we don't expect, i.e.., neither the empty string nor LowVoltage, right?

scruzin (approved):
Just one question, but LGTM.

cmd/oceantv/broadcast_hardware_machine.go
Saxon1: Yep that's right; perhaps `invalidConfigurationEvent` is not quite right for this, it's more a like a `softwareErrorEvent`, but we don't have anything like this yet.

cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -234,6 +234,35 @@ func (s *hardwareStarting) enter() {
...
+		errWrapped := fmt.Errorf("unhandled controller hardware error: %w", hwErr)
+		s.log(errWrapped.Error())
+		s.bus.publish(invalidConfigurationEvent{errWrapped})
+		return
+	default:
+		// This means heErr get failed to get, which at this stage just means
```
ao-david, line 262: this sentence needs some work I think.
at least:
```heErr -> hwErr```

but ```hwErr get failed to get``` doesn't quite sound right to me, maybe ```failed to get hwErr```?

ao-david: Does it only mean that the controller doesn't have the latest firmware, or could something else have happened?

ao-david (approved):
Few small comments about a comment

PR #584, opened 2025-06-10
Bench: Add reset button (#582)

### LINKED ISSUE #582: OTV: Reset hardwareFailure state in broadcast via UI
It seems that the HSM gets stuck in HardwareFailure state and toggling "failure mode" in the UI to off doesn't solve the issue. 

This is an important issue because it means when a broadcast failure due to a hardware failure, someone has to go into the datastore to manually edit the broadcast.. not good.

We just get an error when the broadcast tries to restart: unexpected event startBroadcastRequestEvent in state hardwareFailure:

Image

I propose a reset button that resets the BSM and HSM states so the broadcast can try again from a clean slate.

ao-david:
This change adds a reset state button to the broadcast UI which will allow the user to reset the state of a broadcast without breaking any other values.
closes #582 
image

### DIFF
--- a/cmd/oceanbench/broadcast.go
+++ b/cmd/oceanbench/broadcast.go
@@ -68,6 +68,7 @@ const (
 	broadcastToken
 	broadcastDelete
 	broadcastSelect
+	broadcastResetState
 
 	// Vidforward control API request actions.
 	vidforwardCreate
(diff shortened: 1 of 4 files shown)


saxon-milton:
> This change adds a reset state button to the broadcast UI which will allow the user to reset the state of a broadcast without breaking any other values. closes #582 image

Should we call it "Reset States" so that it corresponds to both ?

cmd/oceantv/main.go
```diff
@@ -47,7 +47,7 @@ import (
 
 const (
 	projectID          = "oceantv"
-	version            = "v0.10.7"
+	version            = "v0.10.8"
```
saxon-milton, line 50: This probably actually a minor bump

cmd/oceanbench/broadcast.go
```diff
@@ -435,6 +436,23 @@ func broadcastHandler(w http.ResponseWriter, r *http.Request) {
 
 		}
 		msg = "slate uploaded successfully"
+	case broadcastResetState:
+		log.Println("Sending Reset State Req")
```
trekhopton, line 440: Lowercase for log messages.

trekhopton: Use full words too, not "Req".

saxon-milton (approved):

trekhopton (approved):

PR #585, opened 2025-06-13
model/subfeed: remove feedID from already unique key

trekhopton:
This was done because have two globally unique IDs in the key is unnecessary.

NOTE: I'll also need to bump the openfish version once I've created a tag for the latest PR (https://github.com/ausocean/openfish/pull/330).
I tested this using the openfish changes locally, and it passed.


saxon-milton (approved):
Commit message needs some work :P

trekhopton:
Models have migrated to another repo but I missed this PR, no use merging here. I'll close this and open a PR on ausoceantv.

PR #586, opened 2025-06-16
otv: notify on entry into failure states

saxon-milton:
While we notify on error events, not all error events lead to entry into a failure state, in which case the broadcast will not stream, so it would be good to differentiate these. Therefore, we are not also notifying on entry into failure states.


cmd/oceantv/broadcast_hardware_machine.go
```diff
@@ -719,10 +724,20 @@ func (s *hardwareFailure) UnmarshalJSON(data []byte) error {
 // New implements registry.Newable for creating a fresh value of
 // hardwareFailure from an existing value.
 func (s hardwareFailure) New(args ...interface{}) (any, error) {
-	return newableWithContext(func(ctx *broadcastContext) any { return newHardwareFailure(nil) }, args...)
+	return newableWithContext(func(ctx *broadcastContext) any { return newHardwareFailure(ctx, nil) }, args...)
+}
+func (s *hardwareFailure) enter() {
+	notifyMsg := "entering hardware failure state"
+	notifyKind := broadcastGeneric
```
scruzin, line 731: For clarity, I think we should introduce a new kind of notification, e.g., `broadcastNonFailureError`, rather than the bland `broadcastGeneric`.

scruzin (changes requested):
LGTM. Just one change suggested.

cmd/oceantv/broadcast_hardware_machine.go
scruzin: Or perhaps `broadcastPreFailure`, since entering the failure state may not result in a failure.

scruzin (approved):

PR #588, opened 2025-06-17
ausoceantv (free): Use env vars for maps API key

trekhopton:
This was done to improve security and not hardcode the API key in the source code.

### DIFF
--- a/ausoceantv.yaml
+++ b/ausoceantv.yaml
@@ -5,6 +5,7 @@ env_variables:
   NETRECEIVER_CREDENTIALS: gs://ausocean/NetReceiver-55f1f53a6d8a.json
   VIDGRIND_CREDENTIALS: gs://ausocean/VideoGrinder-b0ad82abac05.json
   AUSOCEANTV_SECRETS: gs://ausocean/AusOceanTV-secrets.txt
+  AUSOCEANTV_MAPS_API_KEY: gs://ausocean/AusOceanTV-maps-API-key.json
   OAUTH2_CALLBACK: https://ausocean.tv/api/v1/auth/oauth2callback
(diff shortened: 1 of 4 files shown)


scruzin:
Any reason why you didn't handle those json files the same way we handle others, i.e., server side rather than client side?

Basically, use gauth.GetSecrets (which internally uses gauth.ReadStorageBucket or reads a local file) then pass the key as a template var to the rendered typescript.

trekhopton:
> Any reason why you didn't handle those json files the same way we handle others, i.e., server side rather than client side?
> 
> Basically, use gauth.GetSecrets (which internally uses gauth.ReadStorageBucket or reads a local file) then pass the key as a template var to the rendered typescript.

As a general rule for AusOceanTV, we're trying not to use go templates unless absolutely necessary. Templates are harder to maintain than a clearly separated frontend and API-accessed backend, and templates can easily be overused to create slow applications. That being said, I'm not against a hybrid approach if we need it, we just haven't needed it so far.

We still could have the backend read the key and provide it via API, but I chose to have vite build it into the JS to make it faster. (one less API call), this is a common practice, provided the key is restricted, which it is now.

ao-david (approved):
Is the added complexity worth not hardcoding the key? the key is still visible to the user anyway.

If the key is in the vite env file, I think you could avoid some of the parsing values?

scruzin:
> Is the added complexity worth not hardcoding the key? the key is still visible to the user anyway.
> 
> If the key is in the vite env file, I think you could avoid some of the parsing values?

If security is the same then, yes, let's keep it simple.

trekhopton:
Keep it as is for simplicity.

PR #590, opened 2025-06-18
489 esp calibration

ao-david:
This change adds a feature to be able to calibrate the voltage sensors of a controller. This is done with a popover which makes it easy and separate to the rest of the devices form. This should make it easier to configure devices, and will also help towards keeping track of devices that are calibrated and those that aren't in the future.

image

### DIFF
--- a/cmd/oceanbench/set.go
+++ b/cmd/oceanbench/set.go
@@ -374,14 +374,14 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
 // device voltages.
 //
 // Query params:
-//   - ma: MAC address
-//   - vb: 	Battery Voltage
+//   - ma:  MAC address
+//   - vb:  Battery Voltage
 //   - vnw: Network Voltage
 //   - vp1: Power 1 Voltage
 //   - vp2: Power 2 Voltage
 //   - vp3: Power 3 Voltage
-//   - va: 	Alarm Voltage
-//   - vr: 	Alarm Recovery Voltage
+//   - va:  Alarm Voltage
+//   - vr:  Alarm Recovery Voltage
 //
 // NOTE: All voltages are parsed in Volts.
 func calibrateDevicesHandler(w http.ResponseWriter, r *http.Request) {
@@ -441,12 +441,16 @@ func calibrateDevicesHandler(w http.ResponseWriter, r *http.Request) {
 	device, err := model.GetDevice(ctx, settingsStore, model.MacEncode(mac))
 	if err != nil {
 		writeDevices(w, r, "unable to get device to calibrate (%s): %v", mac, err)
+		return
 	}
 
 	// Names of the voltage sensors to calibrate.
 	var voltageSensors = []string{
-		model.NameBatterySensor, model.NameNWVoltage, model.NameP1Voltage,
-		model.NameP2Voltage, model.NameP3Voltage,
+		model.NameBatterySensor,
+		model.NameNWVoltage,
+		model.NameP1Voltage,
+		model.NameP2Voltage,
+		model.NameP3Voltage,
 	}
 
 	// Load the most recent sensor values.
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, "/set/devices?ma="+ma, http.StatusFound)
 }
 
+// calibrateDevicesHandler handles calibration of controller-type
+// device voltages.
+//
+// Query params:
+//   - ma: MAC address
+//   - vb: 	Battery Voltage
```
saxon-milton, line 378: bigger indent ?

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+//   - vnw: Network Voltage
+//   - vp1: Power 1 Voltage
+//   - vp2: Power 2 Voltage
+//   - vp3: Power 3 Voltage
+//   - va: 	Alarm Voltage
```
saxon-milton, line 383: ditto

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	vr, err := strconv.ParseFloat(r.FormValue("vr"), 64)
+	if err != nil && vr != 0 {
+		writeDevices(w, r, "unable to parse alarm recovery voltage: %v", err)
+		return
+	}
+
+	device, err := model.GetDevice(ctx, settingsStore, model.MacEncode(mac))
+	if err != nil {
+		writeDevices(w, r, "unable to get device to calibrate (%s): %v", mac, err)
```
saxon-milton, line 443: should we return right after this ?

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// Names of the voltage sensors to calibrate.
+	var voltageSensors = []string{
+		model.NameBatterySensor, model.NameNWVoltage, model.NameP1Voltage,
```
saxon-milton, line 448: This might be a style preference thing, but I reckon it would look slightly nicer if it was listed like this
```
var voltageSensors = []string{
    model.NameBatterySensor,
    model.NameNWVoltage,
....
```

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+	sensors, err := model.GetSensorsV2(ctx, settingsStore, model.MacEncode(mac))
+	if err != nil {
+		writeDevices(w, r, "unable to get sensors (%s): %v", mac, err)
+		return
+	}
+
+	// Calibrate each of the voltage sensors.
+	var msgs []string
+	for _, sensor := range sensors {
+		if sliceutils.ContainsString(voltageSensors, sensor.Name) {
```
saxon-milton, line 462: Could invert this to reduce nesting i.e. 
```
if !sliceutils.ContainsString(...) {
    continue
}
```

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+			scalar, err := model.GetLatestScalar(ctx, mediaStore, model.ToSID(mac, sensor.Pin))
+			if err != nil {
+				msgs = append(msgs, fmt.Sprintf("unable to get latest scalar for %s: %v", sensor.Name, err))
+				continue
+			}
+			reportedTime := time.Unix(scalar.Timestamp, 0)
+
+			// Check if the scalar was recently reported (last 2 monitor periods).
```
saxon-milton, line 470: Maybe elaborate in the comment why we continue here ?

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+			case model.NameNWVoltage:
+				actual = vnw
+			case model.NameP1Voltage:
+				actual = vp1
+			case model.NameP2Voltage:
+				actual = vp2
+			case model.NameP3Voltage:
+				actual = vp3
+			default:
+				// This shouldn't be possible with the ContainsString check.
```
saxon-milton, line 490: If this isn't possible we should probably panic to be loud

saxon-milton: We should probably have a global panic handler if we don't have one yet, which would allow us to recover from it and notify through email like otv

cmd/oceanbench/set.go
```diff
@@ -368,6 +370,149 @@ func editDevicesHandler(w http.ResponseWriter, r *http.Request) {
...
+			}
+
+			// This most likely means the field was left blank. This is not a
+			// meaningful way of calibrating the system.
+			if actual == 0 {
```
saxon-milton, line 495: Is this bad ? Should we log or notify the user ?

saxon-milton (changes requested):
Cool nice feature!

PR #605, opened 2025-07-15
system: update rig voltage alarm defaults

trekhopton:
This is done to reflect changes in the PCB hardware and hopefully prevent unintentionally discharging batteries if defaults aren't updated.

These values should correspond to roughly 24.5V for AlarmVoltage and 25V for AlarmRecoveryVoltage.


ao-david (approved):
Thanks for doing this

trekhopton:
> Thanks for doing this

No worries, I remembered you said we should change this a while ago.

PR #608, opened 2025-07-22
Bench: Autoconfig for cameras

ao-david:
This change adds the last few requirements for autoconfiguration for camera type devices


ao-david:
> This failed its initial test, still needs to set the mon/act periods as well as the inputs.

These have now been updated

PR #614, opened 2025-08-07
Bench/API: Add sensor data endpoint

ao-david:
This change adds an endpoint to the API which returns the transformed data from a sensor, with a timestamp and float value.

This is required for ausocean/ausoceantv#71

### DIFF
--- a/cmd/oceanbench/api.go
+++ b/cmd/oceanbench/api.go
@@ -28,6 +28,7 @@ LICENSE
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/api.go
```diff
@@ -311,6 +313,104 @@ func getVarsForSiteHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write(data)
 }
 
+// getSensorDataHandler handles requests to get data for a given sensor. The API transforms
+// the scalar data based on the sensor transform function and returns it.
+//
+// Expected path format:
+//
+//	/api/get/sensor/data/<mac>,<pin>,<start Timestamp>,<finish Timestamp>
```
trekhopton, line 321: Can we change this to a query and query parameters?

trekhopton: Better now than changing the API later. This format is strange.

trekhopton (changes requested):

cmd/oceanbench/api.go
ao-david: updated now

trekhopton (approved):

PR #626, opened 2025-09-19
Youtube: Export video status checking

ao-david:
This change returns the video object when uploading a video. This video contains an ID which can then be used to check the status of the upload.

### DIFF
--- a/youtube/upload.go
+++ b/youtube/upload.go
@@ -27,6 +27,7 @@ package youtube
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"time"
@@ -35,6 +36,8 @@ import (
 	"github.com/ausocean/cloud/utils"
 	"google.golang.org/api/youtube/v3"
 )
+
+var ErrUnknownStatus = errors.New("unknown video status")
 
 // VideoUploadOption is a functional option type for configuring YouTube video uploads.
 type VideoUploadOption func(*youtube.Video) error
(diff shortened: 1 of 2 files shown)


youtube/upload.go
```diff
@@ -176,23 +186,62 @@ func UploadVideo(ctx context.Context, media io.Reader, opts ...VideoUploadOption
...
+
+// CheckUploadStatus checks the status for the video with the associated videoID.
+// the returned status will be one of:
+// - UploadStatusUploaded
+// - UploadStatusProcessed
+// - UploadStatusFailed
+// - UploadStatusRejected
+// - UploadStatusDeleted
+func CheckUploadStatus(ctx context.Context, videoID string) (string, error) {
+	// Force using the default account (AusOcean's account).
```
trekhopton, line 216: We're not using the AusOcean main channel right?

trekhopton (approved):

youtube/upload.go
ao-david: Keep in mind that the upload function above this is where the upload being forced to use the ausocean channel

youtube/upload.go
```diff
@@ -176,23 +186,62 @@ func UploadVideo(ctx context.Context, media io.Reader, opts ...VideoUploadOption
...
+
+	if len(vid.Items) == 0 {
+		return "", fmt.Errorf("video not found")
+	}
+
+	switch vid.Items[0].Status.UploadStatus {
+	case "processed":
+		return UploadStatusProcessed, nil
+	case "failed":
+		return UploadStatusFailed, nil
```
Saxon1, line 235: Should these "bad" statuses have us return an error ?

Saxon1 (approved):
Seems OK but maybe we should think about what we do in the case of the weird statuses.

youtube/upload.go
```diff
@@ -36,6 +37,8 @@ import (
 	"google.golang.org/api/youtube/v3"
 )
 
+var ErrUnknownStatus = errors.New("unknown video status")
```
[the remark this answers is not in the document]
ao-david: I think that the decision of what to do with the status is up to the caller, and the little bit of extra detail doesn't hurt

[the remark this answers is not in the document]
ao-david: Is it just the Failed and Rejected statuses that you would consider "bad" statuses? I would assume that the Deleted status is probably only used if the user has called for a deletion, and the caller would probably want to check for that?

youtube/upload.go
ao-david: I'm going to deal with this in another PR

PR #630, opened 2025-10-08
bench/t/search: render GPS / NMEA text

### LINKED ISSUE #629: bench: Permit raw text in search results
At present, searching for a T(ext) pin assumes that the retuned text is log text in JSON format. There needs to be a way to return raw text, such as GPS data.

trekhopton:
This was done for testing/debugging GPS and NMEA sentence searching.

It does basic formatting, and provides the raw text to view and download. 

Closes #629 

(I also tested that viewing logs still works)

image

image


scruzin (approved):
LGTM. The basic NMEA parser is more than we need, but doesn't hurt.

PR #643, opened 2025-11-12
Docs: Add basic broadcasting information

ao-david:
This change adds an overview of the different settings of the broadcast page, as well as how to properly remove a broadcast from failure mode. A link to this documentation has also been added to the broadcast page.


docs/oceanbench/broadcast/failure-mode.md
```diff
@@ -0,0 +1,18 @@
+# Failure Mode
+
+![Failure Mode](../../images/broadcasting-failure-mode.png)
+
+The failure mode toggle indicates that an issue has occurred whilst trying to start a broadcast. The issue must have prevented the broadcast from starting multiple times, as OceanTV allows for 3 retries before entering failure mode.
```
Saxon1: I think it can be entered at any point during broadcast, though it is most likely to occur at starting.

Saxon1 (approved):

Issue #128, opened 2024-06-15
bench: Declarative variables and variable types


scruzin:
Now that we are using the Device.Type property, the variables and their corresponding types for each device type can be declared statically (in code), rather than communicated dynamically.

PR #664, opened 2025-11-27
Use static variable types

ao-david:
This change adds static var types and uses them on the device page.

image
image

part of #128


system/camera/var_types.go
```diff
@@ -0,0 +1,133 @@
+package camera
+
+import (
+	"github.com/ausocean/av/revid/config"
+	"github.com/ausocean/cloud/model"
+)
+
+var VarTypes = []model.VarType{
+	{Name: config.KeyTransformMatrix,
```
trekhopton: Lets import the variable information from prod, and try to minimise the number of places you have to edit / add variable names and types.

trekhopton (approved):

system/controller/var_types.go
```diff
@@ -0,0 +1,24 @@
+package controller
+
+import "github.com/ausocean/cloud/model"
+
+// VarTypes describes the types of the variables used by a controller.
+var VarTypes = []model.VarType{
```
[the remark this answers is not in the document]
ao-david: I might actually add some documentation to the Docs first, and then merge this after, just so we are following the ground truth model from the start.

Saxon1 (changes requested):
I think you should use a factory method to get these instead; a global is mutable which is potentially dangerous.

system/camera/var_types.go
saxon-milton: Also, you'd use a singleton pattern to load it so that there's only one immutable instance of the ICD in each side of the software.

system/controller/var_types.go
```diff
@@ -0,0 +1,32 @@
+package controller
+
+import "github.com/ausocean/cloud/model"
+
+// VarTypes returns a copy of the controller varTypes array, this is to prevent
+// the global variable from being mutated.
+func VarTypes() []model.VarType {
+	safeCopy := make([]model.VarType, len(varTypes))
+	copy(safeCopy, varTypes)
+	return safeCopy
```
saxon-milton: I think you can just do this instead so that you don't risk modifying it for other things.
```
return []model.VarType{
	{Name: "LogLevel", Type: model.VarTypeUint},
	{Name: "Pulses", Type: model.VarTypeUint},
	{Name: "PulseWidth", Type: model.VarTypeUint},
	{Name: "PulseDutyCycle", Type: model.VarTypeUint},
	{Name: "PulseCycle", Type: model.VarTypeUint},
	{Name: "AutoRestart", Type: model.VarTypeUint},
	{Name: "AlarmPeriod", Type: model.VarTypeUint},
	{Name: "AlarmNetwork", Type: model.VarTypeUint},
	{Name: "AlarmVoltage", Type: model.VarTypeUint},
	{Name: "AlarmRecoveryVoltage", Type: model.VarTypeUint},
	{Name: "PeakVoltage", Type: model.VarTypeUint},
	{Name: "HeartbeatPeriod", Type: model.VarTypeUint},
	{Name: "Power0", Type: model.VarTypeBool},
	{Name: "Power1", Type: model.VarTypeBool},
	{Name: "Power2", Type: model.VarTypeBool},
	{Name: "Power3", Type: model.VarTypeBool},
	{Name: "PulseSuppress", Type: model.VarTypeBool},
}
```

ao-david: Ah, that's a bit neater. Done

PR #673, opened 2025-12-02
bench: Use a time picker instead of a date picker on the broadcast page

### LINKED ISSUE #38: broadcast: only need hours and mins on UI now
It no longer makes sense to have the date in the broadcasts configuration page considering it’s now implicit that broadcasts are repeated according to the hour and min if the broadcast’s “Enabled” field is true i.e. the date doesn't have any meaning.

elliot-ausocean:
The broadcast time doesn't use a date anymore, only on a time so it makes more sense to use a time picker. Resolves #38
image

### DIFF
--- a/cmd/oceanbench/s/main.js
+++ b/cmd/oceanbench/s/main.js
@@ -209,13 +209,13 @@ function syncDateTime(datetimeID, timestampID, timezoneID, pickerUsed) {
       timestampElem.value = "";
       return;
     }
-    // We only have a time.
+    // Check if we only have a time.
     if (datetime.length == 5) {
       datetime += ":00"; // Append seconds to make RFC3339 compliant.
       // Prepend a date to make it a valid datetime.
       datetime = "2000-01-01T" + datetime;
     }
-    // We have both a date and a time.
+    // Check if we have both a date and a time.
     else if (datetime.length == 16) {
       datetime += ":00"; // Append seconds to make RFC3339 compliant.
     }


ao-david (changes requested):

cmd/oceanbench/s/main.js
```diff
@@ -184,53 +182,76 @@ function tzParseUTCOffset(offset) {
...
+    // Add the offset (in seconds) to the timestamp.
+    const localTimestamp = parseInt(timestamp) + Math.round(parseFloat(timezone) * 3600);
+    // Convert the timestamp to an ISO string
+    const datetime = new Date(localTimestamp * 1000).toISOString();
+    if (pickerElem.type == "time") {
+      // Extract the hours/mins portion only (HH:mm).
+      pickerElem.value = datetime.slice(11, 16);
+    } else if (pickerElem.type == "datetime-local") {
+      // Extract the date and hours/mins only (YYYY-MM-DDTHH:mm).
+      pickerElem.value = datetime.slice(0, 16);
```
ao-david: Is there a reason that this is now 0,16 and not the 0 and -1 that it used to be?

Its generally preferable to not change stuff that isn't relevant to the change.

cmd/oceanbench/s/main.js
```diff
@@ -184,53 +182,76 @@ function tzParseUTCOffset(offset) {
...
-  }
-  if (dateChanged) {
-    // Go from local date to Unix timestamp.
-    var s = document.getElementById(dateID).value;
-    if (s == "") {
-      document.getElementById(tsID).value = "";
+// syncDateTime syncs a date/time picker and a timestamp input.
+// If either of these are changed, the other is updated to match.
+// pickerUsed signifies if the date/time picker was changed.
+function syncDateTime(datetimeID, timestampID, timezoneID, pickerUsed) {
```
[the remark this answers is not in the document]
ao-david: Also, I would prefer that other variables and lines that aren't relevant to the change aren't changed, just because it makes the PR harder to see what has actually changed. For this PR, it is okay, but just in future, I would recommend to keep variable names as they are.

Also, see the variable naming conventions that we have (technically for golang, but spill over into our other languages): https://docs.google.com/spreadsheets/d/1SuZtrklXvd77LEsghyEZjpT5vxK1D-Ia8GOuChZlw78/edit?gid=0#gid=0

cmd/oceanbench/s/main.js
```diff
@@ -184,53 +182,76 @@ function tzParseUTCOffset(offset) {
...
+  if (pickerUsed) {
+    // Update the timestamp from the time picker.
+    let datetime = pickerElem.value;
+    if (datetime == "") {
+      timestampElem.value = "";
       return;
     }
-    if (s.length == 16) {
-      s += ":00"; // Append seconds to make RFC3339 compliant.
+    // We only have a time.
```
trekhopton: "Check if we only have a time." would make more sense I think.

trekhopton (approved):

Issue #670, opened 2025-12-01
Bench/Broadcast: Make the refresh button not advanced feature


ao-david:
The reset button is quite useful to have, to fully minimise the risk of a refresh page submit. It should be always shown, and not only shown as an advanced option.

<img width="1015" height="339" alt="Image" src="https://github.com/user-attachments/assets/43281fab-e8a3-42bc-ba73-a05b4ae6944c" />

ao-david:
Hi @corno25kumar-creator feel free to fork, make changes, and than make a PR for us to review.

PR #675, opened 2025-12-03
Fix: Refresh button always visible (#670)

pav2345:
This PR makes the Refresh button always visible instead of only in Advanced Options.
This is my first contribution to this project — if I’ve made any mistakes, I sincerely apologize.

✔️ Fixes issue #670


trekhopton (approved):
Thanks.

PR #676, opened 2025-12-03
oceantv: add broadcast and hardware states to log output

### LINKED ISSUE #565: OTV: Put state in logs
It would be nice in OTV logs to put the current state in each request so that it is clear what state handlers will be handling the event, and hence what the expected behaviour from that point will be.

elliot-ausocean:
This will make it clearer what state handler will be used and hence the expected behaviour. Resolves #565.
Example Output: `[OceanTV] 2025/12/03 12:14:53 (name: test, broadcast state: main.directIdle, hardware state: hardwareOff, id: ) broadcast saved`


ao-david (approved):
Rebase this with the other changes and rerun the tests before merging

PR #688, opened 2025-12-19
687 add light sensor to the auto configuration in rig systems

deborah-baker:
Added a light sensor to the auto config of the rig system. Also updated the version of nmea to v0.4.1 to include the light sensor.


ao-david (changes requested):
It seems like you created your new branch from your previous and not from the main branch, which is why there are the other changes than just the ones you wanted.

ao-david (approved):
Nice one!

Issue #697, opened 2026-01-23
Bench: Site and Device Log


ao-david:
It would be good to be able to keep a log of anything of note for each device, and site. For example, if we deploy something bespoke it is good to know what the changes were that were made.

I propose that we have an interface for each device where you can write a note, and optionally upload photos to describe what is happening with the device.

The log should be associated with the current site of the device so that we can see what has been on what site when, and what has been changed, even if a device has been moved to a different site.

The log should also be associated with the device, so that even when the device has moved to a different site, the log is still queryable.

As a starting point, we should implement this without images, as we could always link to a URL where photos are hosted.

ao-david:
addressed in #699 #701 #782

PR #701, opened 2026-01-29
697 bench: making a log page

deborah-baker:
a page has been added to the admin section to create site and device logs.
#697 

image

image

### DIFF
--- a/cmd/oceanbench/t/log.html
+++ b/cmd/oceanbench/t/log.html
@@ -33,7 +33,7 @@
         <form action="/api/set/log/" method="post" novalidate>
             <div class="input-group mb-3">
                 <span class="input-group-text" id="Skey">Site Key: </span>
-                <input type="text" class="form-control" name="sk" value="{{ .Skey }}" aria-label="Skey" aria-describedby="basic-addon1">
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/t/log.html
```diff
@@ -0,0 +1,53 @@
...
+      </nav-menu>
+      <site-menu id="sitemenu" {{ if .Profile }}selected-data="{{ .Profile.Data }}" {{ end }} slot="site-menu"></site-menu>
+    </header-group>
+    <section id="main" class="main">
+      <h1 class="container-md">Log</h1>
+      <div class="border rounded p-4 container-md bg-white">
+        <form action="/api/set/log/" method="post" novalidate>
+            <div class="input-group mb-3">
+                <span class="input-group-text" id="Skey">Site Key: </span>
+                <input type="text" class="form-control" name="sk" value="{{ .Skey }}" aria-label="Skey" aria-describedby="basic-addon1">
```
ao-david: This should probably be `type="number"` with `step=1` to make it so that only integers (which is what a SKey is) can be inputted.

However, since I think this should be temporary, I'm happy to leave this as is.

ao-david (changes requested):
Please add a screenshot to the PR to make it easier for reviewers to see what you have done. This is good practice for UI changes.

trekhopton (approved):

ao-david (approved):

PR #715, opened 2026-03-12
feat: add lit site footer

trekhopton:
Separated from mission-control branch to add site footer lit component.

The styling is currently unchanged, it's just an element now that uses tailwind and can be improved later.

image

### DIFF
--- a/cmd/oceanbench/ts/site-footer.ts
+++ b/cmd/oceanbench/ts/site-footer.ts
@@ -8,11 +8,6 @@ export class SiteFooter extends LitElement {
   @property({ type: String }) orgName = "Australian Ocean Laboratory Limited (AusOcean)";
   @property({ type: String }) licenseHref = "https://www.ausocean.org/license";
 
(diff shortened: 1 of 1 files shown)


cmd/oceanbench/ts/site-footer.ts
```diff
@@ -0,0 +1,41 @@
...
+import { LitElement, html } from "lit";
+import { customElement, property } from "lit/decorators.js";
+
+@customElement("site-footer")
+export class SiteFooter extends LitElement {
+  @property({ type: Number }) startYear = 2019;
+  @property({ type: String }) orgName = "Australian Ocean Laboratory Limited (AusOcean)";
+  @property({ type: String }) licenseHref = "https://www.ausocean.org/license";
+
+  // Render into light DOM so global Tailwind classes apply.
```
ao-david: Why do we have to do this here? We don't need to do this in AusOceanTV do we?

ao-david (approved):

PR #719, opened 2026-03-12
feat: add mission control page

trekhopton:
The idea is that this is a page for super admins (ausocean) to see everything that is going on on the platform without having to browse through many menus or pages.
For starters it lists all currently created sites.
In future it will show devices, broadcasts, variables, etc.

image


trekhopton:
Okay, looking better, here's the updated UI ACTUALLY using tailwind: 

image

PR #721, opened 2026-03-18
bench: add --testdata flag for standalone datastore seeding

trekhopton:
This was done because I would often check things on the production data since there wasn't enough populated data data and it was tedious to do it manually. Now we can populate the test data and test in standalone more thoroughly, which results in safer testing (not working with production data).


ao-david (approved):
Just review the site keys and names

PR #723, opened 2026-03-20
bench: change routing to fiber

trekhopton:
Done to make it easier to write routes.


cmd/oceanbench/api.go
```diff
@@ -66,45 +68,52 @@ type minimalSite struct {
...
-	http.HandleFunc("/api/set/log/", wrapAPI(setLogHandler))
-
-	http.HandleFunc("/api/test/upload/", wrapAPI(testUploadHandler))
-	http.HandleFunc("/api/test/download/", wrapAPI(testDownloadHandler))
+// All routes under /api automatically go through the wrapAPI middleware.
+func setupAPIRoutes(app *fiber.App) {
+	// Create /api group; wrapAPI middleware is applied to every route in the group.
+	api := app.Group("/api", wrapAPI)
+
+	api.Get("/get/site/*", adaptor.HTTPHandlerFunc(getSiteHandler))
```
ao-david: Are the adaptor calls just a temporary solution to convert a http handler into a fiber handler? 

Would it be possible to move this to middleware as well?

cmd/oceanbench/api_v1.go
```diff
@@ -34,35 +34,58 @@ import (
...
-			writeJSONError(w, http.StatusUnauthorized, "user could not be authenticated")
-			return
+// withProfileJSON is Fiber middleware that authenticates the request and stores
+// the *gauth.Profile in c.Locals(profileKey). If authentication fails it writes
+// a JSON error response and aborts the chain.
+func withProfileJSON(c *fiber.Ctx) error {
+	// Build a *http.Request from the Fiber/fasthttp context so we can reuse the
+	// existing getProfile helper which relies on net/http cookies.
+	var r http.Request
+	if err := fasthttpadaptor.ConvertRequest(c.Context(), &r, true); err != nil {
```
ao-david: Is there a reason we use fasthttpadaptor here, and just adaptor in the previous file?

PR #724, opened 2026-03-20
bench: add media manager page to batch delete media

trekhopton:
This was done because we have large amounts of media costing us in cloud storage and no effective way to delete it.

image

### DIFF
--- a/cmd/oceanbench/api_v1.go
+++ b/cmd/oceanbench/api_v1.go
@@ -169,7 +169,15 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
 // getMediaKeysHandler handles GET /api/v1/media.
 //
 // Super-admin only. Returns an array of MtsMedia key IDs for the currently
-// selected site for the past month.
+// selected site.
+//
+// Query Parameters:
+//   - from: Start unix timestamp in seconds (optional, but 'from' or 'to' is required)
+//   - to: End unix timestamp in seconds (optional, but 'from' or 'to' is required)
+//   - mac: Filter by encoded device MAC address (optional)
+//   - pin: Filter by specific pin (e.g., "S0") (optional)
+//   - limit: Maximum number of keys per MID, default 50000 (optional)
+//   - tz: Timezone location for date grouping, default "UTC" (optional)
 func getMediaKeysHandler(c *fiber.Ctx) error {
 	p := c.Locals(profileKey).(*gauth.Profile)
 	if !isSuperAdmin(p.Email) {
@@ -183,6 +191,31 @@ func getMediaKeysHandler(c *fiber.Ctx) error {
 		return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": "could not resolve site: no site selected in profile"})
 	}
 
+	var query struct {
+		From  *int64 `query:"from"`
+		To    *int64 `query:"to"`
+		Mac   *int64 `query:"mac"`
+		Pin   string `query:"pin"`
+		Limit *int   `query:"limit"`
+		TZ    string `query:"tz"`
+	}
+
+	if err := c.QueryParser(&query); err != nil {
+		return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": "invalid query parameters: " + err.Error()})
+	}
+
+	var ts []int64
+	if query.From != nil {
+		ts = append(ts, *query.From)
+	}
+	if query.To != nil {
+		ts = append(ts, *query.To)
+	}
+
+	if len(ts) == 0 {
+		return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": "time range (ts) is required"})
+	}
+
 	ctx, cancel := context.WithTimeout(c.Context(), 5*time.Minute)
 	defer cancel()
 
@@ -191,45 +224,25 @@ func getMediaKeysHandler(c *fiber.Ctx) error {
 		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{"error": fmt.Sprintf("could not get devices: %v", err)})
 	}
 
-	var ts []int64
-	if fromStr := c.Query("from"); fromStr != "" {
-		if fromTS, err := strconv.ParseInt(fromStr, 10, 64); err == nil {
-			ts = append(ts, fromTS)
-		}
-	}
-	if toStr := c.Query("to"); toStr != "" {
-		if toTS, err := strconv.ParseInt(toStr, 10, 64); err == nil {
-			ts = append(ts, toTS)
-		}
+	deviceFilter := int64(-1)
+	if query.Mac != nil {
+		deviceFilter = *query.Mac
 	}
 
-	if len(ts) == 0 {
-		return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": "time range (ts) is required"})
-	}
-
-	deviceFilterStr := c.Query("device")
-	var deviceFilter int64 = -1
-	if deviceFilterStr != "" {
-		if d, err := strconv.ParseInt(deviceFilterStr, 10, 64); err == nil {
-			deviceFilter = d
-		}
-	}
-
-	pinFilterStr := c.Query("pin")
+	pinFilterStr := query.Pin
 
 	limit := 50000
-	if limitStr := c.Query("limit"); limitStr != "" {
-		if l, err := strconv.Atoi(limitStr); err == nil && l >= 0 {
-			limit = l
-		}
+	if query.Limit != nil && *query.Limit >= 0 {
+		limit = *query.Limit
 	}
 
 	location := time.UTC
-	if tzStr := c.Query("tz"); tzStr != "" {
-		if loc, err := time.LoadLocation(tzStr); err == nil {
+	if query.TZ != "" {
+		if loc, err := time.LoadLocation(query.TZ); err == nil {
 			location = loc
 		} else {
-			log.Printf("getMediaKeysHandler: failed to load location %q: %v", tzStr, err)
+			log.Printf("getMediaKeysHandler: failed to load location %q: %v", query.TZ, err)
+			return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": fmt.Sprintf("failed to load location %q: %v", query.TZ, err)})
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/oceanbench/main.go
```diff
@@ -118,9 +118,10 @@ type commonData struct {
 
 // TestDataConfig defines the structure for injecting standalone datastore test configurations via JSON.
 type TestDataConfig struct {
```
trekhopton, line 120: Upon reflection I think this test stuff shouldn't live in main. Will move.

cmd/oceanbench/api_v1.go
```diff
@@ -93,18 +99,6 @@ type minimalSiteV1 struct {
 	Public bool   `json:"Public"`
 }
 
-func writeJSON(w http.ResponseWriter, status int, v interface{}) {
```
ao-david, line 96: Why is this deleted?

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
 	}
 	return c.JSON(out)
 }
+
+// getMediaKeysHandler handles GET /api/v1/media.
+//
+// Super-admin only. Returns an array of MtsMedia key IDs for the currently
+// selected site for the past month.
```
ao-david: For the past month or for a given month?

ao-david: I suspect it is for a given month, and uses unix timestamps. Maybe just add the format

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+			ts = append(ts, fromTS)
+		}
+	}
+	if toStr := c.Query("to"); toStr != "" {
+		if toTS, err := strconv.ParseInt(toStr, 10, 64); err == nil {
+			ts = append(ts, toTS)
+		}
+	}
+
+	if len(ts) == 0 {
```
ao-david, line 215: I think this should be 
```golang
if len(ts) <= 1 {
...
}
```
That would cover if only one of the timestamps was found. It may also be useful to return the error in the response

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+		return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": "time range (ts) is required"})
+	}
+
+	deviceFilterStr := c.Query("device")
```
ao-david: Is this passed as a mac? I'm pretty sure there are functions somewhere to handle MAC encoding/decoding and validity.

It would also be good to use the standard `ma` key if it is also

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+	var deviceFilter int64 = -1
+	if deviceFilterStr != "" {
+		if d, err := strconv.ParseInt(deviceFilterStr, 10, 64); err == nil {
+			deviceFilter = d
+		}
+	}
+
+	pinFilterStr := c.Query("pin")
```
ao-david: use standard `pn` for pin key

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+		for _, pin := range parsePins(dev.Inputs) {
+			if pinFilterStr != "" && !strings.EqualFold(pin, pinFilterStr) {
+				continue
+			}
+			mid := model.ToMID(dev.MAC(), pin)
+			if seenMIDs[mid] {
+				continue
+			}
+			seenMIDs[mid] = true
+			mids = append(mids, mid)
```
ao-david: Why do we need a map and an array here holding the same information?

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+	}
+
+	out := mediaSummaryV1{
+		KeysByDate: make(map[string][]string),
+		Summary:    make(map[string]int),
+	}
+
+	totalKeys := 0
+
+	for _, mid := range mids {
```
ao-david: if this is what the array is for, we can just use a range for loop on the map

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+	ctx, cancel := context.WithTimeout(c.Context(), 5*time.Minute)
+	defer cancel()
+
+	devices, err := model.GetDevicesBySite(ctx, settingsStore, skey)
+	if err != nil {
+		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{"error": fmt.Sprintf("could not get devices: %v", err)})
+	}
+
+	var ts []int64
+	if fromStr := c.Query("from"); fromStr != "" {
```
elliot-ausocean: Probably best to put all the query parameters into a struct and use QueryParser (see https://docs.gofiber.io/v2.x/api/ctx#queryparser). I would also suggest validating these (which should mostly be done for you by QueryParser) before doing the GetDevicesBySite query to avoid unnecessary queries.

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+		for _, k := range keys {
+			keyStr := strconv.FormatUint(uint64(k.ID), 10)
+			_, tsec, _ := datastore.SplitIDKey(k.ID)
+			dateStr := time.Unix(tsec, 0).In(location).Format("2006-01-02")
+			out.Summary[dateStr]++
+			if out.KeysByDate[dateStr] == nil {
+				out.KeysByDate[dateStr] = make([]string, 0)
+			}
+			out.KeysByDate[dateStr] = append(out.KeysByDate[dateStr], keyStr)
+			totalKeys++
```
ao-david: could do this by adding `len(keys)` to totalKeys in the upper level loop

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+	}
+	siteMIDs := make(map[int64]struct{})
+	for _, dev := range devices {
+		for _, pin := range parsePins(dev.Inputs) {
+			siteMIDs[model.ToMID(dev.MAC(), pin)] = struct{}{}
+		}
+	}
+
+	// Extract the lower 32 bits of the site's MIDs for fast key ownership validation
+	siteLowerMIDs := make(map[int64]struct{})
```
ao-david: This is bizaare, why not use bool as before?

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+	for mid := range siteMIDs {
+		lower32 := int64(uint64(mid) & 0xffffffff)
+		siteLowerMIDs[lower32] = struct{}{}
+	}
+
+	// Verify ownership securely using bitwise properties of the Key ID,
```
ao-david, line 357: 'securely'

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+	// eliminating thousands of manual datastore lookups.
```
ao-david, line 358: thanks for the performance description

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+			limit = l
+		}
+	}
+
+	location := time.UTC
+	if tzStr := c.Query("tz"); tzStr != "" {
+		if loc, err := time.LoadLocation(tzStr); err == nil {
+			location = loc
+		} else {
+			log.Printf("getMediaKeysHandler: failed to load location %q: %v", tzStr, err)
```
elliot-ausocean: Should this return an error?

cmd/oceanbench/api_v1.go
```diff
@@ -171,3 +165,260 @@ func getV1UserSitesHandler(c *fiber.Ctx) error {
...
+
+	return c.JSON(fiber.Map{
+		"deleted": deleted,
+	})
+}
+
+
+
+// parsePins splits a comma-separated pin string (e.g. "A0,V0,S0") into a slice of pin names.
+func parsePins(inputs string) []string {
```
ao-david, line 418: This seems like a very reusable function, perhaps would be better in utils? Alternatively there is a MakePins function in Netsender that will return an array of Pins (which contain the pin name)

cmd/oceanbench/main.go
```diff
@@ -687,6 +695,12 @@ func pages(selected string) []page {
 			Level: 1,
 			Perm:  model.AdminPermission,
 		},
+		{
+			Name:  "media manager",
+			URL:   "/admin/mediamanager",
+			Level: 1,
+			Perm:  model.AdminPermission,
```
ao-david, line 702: Should we add a first class Super Admin permission? Otherwise this feature would show up on any admins nav menu, even though they will not be able to view it

ao-david (approved):

cmd/oceanbench/api_v1.go
trekhopton: Unused. Not sure why there were there in the first place.

cmd/oceanbench/api_v1.go
trekhopton: One timestamp is allowed for unbounded results.

cmd/oceanbench/api_v1.go
trekhopton: I don't really want to bring over old standards to the new api just because. I think pin, and mac make more sense and are more readable.

ao-david: I think thats fine, but also think that if we are going to move away from our standards we should try and capture what we want the new standard to be so that we can try and move to a more uniform API

PR #726, opened 2026-03-26
docs: add page for MID

trekhopton:
This was done to better communicate and understand the way we use MIDs for media.

image


docs/netsender/mpeg-ts-media.md
```diff
@@ -0,0 +1,37 @@
+# MPEG-TS Media
+
+MPEG-TS (MTS) is the container format used for media data eg. audio, video, in AusOcean's cloud services.
+
+## Overview
+
+The `MtsMedia` type represents a clip of continuous media data.
```
ao-david: The documentation should include what the type looks like, not just referring to it by name, explaining what each of the fields are, in such a way that someone could implement it as a protocol.

trekhopton:
> Yeah, nice. Fair enough to not document the whole MTS. I think this is a good step!
> 
> ps. sorry for being a pain, but the docs are my baby :)

Yeah, would be nice to eventually perhaps since we do some unique things with it. 

No worries haha I think it's good to have high standards here, but at the same time I need to be able incrementally add to it because it's unlikely I'll sit down and prioritise large amounts of retrospective documentation. 

Thanks.

PR #729, opened 2026-04-08
bench: track sites using session storage and urls 

trekhopton:
Done to make multi-tab multi-site browsing possible.

image

### DIFF
--- a/cmd/oceanbench/auth.go
+++ b/cmd/oceanbench/auth.go
@@ -117,3 +117,17 @@ func profileData(profile *gauth.Profile) (int64, string) {
 	}
 	return key, p[1]
 }
+
+// requestSiteData gets the site key from the request URL if present,
+// otherwise falls back to the user's profile data.
+func requestSiteData(r *http.Request, profile *gauth.Profile) (int64, string) {
+	if siteStr := r.URL.Query().Get("site"); siteStr != "" {
+		if siteKey, err := strconv.ParseInt(siteStr, 10, 64); err == nil {
+			// A valid site key is in the URL.
+			// Ideally we would return the name as well, but we don't have it here
+			// unless we do a DB lookup. The name isn't critical for data retrieval.
+			return siteKey, ""
+		}
+	}
+	return profileData(profile)
+}
--- a/cmd/oceanbench/auth.go
+++ b/cmd/oceanbench/auth.go
@@ -124,8 +124,6 @@ func requestSiteData(r *http.Request, profile *gauth.Profile) (int64, string) {
 	if siteStr := r.URL.Query().Get("site"); siteStr != "" {
 		if siteKey, err := strconv.ParseInt(siteStr, 10, 64); err == nil {
 			// A valid site key is in the URL.
-			// Ideally we would return the name as well, but we don't have it here
-			// unless we do a DB lookup. The name isn't critical for data retrieval.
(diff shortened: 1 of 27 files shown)


cmd/oceanbench/ts/site-menu.ts
```diff
@@ -26,6 +27,9 @@ class SiteMenu extends LitElement {
             padding: 8px;
             border: 1px solid #ccc;
             border-radius: 4px;
+            width: 424px;
```
trekhopton, line 30: This was added because before the options load it would be tiny and then grow to size, which was jarring.

ao-david: How does this work on mobile?

cmd/oceanbench/ts/site-menu.ts
```diff
@@ -52,7 +56,37 @@ class SiteMenu extends LitElement {
     }
 
     firstUpdated() {
-        this.loadSites()
+        // Handle init logic for tab site selection
+        const urlParams = new URLSearchParams(window.location.search);
+        let siteParams = urlParams.get('site');
```
ao-david, line 61: `sk` is our normal param name for a site key

cmd/oceanbench/ts/site-menu.ts
```diff
@@ -52,7 +56,37 @@ class SiteMenu extends LitElement {
     }
 
     firstUpdated() {
-        this.loadSites()
+        // Handle init logic for tab site selection
+        const urlParams = new URLSearchParams(window.location.search);
```
ao-david, line 60: Could be 
```ts
const siteParams = (new URLSearchParams(window.location.search)).get('site')
```

ao-david: ie theres no particular reason that I can see that siteParams shouldn't be const also?

cmd/oceanbench/auth.go
```diff
@@ -117,3 +117,15 @@ func profileData(profile *gauth.Profile) (int64, string) {
...
 	return key, p[1]
 }
+
+// requestSiteData gets the site key from the request URL if present,
+// otherwise falls back to the user's profile data.
+func requestSiteData(r *http.Request, profile *gauth.Profile) (int64, string) {
+	if siteStr := r.URL.Query().Get("site"); siteStr != "" {
+		if siteKey, err := strconv.ParseInt(siteStr, 10, 64); err == nil {
+			// A valid site key is in the URL.
+			return siteKey, ""
```
ao-david, line 127: How often do we actually use the site name, is that going to cause problems?

ao-david (approved):
So keen for this :pray: 

Just a few comments

cmd/oceanbench/ts/site-menu.ts
trekhopton: It's not very nice but it's good enough until we have proper mobile support.

cmd/oceanbench/auth.go
trekhopton: Almost never, I think it was added for the site menu. profileData still exists if people want to use it.

cmd/oceanbench/ts/site-menu.ts
trekhopton: Prioritising readability over efficiency.

PR #730, opened 2026-04-09
bench: stop profileData causing panics for nil profile

trekhopton:
This was done to stop a panic for a non-logged in user.


ao-david (approved):
bump version

PR #732, opened 2026-04-13
update go versions to 1.26

trekhopton:
Done because go121 is no longer supported on app-engine.

All apps build.


elliot-ausocean (approved):

ao-david (approved):
Not sure if this is a good time to do a `go mod tidy` as well if not already done?

PR #734, opened 2026-05-01
Gauth session fix

trekhopton:
This was done because the provided MaxAge was seemingly being overwritten; in bench it defaulted to gorilla's 30 days, in ausocean tv with fiber it defaulted to 0 (sessions logout if the browser is closed).


trekhopton:
Not applicable.

PR #735, opened 2026-05-01
auth: if one-tap token is valid, return from GetProfile

trekhopton:
This was done so that it mirrors the oauth token route and fixes the session expiry token not adopting the maxoauthAge.

I tested this locally and it works. Before the change closing the browser of a one-tap session would require log in at browser open and revisit (Session expiry), now it has a 7 day expiry and the user stays logged in after browser close.

### DIFF
--- a/gauth/userauth.go
+++ b/gauth/userauth.go
@@ -514,6 +514,7 @@ func (ua *UserAuth) GetProfile(h backend.Handler) (*Profile, error) {
 		if err != nil {
 			return nil, fmt.Errorf("unable to set token session key: %w", err)
 		}
+		return profile, nil
 	}
 
 	err = sess.Set(profileKey, profile)


gauth/userauth.go
ao-david, line 479: Not required for this change, but I think this logic should be re-worked to make it more logical

Currently the if statement captures oauth2 tokens (happy path for oauth2 users), but then doesn't complete the logic until after the if statement.

The else statement (for onetap users) returns on all paths and so the following logic only applies to the first if statement group anyway. I propose that it should look more like:

```go
if *idTok != emptyIDTok {
  // Onetap handling.
  validator, err := idtoken.NewValidator(ctx)
  if err != nil {
	  return nil, fmt.Errorf("unable to get new validator: %w", err)
  }

  _, err = validator.Validate(ctx, idTok.IDToken, "")
  if err != nil {
  	return nil, fmt.Errorf("unable to validate idtoken: %w", err)
  }

  err = sess.Set(idTokenSessionKey, idTok)
  if err != nil {
	return nil, fmt.Errorf("unable to set token session key: %w", err)
  }
  return profile, nil
}

if tok.Valid() {
	return profile, nil
}

// Issue a new client request to refresh the OAuth token.
src := ua.cfg.TokenSource(ctx, tok)
log.Printf("%+v", tok)
newTok, err := src.Token()
if err != nil {
	return nil, fmt.Errorf("could not get refreshed token: %w", err)
}
clt := ua.cfg.Client(ctx, newTok)
data := profile.Data // Save optional data.
profile, err = fetchProfile(clt)
if err != nil {
	return nil, fmt.Errorf("fetch profile error: %w", err)
}

profile.Data = data // Restore optional data.
err = sess.Set(oauthTokenSessionKey, newTok)
if err != nil {
	return nil, fmt.Errorf("unable to set token session key: %w", err)
}

err = sess.Set(profileKey, profile)
if err != nil {
	return nil, fmt.Errorf("unable to set profile key: %w", err)
}

err = h.SaveSession(sess)
if err != nil {
	return nil, fmt.Errorf("session save error: %w", err)
}

return profile, nil
```

ao-david (approved):
Nice one!

PR #738, opened 2026-05-18
Add roadmap service in cmd

trekhopton:
This adds the internal roadmap tool to our services in cmd. 

It should have been merged long ago and incrementally developed but since it's not a customer facing service I didn't prioritise it. At some point I need to just get the code on main. It should probably live in its own repo, but we can move it later, I think I just need to get it merged somewhere. 

It's mostly self contained but it does touch gauth/secret.go so have a look at that, otherwise I think it mostly doesn't deserve the same scrutiny as a external facing service.
I can spend some time breaking up the PR if needed.


cmd/roadmap/auth.go
```diff
@@ -0,0 +1,131 @@
+/*
+DESCRIPTION
+  Authentication routes for AusOcean TV using OAuth2.
+
+AUTHORS
+  David Sutton <davidsutton@ausocean.org>
```
ao-david, line 6: I don't think I wrote this...

ao-david: Although this file is clearly a copy, why?

elliot-ausocean (approved):
LGTM. Probably fine keeping it as one big PR as long as you don't squash the commits when merging.

cmd/roadmap/webapp/.env.production
ao-david, line 1: This file should never include secrets right?

cmd/roadmap/webapp/.env.production
```diff
@@ -0,0 +1,5 @@
+# Loaded automatically by Vite when running in production mode (e.g.
+# `npm run build`). The deployed frontend uses this URL to talk to the
+# backend API.
+
+VITE_API_URL=https://roadmap.ausocean.org
```
ao-david, line 5: Doesn't the proxy setup in viteconfig.ts not set this for us?

cmd/roadmap/main.go
```diff
@@ -0,0 +1,389 @@
...
+)
+
+// service defines the properties of our web service.
+type service struct {
+	setupMutex  sync.Mutex
+	store       datastore.Store
+	debug       bool
+	standalone  bool
+	development bool
+	lite        bool
```
ao-david, line 66: I think a lot of these are just copied across from other services and not actually needed. especially ```lite```

cmd/roadmap/main.go
```diff
@@ -0,0 +1,389 @@
...
+	if err := c.BodyParser(&payload); err != nil {
+		return c.Status(fiber.StatusBadRequest).SendString(fmt.Sprintf("Invalid request: %v", err))
+	}
+
+	if len(payload.Tasks) == 0 {
+		return c.Status(fiber.StatusBadRequest).SendString("No tasks to update")
+	}
+
+	ctx := context.Background()
+	credentials, err := loadCredentials(ctx, "AUSOCEAN_ROADMAP_CREDENTIALS", "https://www.googleapis.com/auth/spreadsheets")
```
ao-david, line 303: Is there a reason why we don't have these credentials loaded during setup? They seem to get loaded on every request.

cmd/roadmap/main.go
```diff
@@ -0,0 +1,389 @@
...
+	if err != nil {
+		return c.Status(fiber.StatusInternalServerError).SendString(fmt.Sprintf("Auth failed: %v", err))
```
ao-david, line 305: isn't the errorHandler func supposed to handle this with a proper JSON return?

ao-david (commented):
Few comments if you want to address, mostly curious why auth.go has been copied again?

trekhopton:
Will address comments in upcoming PRs.

PR #743, opened 2026-06-02
broadcast page (show): load broadcast config from API endpoint 

trekhopton:
Done so that we can phase out go templates and more easily add UI elements.

This is in preparation for a cleaner user friendly UI element for filtering broadcast notifications.

Screenshot from 2026-06-02 17-36-59


cmd/oceanbench/api_v1.go
```diff
@@ -428,3 +429,38 @@ func parsePins(inputs string) []string {
 	}
 	return pins
 }
+
+// getV1BroadcastHandler handles GET /api/v1/broadcasts/:uuid
+// Returns the broadcast configuration matching the provided UUID within the user's selected site.
+func getV1BroadcastHandler(c *fiber.Ctx) error {
```
ao-david, line 435: Was there a particular reason to not use the existing endpoint?

`/api/get/broadcast/config`

PR #745, opened 2026-06-03
tv (show): move notification suppression to logAndNotify

trekhopton:
This was done because we weren't checking for suppression rules when using logAndNotify directly. This resulted in unwanted emails.


saxon-milton:
Looks like you've got some build errors FYI

saxon-milton (approved):

PR #746, opened 2026-06-03
bench (show): commit version in footer on all pages

trekhopton:
Added the commit version to the footer with the program (bench) version so it's easier to see what code is running (despite not bumping the version in main.go).

Applied the lit element footer to all pages and deleted html template footer for consistency.

Button to TV Overview is there temporarily, needs to be move to the nav bar.

image

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -61,6 +61,7 @@ import (
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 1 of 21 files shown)


ao-david:
RE: tv overview in nav bar, I think we should consider adding a super admin section to the nav bar, or adding a seperate way of navigating to super admin features.

PR #747, opened 2026-06-03
OTV: Refactor direct broadcast event handling

ao-david:
This change factors out event handling into state member functions. This makes it easier to trace all possible transitions out of states for all events.

There has been a stateWithBroadcastEventHandler interface added to allow for a partial refactor.

### DIFF
--- a/cmd/oceantv/broadcast_states.go
+++ b/cmd/oceantv/broadcast_states.go
@@ -75,6 +75,10 @@ type stateFields struct{}
 func (s *stateFields) enter() {}
 func (s *stateFields) exit()  {}
 
+type stateWithBroadcastEventHandler interface {
+	handleEvent(sm *broadcastStateMachine, event event)
+}
+
 type fixableState interface {
 	state
 	fix()
--- a/cmd/oceantv/broadcast_machine.go
+++ b/cmd/oceantv/broadcast_machine.go
@@ -212,8 +212,7 @@ func (sm *broadcastStateMachine) handleInvalidConfigurationEvent(event invalidCo
 		sm.transition(newVidforwardSecondaryIdle(sm.ctx))
 
 	case
-		*directStarting,
-		*directLiveUnhealthy:
+		*directStarting:
 
 		sm.transition(newDirectFailure(sm.ctx, event))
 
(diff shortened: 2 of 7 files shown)


trekhopton (approved):
Sounds good to me. I like how it reads better. When do you think the rest of it will be refactored? Might be worth writing an ADR or something so that the decision is recorded somewhere and can be outworked over time.

ao-david:
> Sounds good to me. I like how it reads better. When do you think the rest of it will be refactored? Might be worth writing an ADR or something so that the decision is recorded somewhere and can be outworked over time.

It should be relatively trivial to refactor the rest of the states once we have the form agreed on. This PR is small purely to make it easier to understand and show a proof of concept

cmd/oceantv/broadcast_states.go
```diff
@@ -75,6 +75,10 @@ type stateFields struct{}
 func (s *stateFields) enter() {}
 func (s *stateFields) exit()  {}
 
+type stateWithBroadcastEventHandler interface {
```
saxon-milton, line 78: probably can be called `stateWithEventHandler` ?

saxon-milton (approved):
LGTM. I think we're probably finding a good trade off here. The benefit of the SM level event handling was that we could have shared handling of events for similar states e.g. the "start" states, and also that states could remain decoupled of other states (states now need to know about other states to transition). Having said this, a good argument can be made for the readability improvements we get here.

cmd/oceantv/broadcast_states.go
ao-david: I was thinking about this, but we may want to use a similar setup for hardware states as well. Currently there isn't (at least that I know of/can find) an interface that allows us to use either a broadcastStateMachine or a hardwareStateMachine.

I was thinking about adding a statemachineContextor interface:
```golang
type stateMachineCtxer interface {
  Ctx() *broadcastCtx
}
```

Which would then mean we could do:
```golang
type stateWithEventHandler interface {
  handleEvent(sm *stateMachineCtxer, event event)
}
```

Or even, once all states have `handleEvent` implemented just add the member to the base `state` interface.

ao-david:
> The benefit of the SM level event handling was that we could have shared handling of events for similar states e.g. the "start" states

I wonder if this actually somewhat added to the readability issue, as the state machines aren't at all separated out, and so it makes it harder to trace one state machine

saxon-milton:
> > The benefit of the SM level event handling was that we could have shared handling of events for similar states e.g. the "start" states
> 
> I wonder if this actually somewhat added to the readability issue, as the state machines aren't at all separated out, and so it makes it harder to trace one state machine

Yes, that is true; if we've made the determination they actually are separate state machines. Not sure if we've committed to that yet. If do we have them as separate state machines, it's just a lot more code and boilerplate, which this code already suffers (more a problem with go as a language IMO).

PR #750, opened 2026-06-04
ADR: Add intro and ADR 0001 to docs

ao-david:


docs/decisions/0001-per-state-event-handling.md
ao-david, line 1: NOTE: This is a straight copy from AOTV

trekhopton (approved):

PR #766, opened 2026-06-15
Remove FileStore-specific device lookup logic

pav2345:
I removed the code  

_, filestore := store.(*datastore.FileStore)
if filestore {
    return getDevicesBySiteFromFileStore(ctx, store, skey)
} 

go test .model/-v test all passed

### DIFF
--- a/model/device.go
+++ b/model/device.go
@@ -28,7 +28,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


model/device.go
ao-david: Should be able to remove this function now too, as nothing should be calling it.

ao-david (changes requested):

model/device.go
pav2345: Removed the unused `getDevicesBySiteFromFileStore` function and cleaned up the now-unused `sort` import. Reran `go test ./model -v` successfully.
  thank you David for guiding me  to learn while contributing to opensource these is my first Contribition in golang

ao-david (approved):
LGTM!

PR #778, opened 2026-06-26
Bench: Upgrade Development tools

ao-david:
This change upgrades the build pipeline for oceanbench by removing
rollup, and moving to vite. This significantly decreases build times. An
air config file has also been added, this allows air to proxy requests
meaning that air will refresh the browser on changes, which makes
developing ts/js changes more seamless, although not quite as good as
pure vite handling yet.


cmd/oceanbench/vite.config.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export default defineConfig({
+  root: "./",
+  base: "/s/",
+  plugins: [tailwindcss()],
+  build: {
+    outDir: "s/lit",
+    rollupOptions: {
+      input: [
+        "ts/broadcast-states.ts",
```
elliot-ausocean: Would probably be better to make this a glob like in ausoceantv so that we don't have to change this every time we create a new file.

elliot-ausocean (approved):

trekhopton (approved):

PR #779, opened 2026-06-26
Bench/Broadcast: Add state progress bar

ao-david:
This change adds a stepped progress bar to the top of the broadcast page which polls for updates to the current broadcast.

requires #778 

image

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -76,7 +76,7 @@ import (
 )
 
 const (
-	version     = "v0.36.5"
+	version     = "v0.37.0"
 	localSite   = "localhost"
 	localDevice = "localdevice"
 	localEmail  = "localuser@localhost"
--- a/cmd/oceanbench/t/broadcast.html
+++ b/cmd/oceanbench/t/broadcast.html
@@ -12,6 +12,7 @@
       }
     </style>
     <script type="module" src="/s/lit/header-group.js"></script>
+    <script type="module" src="/s/lit/broadcast-states.js"></script>
     <script type="text/javascript" src="/s/main.js"></script>
     <script type="text/javascript" src="/s/broadcast.js"></script>
     <script type="module" src="/s/lit/site-footer.js"></script>
@@ -40,6 +41,7 @@
 
       <h1 class="container-md">Broadcast</h1>
       <div class="border rounded p-4 container-md bg-white position-relative">
+        <broadcast-states id="broadcast-state-disp"></broadcast-states>
         <div id="loading-overlay" class="d-none position-absolute top-0 start-0 w-100 h-100 bg-white d-flex justify-content-center align-items-center" style="z-index: 10; opacity: 0.7">
           <div class="spinner-border text-primary" role="status">
             <span class="visually-hidden">Loading...</span>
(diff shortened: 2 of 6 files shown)


elliot-ausocean (approved):

trekhopton (approved):
Nice one!

cmd/oceanbench/main.go
```diff
@@ -76,7 +76,7 @@ import (
 )
 
 const (
-	version     = "v0.36.5"
+	version     = "v0.36.6"
```
SaxonOB: Probably a minor bump given new feature

cmd/oceanbench/ts/broadcast-states.ts
```diff
@@ -0,0 +1,171 @@
+import { html, PropertyValues } from "lit";
+import { customElement, property, state } from "lit/decorators.js";
+import { TailwindElement } from "./shared/tailwind.element";
+import { type Broadcast } from "./types/broadcast";
+
+// Maps the state to the progress amount.
+enum Progress {
+  IDLE = 0,
+  STARTING = 34,
+  LIVE = 68,
```
SaxonOB: Maybe a dumb question but why is this a sparse enum ?

cmd/oceanbench/t/broadcast.html
```diff
@@ -40,6 +41,7 @@
 
       <h1 class="container-md">Broadcast</h1>
       <div class="border rounded p-4 container-md bg-white position-relative">
+        <broadcast-states id="bc-state-disp"></broadcast-states>
```
SaxonOB: Any reason id can't be `broadcast-state-disp` ? i don't think it's used frequently enough to warrant an abbreviation for broadcast.

cmd/oceanbench/ts/broadcast-states.ts
```diff
@@ -0,0 +1,171 @@
+import { html, PropertyValues } from "lit";
+import { customElement, property, state } from "lit/decorators.js";
+import { TailwindElement } from "./shared/tailwind.element";
+import { type Broadcast } from "./types/broadcast";
+
+// Maps the state to the progress amount.
+enum Progress {
```
SaxonOB: Should this be `State` ? Progress sounds more like a numeric percentage or something.

SaxonOB: Actually I just realised what this is doing. I'm thinking maybe this shouldn't be an enum where the values are direct maps to a numeric progres. I think maybe you need an enum, and then a map to map from enum values to progress amounts. This is probably a lot more intuitive.

SaxonOB: Alternatively if you don't think it's useful having a state enum, maybe you should just have a map of the string state from the broadcast config to the percentage. Then you won't need the switch later in the code.

SaxonOB (changes requested):

cmd/oceanbench/ts/broadcast-states.ts
ao-david: I think the map probably makes the most sense.

ao-david:
@SaxonOB your suggestions have been addressed, I have used a map instead, which as you said removes the switch statement.

SaxonOB (approved):

PR #785, opened 2026-07-10
OTV: Create events package

ao-david:
This change adds a new package for events, as well as making all the mechanical find and replace changes.


cmd/oceantv/event/event.go
ao-david, line 1: This is just a copy paste + exporting required types and methods.

cmd/oceantv/event/event_test.go
ao-david, line 1: This is just a copy paste + updating references

cmd/oceantv/event/hardware_event.go
ao-david, line 1: Copy paste + exporting references

cmd/oceantv/notifier/kinds.go
ao-david, line 1: Makes a new package so that the event package can import these

cmd/oceantv/notifier/notifier.go
ao-david, line 1: Creates a global notifier to be shareable between main and packages.

ao-david:
I would recommend only reviewing files that are in child directories. All files in `oceantv/` are just going to be find and replace changes to update references.

cmd/oceantv/broadcast/errors.go
```diff
@@ -0,0 +1,28 @@
+/*
+AUTHORS
+  David Sutton <davidsutton@ausocean.org>
```
Saxon1, line 3: Did you soley author all these files 😆 ?

Saxon1 (approved):
Looks fine, but would appreciate being added as an author on the files. I ack that it looks like I didn't have a header on the original files.

cmd/oceantv/broadcast/errors.go
ao-david: I put you as the author on most of them, must have missed a couple. Will update

PR #804, opened 2026-07-21
Bench: Prefix pages in nav menu

ao-david:
This change adds a skey prefix to the nav menu. Currently this only happens for sites that are ready to handle this.

### DIFF
--- a/cmd/oceanbench/main.go
+++ b/cmd/oceanbench/main.go
@@ -467,7 +467,7 @@ func indexHandler(c *fiber.Ctx) error {
 
 	profile, err := getProfile(c)
 	data := commonData{
-		Pages:   pages("home"),
+		Pages:   pages(c, "home"),
 		Profile: profile,
 	}
 	if err != nil {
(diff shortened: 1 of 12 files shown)


trekhopton (approved):

cmd/oceanbench/main.go
```diff
@@ -635,6 +635,24 @@ func hasPermission(ctx context.Context, p *gauth.Profile, mid, perm int64) (bool
...
+// 1. From the URL.
+// 2. The user's default site key.
+func getCurrentSkey(c *fiber.Ctx, profile *gauth.Profile) (int64, error) {
+	// First try to get from the URL param.
+	var skey int64
+	skFromParam := c.Params(skeyParamKey)
+	skey, err := strconv.ParseInt(skFromParam, 10, 64)
+	if err != nil {
+		log.Printf("unable to get skey from URL param: %v", err)
+		skey, err = getDefaultSkey(c.UserContext(), profile)
```
trekhopton, line 648: Did we decide to deprecate this method of storing and getting the site key ie. from user profile? Or is it still useful?

ao-david: It's not stored in the user profile, instead being stored in a variable. However, this should only be used on navigating to pages, and not in apis.

PR #813, opened 2026-07-21
Bench: prefix superadmin tools with skey

ao-david:
This change prefixes superadmin tools with skey, and manages authentication directly in the handlers, rather than through the catch all admin handler, simplifying the flow


trekhopton (approved):
Nice

PR #819, opened 2026-07-29
datablue: add mtsv2 endpoint and mtsmediav2 model

elliot-ausocean:
This change adds a new mtsv2 endpoint and model that receives and stores metadata about mts clips that are stored in a storage bucket. This is the first step in implementing OceanMedia.

### DIFF
--- a/cmd/datablue/main.go
+++ b/cmd/datablue/main.go
@@ -96,6 +96,7 @@ func main() {
 	http.HandleFunc("/act", actHandler)
 	http.HandleFunc("/vars", varsHandler)
 	http.HandleFunc("/mts", mtsHandler)
+	http.HandleFunc("/v2/mts", mtsHandlerV2)
 	http.HandleFunc("/recv", mtsHandler) // For backwards compatibility.
 	http.HandleFunc("/api", apiHandler)
 	http.HandleFunc("/api/", apiHandler)
--- a/cmd/datablue/mts_handler.go
+++ b/cmd/datablue/mts_handler.go
@@ -148,6 +148,127 @@ func mtsHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprint(w, string(jsn))
 }
 
+// mtsHandlerV2 receives audio/video data from devices in the form of
+// JSON blobs containing metadata, with the video stored in a storage bucket.
+// The JSON blob follows the same format as MTSMediaV2 but the MID, Geohash, and
+// Timestamp are ignored (they will be set by the server).
+// The response is in JSON format. For a normal response, the response mirrors the request
+// query params and their values, plus a timestamp (and minus the
+// device key which is never revealed to clients). For errors, the
+// response includes the "er" param. Server-side errors are also
+// logged. Where we receive multiple pin params, POST data represents
+// concatenated metadata blobs and the pin value indicates the size of each
+// blob. It is therefore possible to send multiple audio/video blobs in the same request.
+// The supplied MAC address (ma) must correspond to a valid
+// ausocean device and the supplied device key (dk) must to match
+// the device's. The pin type (pn) must be either V(ideo) or S(ound).
+func mtsHandlerV2(w http.ResponseWriter, r *http.Request) {
+	logRequest(r)
+	ctx := r.Context()
+
+	q := r.URL.Query()
+	macAddr := q.Get("ma")
+	dk := q.Get("dk")
+
+	// Is this request for a valid device?
+	setup(ctx)
+	dev, err := model.CheckDevice(ctx, settingsStore, macAddr, dk)
+	if err != nil {
+		writeDeviceError(w, dev, err)
+		return
+	}
+
+	geoHash := q.Get("gh")
+
+	t := q.Get("ts")
+	var timestamp int64
+	if t != "" {
+		timestamp, err = strconv.ParseInt(t, 10, 64)
+		if err != nil {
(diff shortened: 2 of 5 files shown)


model/mtsmediav2.go
```diff
@@ -0,0 +1,91 @@
...
+	typeMtsMediaV2 = "MtsMediaV2" // MtsMediaV2 datastore type.
+)
+
+type MtsMediaV2 struct {
+	MID           int64     // Media ID.
+	Geohash       string    // Geohash, if any.
+	Timestamp     int64     // Timestamp (in seconds).
+	Duration      int64     // Duration of the clip (in milliseconds).
+	StorageURI    string    // URI of the clip in the storage bucket.
+	BroadcastID   string    // OceanMedia broadcast ID.
```
ao-david: Does `BroadcastID` make sense for Audio? If we had a hydrophone, would it also be stored using `MtsMediaV2`, and if so, would BroadcastID still apply, or would it be ignored because the audio isn't necessarily discretised into a "broadcast"?

model/mtsmediav2.go
```diff
@@ -0,0 +1,91 @@
...
+	Date          time.Time // Date/time this record was created.
+	datastore.NoCache
+}
+
+// Implements Copy from the Entity interface.
+func (mts *MtsMediaV2) Copy(dst datastore.Entity) (datastore.Entity, error) {
+	return datastore.CopyEntity(mts, dst)
+}
+
+func MtsMediaV2Key(store datastore.Store, mid int64, timestamp int64) *datastore.Key {
```
ao-david: Write documenting comments for all exported methods.

ao-david: To keep with the datastore abstraction, I don't think we want to be exporting the key generator?

model/mtsmediav2.go
```diff
@@ -0,0 +1,91 @@
...
+	return store.NameKey(typeMtsMediaV2, fmt.Sprintf("%d.%d", mid, timestamp))
+}
+
+func CreateMtsMediaV2(ctx context.Context, store datastore.Store, mts MtsMediaV2) (*MtsMediaV2, error) {
+	key := MtsMediaV2Key(store, mts.MID, mts.Timestamp)
+	mts.Date = time.Now()
```
ao-david, line 66: Does the date have any impact on playback, or is it only for querying? If it relates to playback, do we need to be careful about where the time is calculated?

model/mtsmediav2.go
```diff
@@ -0,0 +1,91 @@
...
+
+	This is distributed in the hope that it will be useful, but WITHOUT
+	ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+	or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+	License for more details.
+
+	You should have received a copy of the GNU General Public License in
+	gpl.txt. If not, see http://www.gnu.org/licenses/.
+*/
+package model
```
ao-david, line 25: I think the spacing of the header is a little strange as there needs to be a blank line between the comment and the package name, otherwise go treats the comment as a description of the package (which we don't want)

cmd/datablue/main.go
```diff
@@ -96,6 +96,7 @@ func main() {
 	http.HandleFunc("/act", actHandler)
 	http.HandleFunc("/vars", varsHandler)
 	http.HandleFunc("/mts", mtsHandler)
+	http.HandleFunc("/mtsv2", mtsHandlerV2)
```
ao-david: We should consider using the versioning system which we are adopting on `cloudblue` and `ausoceantv`:
```/v1/mtsHandler```
```/v2/mtsHandler```

File: cmd/datablue/mts_handler.go [hunk 6]
```diff
@@ -148,6 +148,144 @@ func mtsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	resp := make(map[string]interface{})
+	resp["ma"] = macAddr
+
+	var found bool
+	for _, pin := range dev.InputList() {
+		if !isMtsPin(pin) {
+			continue
+		}
+		found = true
```
ao-david: I think found = true should go after `v == ""` check

cmd/datablue/mts_handler.go
```diff
@@ -148,6 +148,144 @@ func mtsHandler(w http.ResponseWriter, r *http.Request) {
...
+			break
+		}
+		if n != sz {
+			log.Printf("invalid size: n = %d, sz=%d", n, sz)
+			resp["er"] = errInvalidSize.Error()
+			break
+		}
+
+		var mtsData mtsMetadata
+		err = json.Unmarshal(clip, &mtsData)
```
ao-david: Is there a reason why we don't just unmarshal directly into the mtsmediav2 type? rather than creating a new type just for this?

ao-david (changes requested):
A few suggestions.

model/mtsmediav2.go
elliot-ausocean: I think that it would be a good idea to make hydrophones broadcasts too to unify scheduling. This would also make sense if we want to do audio only broadcasts on AusOceanTV, which I think we would want to do.

model/mtsmediav2.go
elliot-ausocean: This is only for querying. I think being able to find clips at a certain datetime regardless of broadcast could be useful.

ao-david (approved):

PR #820, opened 2026-07-29
bench (show): add air installation instructions to readme

trekhopton:
Done because aair will be the main way we develop on bench.


ao-david (changes requested):
We should actually add air to the tools in the go mod, and call it using `go tools air` like we do on ausoceantv, this means a dev doesn't need to install air manually

ao-david (approved):

PR #823, opened 2026-08-04
oceantv: refactor yt

elliot-ausocean:
This change refactors the yt package by splitting it into two separate packages - youtube and broadcast_host. Youtube is split out as another service as another service is using it. yt is renamed to broadcast_host as it is a more generic name. BroadcastService is renamed to BroadcastHost as there are a lot of things named service and it isn't referring to an OceanTV service but an external host of the live stream.

### DIFF
--- a/cmd/oceantv/broadcast_host/broadcast_host.go
+++ b/cmd/oceantv/broadcast_host/broadcast_host.go
@@ -1,0 +1,69 @@
+/*
+AUTHORS
+  Saxon Nelson-Milton <saxon@ausocean.org>
+
+LICENSE
+  Copyright (C) 2026 the Australian Ocean Lab (AusOcean)
+
+  This file is part of Ocean TV. Ocean TV is free software: you can
+  redistribute it and/or modify it under the terms of the GNU
+  General Public License as published by the Free Software
+  Foundation, either version 3 of the License, or (at your option)
+  any later version.
+
+  Ocean TV is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  in gpl.txt. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package broadcast_host
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/ausocean/cloud/youtube"
+)
+
+// ServerResponse is an interface for a server response.
+type ServerResponse interface {
+	fmt.Stringer
+	StatusCode() int
+	HTTPHeader() http.Header
+}
+
+// BroadcastOption is an option to pass when creating a new broadcast host.
+type BroadcastOption func(interface{}) error
+
+// BroadcastHost is an interface for a broadcast host that the camera will stream to.
+// For example, YouTube and OceanMedia.
+type BroadcastHost interface {
+	CreateBroadcast(
+		ctx context.Context,
+		broadcastName, description, streamName, privacy, resolution string,
+		start, end time.Time,
+		opts ...BroadcastOption,
+	) (ServerResponse, youtube.IDs, string, error)
+
+	StartBroadcast(
+		name, bID, sID string,
+		saveLink func(key, link string) error,
+		extStart, extStop func() error,
+		notify func(msg string) error,
+		onLiveActions func() error,
+	) error
+
+	BroadcastStatus(ctx context.Context, id string) (string, error)
+	BroadcastScheduledStartTime(ctx context.Context, id string) (time.Time, error)
+	BroadcastHealth(ctx context.Context, sid string) (string, error)
+	RTMPKey(ctx context.Context, streamName string) (string, error)
+	CompleteBroadcast(ctx context.Context, id string) error
+	PostChatMessage(cID, msg string) error
+	SetBroadcastPrivacy(ctx context.Context, id, privacy string) error
+}
--- a/cmd/oceantv/broadcast_host/youtube-standalone.go
+++ b/cmd/oceantv/broadcast_host/youtube-standalone.go
@@ -27,9 +27,9 @@ LICENSE
   in gpl.txt. If not, see <http://www.gnu.org/licenses/>.
 */
 
-// package yt provides functionality for setting up a YouTube livestream
+// package youtube provides functionality for setting up a YouTube livestream
 // service and broadcast scheduling.
-package yt
+package broadcast_host
 
 import (
 	"log"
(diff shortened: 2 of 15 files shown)


cmd/oceantv/broadcast_host/youtube-standalone.go
```diff
@@ -27,9 +27,9 @@ LICENSE
   in gpl.txt. If not, see <http://www.gnu.org/licenses/>.
 */
 
-// package yt provides functionality for setting up a YouTube livestream
+// package youtube provides functionality for setting up a YouTube livestream
```
ao-david: This comment doesn't line up with the package name.

cmd/oceantv/broadcast_host/youtube.go
```diff
@@ -20,7 +20,7 @@ LICENSE
   in gpl.txt. If not, see <http://www.gnu.org/licenses/>.
 */
 
-package yt
+package broadcast_host
```
ao-david: Review these guidelines: https://go.dev/blog/package-names

Consider a different name for the package, even if it is just `broadcasthost`

cmd/oceantv/broadcast_host/youtube.go
```diff
@@ -79,26 +43,26 @@ func (y YouTubeResponse) String() string          { return fmt.Sprintf("%v", goo
 func (y YouTubeResponse) StatusCode() int         { return googleapi.ServerResponse(y).HTTPStatusCode }
 func (y YouTubeResponse) HTTPHeader() http.Header { return googleapi.ServerResponse(y).Header }
 
-// YouTubeBroadcastService is a BroadcastService implementation for YouTube.
-type YouTubeBroadcastService struct {
+// YouTubeBroadcastHost is a BroadcastHost implementation for YouTube.
+type YouTubeBroadcastHost struct {
```
ao-david: Again following go idioms, we should try to avoid repeating the package name in exported types and methods.
Consider:
```go
broadcasthost.YoutubeBroadcastHost // Includes repitition.
// vs
broadcasthost.Youtube // No repitition but still clear.
// or
broadcasthost.YoutubeHost // A little repitition but more concise.
```

cmd/oceantv/broadcast_host/youtube.go
```diff
@@ -79,26 +43,26 @@ func (y YouTubeResponse) String() string          { return fmt.Sprintf("%v", goo
...
 	limiter  ratelimit.RateLimiter
 	log      func(string, ...interface{})
 	tokenURI string
 }
 
-func NewYouTubeBroadcastService(tokenURI string, log func(string, ...interface{})) *YouTubeBroadcastService {
-	return &YouTubeBroadcastService{log: log, tokenURI: tokenURI}
+func NewYouTubeBroadcastHost(tokenURI string, log func(string, ...interface{})) *YouTubeBroadcastHost {
```
ao-david: consider:
```go
broadcasthost.NewYoutube
// or 
broadcasthost.NewYoutubeHost
```

cmd/oceantv/broadcast_host/broadcast_host.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/ausocean/cloud/youtube"
+)
+
+// ServerResponse is an interface for a server response.
+type ServerResponse interface {
```
ao-david: Should Server response even just be response, since the response is coming from the broadcast host?
```go
broadcasthost.ServerResponse
// vs
broadcasthost.Response
```

cmd/oceantv/broadcast_host/broadcast_host.go
```diff
@@ -0,0 +1,69 @@
...
+	fmt.Stringer
+	StatusCode() int
+	HTTPHeader() http.Header
+}
+
+// BroadcastOption is an option to pass when creating a new broadcast host.
+type BroadcastOption func(interface{}) error
```
ao-david: I think this should just be `Option` (`broadcasthost.Option` vs `broadcasthost.BroadcastOption`)

ao-david: Also, we should be able to be more precise with the args:
```go
type Option func(h *BroadcastHost) error
```

cmd/oceantv/broadcast_host/youtube.go
```diff
@@ -79,26 +43,26 @@ func (y YouTubeResponse) String() string          { return fmt.Sprintf("%v", goo
...
+// YouTubeBroadcastHost.
 func WithRateLimiter(limiter ratelimit.RateLimiter) BroadcastOption {
 	return func(i interface{}) error {
-		if s, ok := i.(*YouTubeBroadcastService); ok {
+		if s, ok := i.(*YouTubeBroadcastHost); ok {
 			s.limiter = limiter
 			return nil
 		}
-		return errors.New("this option is not for YouTubeBroadcastService")
+		return errors.New("this option is not for YouTubeBroadcastHost")
```
ao-david: I know this wasn't your error message, but I think this error logic is backwards, should be: `this option is only for YouTubeBroadcastHost`

cmd/oceantv/broadcasthost/broadcasthost.go
```diff
@@ -0,0 +1,69 @@
...
+	StatusCode() int
+	HTTPHeader() http.Header
+}
+
+// BroadcastOption is an option to pass when creating a new broadcast host.
+type BroadcastOption func(interface{}) error
+
+// BroadcastHost is an interface for a broadcast host that the camera will stream to.
+// For example, YouTube and OceanMedia.
+type BroadcastHost interface {
```
ao-david: We could also consider calling this just `Host`
```golang
broadcasthost.BroadcastHost
// vs
broadcasthost.Host // broadcast context is contained in the package name
```

youtube/storage.go
```diff
@@ -25,7 +25,7 @@ LICENSE
   in gpl.txt. If not, see <http://www.gnu.org/licenses/>.
 */
 
-package yt
+package youtube
```
ao-david: I'm not convinced by this name change. The youtube package from google is already used in other parts of the codebase. From the go package name guide:

> Avoid unnecessary package name collisions. While packages in different directories may have the same name, packages that are frequently used together should have distinct names. This reduces confusion and the need for local renaming in client code. For the same reason, avoid using the same name as popular standard packages like io or http.

Since youtube is a popular package name already, we should avoid overlapping. Unless, we are completely not calling the OG youtube package in any code, and only this package.

I support renaming it, but I think `youtube` is worse. Consider maybe `ytclient`

ao-david (changes requested):
I think the naming could be improved to better match go idioms

cmd/oceantv/broadcasthost/broadcasthost.go
elliot-ausocean: I personally prefer leaving it as BroadcastHost but I could be convinced either way.

ao-david: I think unless we have a good reason, we should follow the go idiom.

